PR #3, opened 2024-04-08
chore: add .gitignore file for python,c,c++

onuralpszr:


karpathy:
I don't super like these boilerplate massive gitignore files. Might add a targeted one later.

PR #4, opened 2024-04-08
Minor correction for openmp check

Ricardicus:
Hi,

Amazing project :D 

Thank you! 

I needed this to work for my mac, I didnt have omp

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,16 +10,16 @@ INCLUDES =
 # e.g. on MacOS: brew install libomp
 # e.g. on Ubuntu: sudo apt-get install libomp-dev
 # later, run the program by prepending the number of threads, e.g.: OMP_NUM_THREADS=8 ./gpt2
-ifeq ($(shell echo | $(CC) -Xpreprocessor -fopenmp -x c -E - > /dev/null 2>&1; echo $$?), 0)
+ifeq ($(shell echo | $(CC) -fopenmp -x c -E - > /dev/null 2>&1; echo $$?), 0)
   ifeq ($(shell uname), Darwin)
     # macOS with Homebrew
-    CFLAGS += -Xclang -fopenmp
+    CFLAGS += -Xclang -fopenmp -DOMP
     LDFLAGS += -L/opt/homebrew/opt/libomp/lib
     LDLIBS += -lomp
     INCLUDES += -I/opt/homebrew/opt/libomp/include
   else
     # Ubuntu or other Linux distributions
-    CFLAGS += -fopenmp
+    CFLAGS += -fopenmp -DOMP
     LDLIBS += -lgomp
   endif
   $(info NICE Compiling with OpenMP support)
--- a/train_gpt2.c
+++ b/train_gpt2.c
@@ -14,7 +14,9 @@ There will be other versions of this code that specialize it and make it fast.
 #include <time.h>
 #include <string.h>
 #include <unistd.h>
+#ifdef OMP
 #include <omp.h>
+#endif
 
 // ----------------------------------------------------------------------------
 // all the individual layers' forward and backward passes


karpathy:
oh, yeah, oops. I see the need for the guard around omp.h import. Is the other part in Makefile necessary?

karpathy:
(referring to the preprocessor absence, ok).

XiaoConstantine:
I think removing ` -Xpreprocessor ` actually break the check of proper installation of `libomp`

Ricardicus:
The reason I removed the -Xpreprocessor, was that it was always returning 0 even if openmp wasn't installed.

```   
$ echo |clang -Xpreprocessor -fopenmp -x c -E -;echo $?
# 1 "<stdin>"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 401 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "<stdin>" 2

0
``` 
I got a consistent negative response without it:
``` 
$ echo |clang -fopenmp -x c -E -;echo $? 
clang: error: unsupported option '-fopenmp'
1
``` 

It seems like installing libomp with brew install libomp is not enough on my mac.
My clang compiler still complains about -fopenmp.

If I pass -Xclang or -Xpreprocessor to the -fopenmp flag it is silenced. However,
I don't think openmp is used in that case. 

On my linux, I can compile with libopenmp using gcc.. 

Perhaps change CC to gcc ?

Ricardicus:
I included something that I think works on both my mac and on Linux.
It combines the original check, with a more basic directory check on mac.
There seems to be something different between the clang I use on Linux and the one in Mac, where
-fopenmp passed directly to clang creates a problem; but passing it via the -Xclang argument makes it work.
I don't know why this is, but I try to get around it here:
https://github.com/karpathy/llm.c/pull/14

Ricardicus:
I included something that works on both my mac and on Linux.
It combines the original check, with a more basic directory check on mac.
There seems to be something different between the clang I use on Linux and the one in Mac, where
-fopenmp passed directly to clang creates a problem; but passing it via the -Xclang argument makes it work.
I don't know why this is, but I try to get around it in a new PR here: https://github.com/karpathy/llm.c/pull/14

XiaoConstantine:
> The reason I removed the -Xpreprocessor, was that it was always returning 0 even if openmp wasn't installed.
> 
> ```
> $ echo |clang -Xpreprocessor -fopenmp -x c -E -;echo $?
> # 1 "<stdin>"
> # 1 "<built-in>" 1
> # 1 "<built-in>" 3
> # 401 "<built-in>" 3
> # 1 "<command line>" 1
> # 1 "<built-in>" 2
> # 1 "<stdin>" 2
> 
> 0
> ```
> 
> I got a consistent negative response without it:
> 
> ```
> $ echo |clang -fopenmp -x c -E -;echo $? 
> clang: error: unsupported option '-fopenmp'
> 1
> ```
> 
> It seems like installing libomp with brew install libomp is not enough on my mac. My clang compiler still complains about -fopenmp.
> 
> If I pass -Xclang or -Xpreprocessor to the -fopenmp flag it is silenced. However, I don't think openmp is used in that case.
> 
> On my linux, this check works and I can compile with libopenmp still. Maybe there is some issue with the standard clang included in mac and openmp? On Linux this is not an issue, and this check works there.
> 
> edit: Using -Xclang on mac actually works and openmp is included. However, there is an inconsistency here where the -X argument is not needed on Linux.

Thanks! I noticed the same behavior with `-Xpreprocessor` as well, evaluate result is always `0`, similarly when removing it , always evaluate to `1`.

On my m3 pro (with clang 15.0.0), I also noticed that following behavior:
```bash
 ${CC:-cc} -fopenmp -x c -E - > /dev/null 2>&1; echo $?
1
```

```bash
 ${CC:-cc} -Xpreprocessor -fopenmp -x c -E - > /dev/null 2>&1; echo $?


```
(this will hang)

seems related to clang

Issue #10, opened 2024-04-08
Calloc NULL not handled


chadbrewbaker:
``` bash
make -f Makefile_gnu  # changed to gcc-13, passing -fanalyzer to CFLAGS
```

Mostly clean but flagged that calloc can return NULL.

https://github.com/karpathy/llm.c/blob/3bcb9ba7d2e37f48e4b97806736783acd6da2f41/train_gpt2.c#L893

PR #12, opened 2024-04-08
Fixed CALLOC NULL w error message

chadbrewbaker:
Fixes https://github.com/karpathy/llm.c/issues/10

### DIFF
--- a/train_gpt2.c
+++ b/train_gpt2.c
@@ -18,6 +18,16 @@ There will be other versions of this code that specialize it and make it fast.
 #include <omp.h>
 #endif
 
+#define CHECK_NULL(ptr) \
+    if ((ptr) == NULL) { \
+        fprintf(stderr, "Null pointer encountered at %s:%d\n", __FILE__, __LINE__); \
+        fflush(stdout);  \
+        fflush(stderr);  \
+        exit(1); \
+    } 
+
+
+
 // ----------------------------------------------------------------------------
 // all the individual layers' forward and backward passes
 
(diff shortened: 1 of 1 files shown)


train_gpt2.c
```diff
@@ -18,6 +18,18 @@ There will be other versions of this code that specialize it and make it fast.
 #include <omp.h>
 #endif
 
+#ifndef NDEBUG // Make zero cost when not debugging
+    #define CHECK_NULL(ptr) \
+    if ((ptr) == NULL) { \
+        fprintf(stderr, "Null pointer encountered at %s:%d\n", __FILE__, __LINE__); \
+        fflush(stdout);  \
+        fflush(stderr);  \
```
N-R-K, line 26: Exiting the process causes all open streams to be flushed already.

train_gpt2.c
```diff
@@ -18,6 +18,18 @@ There will be other versions of this code that specialize it and make it fast.
 #include <omp.h>
 #endif
 
+#ifndef NDEBUG // Make zero cost when not debugging
```
N-R-K, line 21: If the assumptions is that it's "fine" (as in process terminating/crashing) to deref a null pointer, then what's the point of handling in debug mode?

If anything, it makes more sense to swap it around. Because if you're debugging, you are (or should be) already running in a debugger and/or have Undefined Sanitizer enabled which will break on the null deref, rendering the error printout as useless. Whereas on release builds the printout would've been more useful.

N-R-K (changes requested):

train_gpt2.c
chadbrewbaker: Not wrong. As I thought about this more maybe the callocs should not exist at all? Ideally there is one malloc() for an arena.

PR #14, opened 2024-04-09
Using the compiler at hand

Ricardicus:
I was on a Linux with gcc but no clang installed and when I compiled with gcc I got a bunch of these:

` warning: ignoring return value of 'fread' declared with attribute 'warn_unused_result' [-Wunused-result]`

So I thought this might be helpful to silence these warnings, but also to give a little flexibility as to what compiler to use.

After installing clang in Ubuntu, and building with clang, there was no such warnings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
-CC = clang
-CFLAGS = -O3 -Ofast
+CC ?= clang
+CFLAGS = -O3 -Ofast -Wno-unused-result
 LDFLAGS =
 LDLIBS = -lm
 INCLUDES =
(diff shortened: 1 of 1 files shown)


Ricardicus:
I have also taken into account what we discussed here: https://github.com/karpathy/llm.c/pull/4 about the -fopenmp flag.. This PR includes a change to the Makefile that combines the check used before with a more hands on solution in the Darwin case (checking if the library exists at /opt/homebrew/opt/libomp/lib).

karpathy:
like, ty

PR #24, opened 2024-04-09
enhanced tensor comparison with higher precision.

anurag12-webster:
instead of directly comparing the absolute difference of elements, we can square the difference 
```(a[i] - b[i]) * (a[i] - b[i])``` and compare it against a smaller threshold `1e-4.`


karpathy:
What is the advantage?

anurag12-webster:
> What is the advantage?

square operation is generally faster than the fabs function, as it avoids the additional logic required to handle the sign and on top of it  the square difference as fundamental can be easily further be improved through SIMD vectorization.

PR #25, opened 2024-04-09
tweak: instead of using -10000.0f for finding the max, use the first item

NunoSempere:


karpathy:
yeah, agree i think. there are two instances of this btw

NunoSempere:
I'm a bit unclear how to do the other case parsimoniously, but what do you think about the current state?

karpathy:
i'll come around to this eventually

PR #37, opened 2024-04-09
HIP support multigpu, AMD, Nvidia.

Avicted:
Added:
- Python virtual environment + automated setup (venv init + requirements.txt installation) through the makefile.
- AMD HIP support, to target Nvidia or AMD through compilation with the hipcc compiler.

Tested on Linux (Manjaro):
```bash
# Test setup
$ hipcc --version

HIP version: 6.0.32830-d62f6a171
AMD clang version 17.0.0 (https://github.com/RadeonOpenCompute/llvm-project roc-6.0.0 23483 7208e8d15fbf218deb74483ea8c549c67ca4985e)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /opt/rocm-6.0.0/llvm/bin
```

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -477,7 +477,7 @@ void matmul_forward_cublaslt(float* out,
         weightLayout, inputLayout, outputLayout, outputLayout,
         preference, 1, &heuristic, &returnedResults));
     if (returnedResults == 0) {
-        printf("No cuBLASLt algorithm: B: %d, T: %d, C: %d, OC: %d, bias: %d, gelu: %d\n", B, T, C, OC, has_bias);
+        printf("No cuBLASLt algorithm: B: %d, T: %d, C: %d, OC: %d, bias: %d\n", B, T, C, OC, has_bias);
         exit(EXIT_FAILURE);
     }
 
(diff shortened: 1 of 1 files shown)


Avicted:
Updated the pull request to only include the relevant updates.

Makefile
```diff
@@ -1,8 +1,19 @@
-CC ?= clang
 CFLAGS = -O3 -Ofast -Wno-unused-result
 LDFLAGS =
 LDLIBS = -lm
-INCLUDES =
+INCLUDES = 
+
+# AMD HIP, Nvidia and AMD support
+USE_HIP = false
```
Ricardicus, line 7: Consider the ?= operator.. Then I would be able to build like this:
USE_HIP=true make

also, for the CC , use ?= so that I can build with e.g. gcc if I don't have clang installed

Avicted: Good idea! I updated the makefile according to your suggestions.

Ricardicus: Nice, but iemove it for the include, it is only relevant for setting logics and compiler in my opinin :)

Ricardicus: Nice. But let INCLUDES have the += still. Also, consider rebasing this PR so it is only one big commit. That probably makes it feel less cluttered for Andrej. Also, his silence probably suggests that he needs time to review your changes. I for myself think it looks very interesting but I havent tested it.

Issue #43, opened 2024-04-10
Add .gitignore to the project. 


this-is-batman:
Add .gitignore to the project to keep the working tree clean.

PR #44, opened 2024-04-10
Added the .gitignore file.

this-is-batman:
- This will help in keeping the working tree clean.
- All the .bin files and the /data/ folder have been added to gitignore.


this-is-batman:
Fixes https://github.com/karpathy/llm.c/issues/43

this-is-batman:
Athough I realise this is not a necessary thing to do, and we can ignore this most of the time, this helps keep a clean working directory. I have tried to keep only the basic things in the .gitignore file.

davmacario:
What do you think about adding entries for commonly ignored files?
Something along the lines of [this](https://github.com/github/gitignore/blob/main/Python.gitignore).

this-is-batman:
> What do you think about adding entries for commonly ignored files? Something along the lines of [this](https://github.com/github/gitignore/blob/main/Python.gitignore).

This looks good, but I prefer to keep a small gitignore file, and improve upon it as the project matures.

PR #48, opened 2024-04-10
fix error in small typos in matmul_forward.cu

lancerts:
The sqrt_block_size is 32 in `    matmul_forward(kernel_num, d_out, d_inp, d_weight, d_bias, B, T, C, OC, 32);`
thus the printf should have 1024.


karpathy:
bleh copy paste and debugging gone wrong, ty

PR #51, opened 2024-04-10
fully fused layer-norm kernel

ngc92:
an implementation of layer-norm that doesn't require shared memory or other intermediate buffers. On my A4000, I get  a speedup from 192 GB/s for kernel two at block size 64 to 303 GB/s for this kernel at block size 32.

Not sure if you are OK with this code, since technically, it isn't pure C, making use of namespaces and templates. I would say it's still pure CUDA, though, since these things do come directly from CUDA, not any additional library.


ngc92:
updated to optionally also calculate mean and rstd. 
For pure inference, we can set these to 0 and don't have to waste memory bandwidth

PR #53, opened 2024-04-10
readability updates: param_size calcs

johnrose3000:
In train_gpt2.c[u], function gpt2_build_from_checkpoint() assigns model param_sizes based on ParameterTensor struct members. Propose inline comments to clarify.


karpathy:
added for both params and sizes in 1751af1

PR #59, opened 2024-04-11
Add CMake project for cross platform support and easier quick start setup

abuneri:
I'm a windows guy so I wanted to get this working on there. I also automated the data setup for tinyshakespeare, and I can do it for tinystories as well if wanted.

## Build Instructions

To configure, run the following. If you want the software/optional OpenMP implementation, don't add the `-cuda` flag:
`cmake --preset gpt2-tinyshakepeare[-cuda] [-G <generator>]`

To build, run the following. If you want to train only, don't add the `-test` portion:
`cmake --build --config <Release|Debug> --preset gpt2-tinyshakespeare[-cuda][-test]`

## Windows
Example using `gpt2-tinyshakepeare-cuda-test` on Windows

### Configure
windows_cuda_configure

### Build
windows_cuda_build

### Train
windows_cuda_train

### Test
windows_cuda_test

## Ubuntu 22.04 (WSL2)
`gpt2-tinyshakepeare-cuda-test` also builds for me on Ubuntu 22.04 (WSL2), but here is an example of just training via `gpt2-tinyshakespeare` using OpenMP impl instead

### Configure
ubuntu_openmp_configure

### Build
ubuntu_openmp_build

### Train
image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,129 @@
+cmake_minimum_required(VERSION 3.23)
+
+project(llm.c
+    LANGUAGES C
+    VERSION 0.1.0
+)
+
+set(CMAKE_C_STANDARD 11)
+set(CMAKE_C_STANDARD_REQUIRED True)
+
+set(GPT2_TRAIN_SOURCES platform_utils.h)
+set(GPT2_TRAIN_DEFINES)
+set(GPT2_TEST_SOURCES platform_utils.h)
+set(GPT2_LIB_DEPENDENCIES)
+set(GPT2_CUSTOM_DEPENDENCIES)
+set(GPT2_COMPILE_FLAGS)
+
+# optional cuda support
+if(ENABLE_CUDA)
+    find_package(CUDAToolkit)
+    if(NOT CUDAToolkit_FOUND AND WIN32)
+        message(SEND_ERROR "CUDA is not installed. Go to https://developer.nvidia.com/cuda-toolkit")
+    endif()
+
+    enable_language(CUDA)
+    set(CMAKE_CUDA_STANDARD 11)
+    set(CMAKE_CUDA_STANDARD_REQUIRED True)
+
+    list(APPEND GPT2_TRAIN_SOURCES train_gpt2.cu)
+    list(APPEND GPT2_TEST_SOURCES test_gpt2.cu)
+
+    list(APPEND GPT2_LIB_DEPENDENCIES CUDA::cublas)
+
+    # https://forums.developer.nvidia.com/t/suppress-warning-conversion-from-a-string-literal-to-char-is-deprecated/258050/3
+    list(APPEND GPT2_COMPILE_FLAGS "-diag-suppress" "2464")
+else()
+    list(APPEND GPT2_TRAIN_SOURCES train_gpt2.c)
+    list(APPEND GPT2_TEST_SOURCES test_gpt2.c)
+
+    find_package(OpenMP)
+    if(OpenMP_C_FOUND)
+        list(APPEND GPT2_LIB_DEPENDENCIES OpenMP::OpenMP_C)
+        list(APPEND GPT2_TRAIN_DEFINES OMP)
+    endif()
+endif()
+
+if(NOT WIN32)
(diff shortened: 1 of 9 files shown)


train_gpt2.c
```diff
@@ -10,14 +10,22 @@ There will be other versions of this code that specialize it and make it fast.
...
 #include <math.h>
 #include <time.h>
 #include <string.h>
-#include <unistd.h>
-#ifdef OMP
+#if defined(OMP) && !(_OPENMP < 202011)
 #include <omp.h>
+#define OMP_ENABLED
+#else
+#pragma message \
```
abuneri, line 23: Unfortunately, my version of Visual Studio 2022 is still only shipping with OpenMP 2.0, which is too old to use the syntax of some of your directives, the collapse clause for an example apparently was only added in 5.2 apparently: https://devblogs.microsoft.com/cppblog/open-mp-improvements-in-visual-studio-cpp/#collapse-clause-in-openmp-standard-5-2

train_gpt2.c
```diff
@@ -10,14 +10,22 @@ There will be other versions of this code that specialize it and make it fast.
...
 #include <stdlib.h>
+#ifdef _WIN32
+#define _USE_MATH_DEFINES
+#endif
 #include <math.h>
 #include <time.h>
 #include <string.h>
-#include <unistd.h>
-#ifdef OMP
+#if defined(OMP) && !(_OPENMP < 202011)
```
abuneri, line 19: Note that I don't think we need to manually define OMP anymore with this version check that is automatically defined if you depend on OpenMP. Not sure you even need it in the manual Makefile

train_gpt2.c
```diff
@@ -1054,8 +1075,7 @@ int main() {
 
     // some memory for generating samples from the model
     unsigned long long rng_state = 1337;
-    const int gen_max_length = 64;
-    int gen_tokens[gen_max_length];
+    int gen_tokens[GEN_MAX_LENGTH];
```
abuneri, line 1078: Windows C11 impl wasn't happy with this. Not sure what C standard you're using, but typically GCC/Clang are more strict so this was surprising

test_gpt2.cu
```diff
@@ -30,10 +31,10 @@ int main(int argc, char *argv[]) {
     GPT2 model;
     gpt2_build_from_checkpoint(&model, "gpt2_124M.bin");
 
-    int C = model.config.channels;
+    //int C = model.config.channels;
```
abuneri, line 34: The following warning I got on Windows CUDA:

```
`warning #177-D: variable "<var>" was declared but never referenced`
```
Only commented all these out instead of removing in case you still want them for something

train_gpt2.c
ross-wheeler: If I remember correctly, VLA's are a C99 feature that Microsoft does not support with their compiler. Be great if they did but they don't. I would recommend adding -Werror=vla to the GCC compiler options to avoid this in the future.

platform_utils.h
ross-wheeler, line 1: Another option is to skip the conditional compilation code and use separate platform c files with the same system functions for the linux, mac, and windows builds.

train_gpt2.c
ross-wheeler: I have this working (finally) on Windows. Have a number of changes and am hoping you can add your CMake changes to these changes. Should be pushed up over the weekend.

CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+
+    list(APPEND GPT2_LIB_DEPENDENCIES CUDA::cublas)
+
+    # https://forums.developer.nvidia.com/t/suppress-warning-conversion-from-a-string-literal-to-char-is-deprecated/258050/3
+    list(APPEND GPT2_COMPILE_FLAGS "-diag-suppress" "2464")
+else()
+    list(APPEND GPT2_TRAIN_SOURCES train_gpt2.c)
+    list(APPEND GPT2_TEST_SOURCES test_gpt2.c)
+
+    find_package(OpenMP)
```
lum1n0us, line 40:
```suggestion
    find_package(OpenMP 5.2)
```

Avoid involving the un-appropriate openmp building at the very beginning

Issue #57, opened 2024-04-11
Precompute the scaling factor in gelu_forward and gelu_backward


ryanmcdermott:
Thank you so much for the amazing repo!

The scaling factor in [gelu_forward](https://github.com/karpathy/llm.c/blob/d8e2a36f1538dbb38b43874731574aa9d224ef2a/train_gpt2.c#L339C5-L339C34) and [gelu_backward](https://github.com/karpathy/llm.c/blob/d8e2a36f1538dbb38b43874731574aa9d224ef2a/train_gpt2.c#L348) is computed each time the functions are run: `float s = sqrtf(2.0f / M_PI);`

This can be precomputed ahead of time and stored as a constant.

ayushanshul07:
Is this good enough?

andylolu2:
Doesn't it get optimised away by the compiler anyway? (Haven't actually checked though) Plus pointwise operations are bandwidth-limited anyway, so adding a few more flops shouldn't make a difference.

karpathy:
fixed

ayushanshul07:
Doesn't #defining constant is just replacement and the actual calculation is still repeated?
https://github.com/karpathy/llm.c/commit/f26cf00a615e4f643a3660071e4ef86010045db7

PR #71, opened 2024-04-11
Organize defined constants

modigeko:
Update for:
- [#57 - Precompute the scaling factor in gelu_forward and gelu_backward](https://github.com/karpathy/llm.c/issues/57)
- group and move all defined constants to the top


karpathy:
Fixed here for both CPU and GPU ty 
https://github.com/karpathy/llm.c/commit/f26cf00a615e4f643a3660071e4ef86010045db7

PR #72, opened 2024-04-11
-O3 cannot go with -Ofast

Soldy:


Soldy:
Small thing.

karpathy:
You're right my understanding is -Ofast is basically -O3 ++
docs https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

PR #79, opened 2024-04-11
Include the online softmax CPU code and a fully parallelized GPU kernal

lancerts:
1. Include the online softmax CPU code (from the paper [Online normalizer calculation for softmax](https://arxiv.org/pdf/1805.02867.pdf)).
2. Its native port to GPU kernel `kernel 5` (for education comparison).
3. Include the fully parallel kernel `kernel 6`, where the [reduction op](https://github.com/karpathy/llm.c/pull/79/files#diff-a00ef278da39f24a9d5cb4306c15626b921d437013fb5aa60ac2d8df6b5a5508R362) uses

image

and cooperative group (from @ngc92 ) .

RTX3070: 

`B=8, T=1024`
Kernal 4:
image

Kernel 6:
image

`B=64, T=1024`
Kernal 4:
image

Kernal 6:
image

### DIFF
--- a/dev/cuda/softmax_forward.cu
+++ b/dev/cuda/softmax_forward.cu
@@ -24,11 +24,9 @@ version 6 is softmax_online that parallelizes over all of B,T,C
 ./softmax_forward 6
 */
 
-#include <cuda_runtime.h>
-#include "device_launch_parameters.h"
-#include <math.h> 
 #include <stdio.h>
 #include <stdlib.h>
+#include <cuda_runtime.h>
 #include <cooperative_groups.h>
 #include <cooperative_groups/reduce.h>
 
--- a/dev/cuda/softmax_forward.cu
+++ b/dev/cuda/softmax_forward.cu
@@ -16,6 +16,9 @@ version 3 uses intra-warp reductions for maxval and sumval, must use block_size=
 version 4 uses both intra-warp reductions and shared memory for inter-warp reductions
 so it can tolerate any block_size % 32 == 0. this is hopefully the most efficient version
 ./softmax_forward 4
+
+version 5 is naive port from CPU code (softmax_online) to kernel: parallelizes over B,T, loops over C
+./softmax_forward 5
 */
 
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


karpathy:
What is the benefit of the online softmax for us?

lancerts:
> What is the benefit of the online softmax for us?

- It reduces the 3 for loops (1 compute max, 2 compute sum, 3 compute output) to 2 for loops (1 compute max and sum in one loop, 2 compute output). Therefore, it is an algorithmic improvement for faster computation.
- It is used in flash attention 1&2 as the building block in the derivation.

Need to implement the corresponding optimization used in kernel 2-4 for the online softmax as follow-ups...

dev/cuda/softmax_forward.cu
```diff
@@ -289,6 +318,30 @@ __global__ void softmax_forward_kernel4(float* out, float* inp, int N, int C) {
...
+        float* out_row = out + i * C;
+
+        float maxval = -INFINITY;
+        float sum = 0.0f;
+        for (int j = 0; j < C; j++) {
+            float maxval_prev = maxval;
+            if (inp_row[j] > maxval) {
+                maxval = inp_row[j];
+            }
+            sum = sum * expf(maxval_prev - maxval) + expf(inp_row[j] - maxval);
```
ngc92, line 336: this value will only be different from 1 if the previous if branch is actually taken, right? So we probably can skip the evaluation of exp most of the time

lancerts: Thanks for the comment, updated the code.

lancerts:
Updated with a full parallel kernel.

karpathy:
On my A100 I am seeing:

kernel 4:
```
block_size   32 | time 0.221143 ms
block_size   64 | time 0.096894 ms
block_size  128 | time 0.069505 ms
block_size  256 | time 0.066208 ms
block_size  512 | time 0.076968 ms
block_size 1024 | time 0.116055 ms
```

kernel 6:
```
block_size   32 | time 0.101643 ms
block_size   64 | time 0.102056 ms
block_size  128 | time 0.085717 ms
block_size  256 | time 0.084556 ms
block_size  512 | time 0.085275 ms
block_size 1024 | time 0.085103 ms
```

so yes as you mentioned in Discord the kernel is more consistent over block size, but maybe not as fast provided you choose the right block size.

PR #80, opened 2024-04-11
Draft: Layer norm v2

ngc92:
This isn't intended to be merged as it is, but a demonstration of more speed-ups that can be achieved if you can make a few more assumptions, e.g., a more recent CUDA compiler, templates, and a channel-dimension that is a multiple of 4.

Note that I've changed the number of channels in the benchmark part, because for more channels the code in the other PR would run out of L1 and slow down considerably.


ngc92:
outdated

PR #82, opened 2024-04-11
cooperative groups and fused scale kernel

ngc92:
Uses a single warp (instead of a block) per token, therefore relying entirely on wrap-level shuffle functionality provided by `cg::reduce`.
Instead of achieving causal attention through masking, since we're looping manually, we may just as well fix the loop bounds to be correct. Finally, do the scaling as part of the computation here, saving us one more round-trip to vram.

I believe this is orthogonal to #79, but I wanted to get the  simple version working first.


ngc92:
Note that this code implicitly relies on the fact that the out memory comes zero-initialized. If we cannot assume that, we need to add additional zero-writing, but that causes a measurable slowdown.

PR #87, opened 2024-04-12
Use the command 'brew --prefix libomp' to retrieve the location where libomp would be installed on macOS.

linmajia:
The following unexpected cases have been tested:
1. brew or libomp is not installed.
2. Homebrew is installed in a location other than the default one, e.g., $HOME/homebrew.
3. Homebrew's install path contains blanks.


chadbrewbaker:
Should be in Readme or brew_requirements.txt? Makefiles messing with system libraries are not cool.

PR #95, opened 2024-04-12
Print total training time

krrishnarraj:
Will be useful to track incremental performance improvements


karpathy:
ok to stick with iteration time for now

PR #98, opened 2024-04-12
use cublaslt and optionally tf32, which fuses bias

karpathy:
cc @ademeure , this is cuBLASLt version of matmul only, for now, with a bunch of smaller changes. E.g. I tried to take out global variables and settings, which by default I prefer to do as much as possible. I think the workspace is the only thing that has to stay global (?). (Without streams/graphs involved). I also changed the workspace to 4MiB instead of 64MiB per [official docs recommendations](https://docs.nvidia.com/cuda/cublas/#cublassetworkspace).

When compiling we now add `-lcublasLt`:

```bash
nvcc -O3 --use_fast_math -Xcompiler -fopenmp matmul_forward.cu -o matmul_forward -lcublas -lcublasLt
```

And then these are the timings I am seeing on my A100 40GB PCIe:

```
(pytorch2) ubuntu:~/llm.c/dev/cuda$ ./matmul_forward 2
Using kernel 2
4.452381 4.452382
1.007628 1.007629
0.703605 0.703604
-2.559672 -2.559671
-1.964475 -1.964474
Results match at block_size=1024!
sqrt_block_size    4 | time 459.379211 ms | tflops 8.414553
sqrt_block_size    8 | time 283.652100 ms | tflops 13.627505
sqrt_block_size   16 | time 272.855042 ms | tflops 14.166756
sqrt_block_size   32 | time 279.780365 ms | tflops 13.816090
(pytorch2) ubuntu:~/llm.c/dev/cuda$ ./matmul_forward 3
Using kernel 3
4.452381 4.452382
1.007628 1.007629
0.703605 0.703604
-2.559672 -2.559671
-1.964475 -1.964474
Results match at block_size=1024!
sqrt_block_size    4 | time 298.320618 ms | tflops 12.957438
sqrt_block_size    8 | time 229.404678 ms | tflops 16.850008
sqrt_block_size   16 | time 231.020538 ms | tflops 16.732151
sqrt_block_size   32 | time 231.165955 ms | tflops 16.721626
(pytorch2) ubuntu:~/llm.c/dev/cuda$ ./matmul_forward 4
Using kernel 4
4.452381 4.449594
1.007628 1.007989
0.703605 0.705707
-2.559672 -2.559314
-1.964475 -1.965181
Results match at block_size=1024!
sqrt_block_size    4 | time 65.471039 ms | tflops 59.040924
sqrt_block_size    8 | time 65.196030 ms | tflops 59.289970
sqrt_block_size   16 | time 65.218559 ms | tflops 59.269489
sqrt_block_size   32 | time 58.589184 ms | tflops 65.975838
```

Wow, the TF32 version flies! This is a ~4X improvement! 

Note: I had to increase the correctness check threshold from 1e-4 to 1e-1 because TF32 is more approximate than I realized.

Question 1: The new kernel is block size independent so why does `sqrt_block_size=4` seem to be slower than the rest, for kernel 3? This looks wrong I'm not sure I get it.

Question 2: I'm hardcoding `has_gelu = 0` to not fuse GeLU. My concern is that I'm not sure we can do this without compromising the backward pass later. We may need the pre-GeLU activations? Not sure if GeLU can be backward in place.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,10 +53,10 @@ test_gpt2: test_gpt2.c
 
 # possibly may want to disable warnings? e.g. append -Xcompiler -Wno-unused-result
 train_gpt2cu: train_gpt2.cu
-	nvcc -O3 --use_fast_math $< -lcublas -o $@
+	nvcc -O3 --use_fast_math $< -lcublas -lcublasLt -o $@
 
 test_gpt2cu: test_gpt2.cu
-	nvcc -O3 --use_fast_math $< -lcublas -o $@
+	nvcc -O3 --use_fast_math $< -lcublas -lcublasLt -o $@
 
 clean:
 	rm -f train_gpt2 test_gpt2 train_gpt2cu test_gpt2cu
--- a/dev/cuda/matmul_forward.cu
+++ b/dev/cuda/matmul_forward.cu
@@ -13,9 +13,6 @@ OMP_NUM_THREADS=32 ./matmul_forward 2
 
 version 3 calls cuBLASLt, should be even faster
 OMP_NUM_THREADS=32 ./matmul_forward 3
-
-version 4 calls cuBLASLt with TF32, should be even even faster
-OMP_NUM_THREADS=32 ./matmul_forward 4
 */
 
 #include <stdio.h>
@@ -52,6 +49,9 @@ void cublasCheck(cublasStatus_t status, const char *file, int line)
 // cuBLAS workspace. Hardcoding to 32MiB but only Hopper needs 32, for others 4 is OK
 static size_t cublaslt_workspace_size = 32 * 1024 * 1024;
 static void* cublaslt_workspace = NULL;
+static cublasComputeType_t cublas_compute_type;
+cublasHandle_t cublas_handle;
+cublasLtHandle_t cublaslt_handle;
 
 // ----------------------------------------------------------------------------
 // CPU code reference
@@ -137,8 +137,6 @@ void matmul_forward2(float* out,
                      float* inp, float* weight, float* bias,
                      int B, int T, int C, int OC,
                      const int sqrt_block_size) {
-    cublasHandle_t handle; // cuBLAS context
-    cublasStatus_t stat = cublasCreate(&handle); // initialize CUBLAS context
     // for reference API is:
     // cublasStatus_t cublasSgemm(cublasHandle_t handle,
     //                        cublasOperation_t transa, cublasOperation_t transb,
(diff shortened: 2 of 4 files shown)


ademeure:
Looks great!

Q1) Did a few quick tests on my RTX 4090 and I think it's a combination of DVFS (clock frequency taking a bit to increase + thermal throttling) and the very 1st kernel launch being slower than average. After running `sudo nvidia-smi -lgc 1455,1455` (and adding an extra outer loop to run everything 2x), I see much more stable results:

> sqrt_block_size    4 | time 136.113480 ms | tflops 28.398882
sqrt_block_size    8 | time 136.060928 ms | tflops 28.409849
sqrt_block_size   16 | time 136.058853 ms | tflops 28.410284
sqrt_block_size   32 | time 136.154114 ms | tflops 28.390408
sqrt_block_size    4 | time 136.085510 ms | tflops 28.404718
sqrt_block_size    8 | time 136.040314 ms | tflops 28.414156
sqrt_block_size   16 | time 136.059906 ms | tflops 28.410063
sqrt_block_size   32 | time 136.049667 ms | tflops 28.412201

But if I reduce repeat_count to 1, the very 1st iteration is still slower:

> sqrt_block_size    4 | time 1.587360 ms | tflops 24.351566
sqrt_block_size    8 | time 1.482752 ms | tflops 26.069569
sqrt_block_size   16 | time 1.472512 ms | tflops 26.250858
sqrt_block_size   32 | time 1.494016 ms | tflops 25.873020
sqrt_block_size    4 | time 1.467392 ms | tflops 26.342455
sqrt_block_size    8 | time 1.472512 ms | tflops 26.250858
sqrt_block_size   16 | time 1.469440 ms | tflops 26.305740
sqrt_block_size   32 | time 1.483776 ms | tflops 26.051577

Ahhh, the joys of GPU benchmarking! :)

Q2) You're right, cuBLASLt has a special mode to output both the pre-GELU and post-GELU outputs: `CUBLASLT_EPILOGUE_GELU_AUX` which is what I should have used.

But I noticed today in Nsight Compute that cuBLASLt doesn't actually merge the GELU anyway, there's an extra "cublasLt::epilogue" kernel for the GELU case. I'm not sure whether the GELU_AUX path is any more fused than the regular GELU though... if not, then there's be no point using it (at least until NVIDIA fixes this!)

ademeure:
Also is it a good idea to hardcode 4MiB if NVIDIA recommends 32MiB for H100? It's possible to detect H100 by checking whether the SM major version is >= 9 using: 

`cudaDeviceProp deviceProp;
cudaGetDeviceProperties(&deviceProp, dev);
if (deviceProp.major >= 9) { ... }`

But that's possibly a bit complicated to save 28MiB, so the other option would be to hardcode it to 32MiB.

karpathy:
Oh interesting, I understand the first iteration being slower, but we are repeating the kernel 100 times. Shouldn't it roughly average out for the last 99 into something stable?

+1 on hardcoding to 32MiB for this, as the 28MiB delta doesn't seem worth going into splicing and dicing the specific GPU configuration for now.

karpathy:
We are now at 59ms/iter. PyTorch comparison (with tf32, and torch compile) is 25ms. So this is difference of 2.36X, still a lot but down from 4X from 2 days ago.

PR #100, opened 2024-04-13
Use cudaHostMalloc for inputs/targets and cpu_losses

ademeure:
Unfortunately the performance gain is ~noise because the amount of data to transfer is currently negligible, but it's theoretically still the "right" way to do it performance-wise. Also moves the malloc/free cpu_losses out of the main loop.

See: https://developer.nvidia.com/blog/how-optimize-data-transfers-cuda-cc/


karpathy:
LGTM 👍

Issue #19, opened 2024-04-09
Error: must forward with targets before backward


lizhipengpeng:
![image](https://github.com/karpathy/llm.c/assets/60966176/a151ac58-5174-4a86-9988-ee1593a51c7a)

when i ./train_gpt2.c , i get the result above Error.

ent0n29:
did you check if you have OpenMP installed?

```bash
# on macOS
$ brew install libomp
# on ubuntu
$ sudo apt-get install libomp-dev
```
after installing build it again with `make train_gpt2`, and tune the `OMP_NUM_THREADS=8` based on how many cores your cpu have.

ent0n29:
we have this error when the mean_loss is = -1.0f which is the initial value, and this attribute is only updeted in the gpt2_forward function when the target param is not NULL. So somewhere in the training loop or in the dataloader gpt2_backward function is called before than the gpt2_forward.

aircwt:
I got the same case.
Edit the Makefile, remove -Ofast flag as below:
CFLAGS = -O3 -Wno-unused-result

ent0n29:
#21 solves this

karpathy:
Does anyone understand how -O3 can possibly be causing this issue?

ent0n29:
-Ofast already enables all the optimizations from -O3 and also other aggressive optimizations, maybe it's better to either use -O3 or -Ofast, not both?

ent0n29:
@karpathy or maybe the optim could be custom made llvm passes targeting train_gpt2 and not automatically generated from the -O flags

ent0n29:
The -Ofast flag enables the `-ffast-math` flag.
The -ffast-math flag allows non-finite values and it can also ignore some mathematical rules that dictate that certain optimizations are not allowed.
It's possible that -Ofast and -ffast-math are causing errors in the floating point arithmetic

ent0n29:
Ok I guess I found the problem, now -Ofast and -O3 enabled work togheter, the issue was caused by -ffast-math being enabled messing up with floating point arithmetic.
This will work also:

`CFLAGS = -O3 -Ofast -fno-fast-math -Wno-unused-result`

Infatoshi:
I still get the "backward before forward" error when using `CFLAGS = -O3 -Ofast -fno-fast-math -Wno-unused-result`. I'm not a systems guy but omitting fast math seemed to work. Try changing to the following at the top of the make file:
```CC ?= clang
[...]

PR #101, opened 2024-04-13
Building on Windows

azret:
Adding **build_msvc.bat** similar to [llama2.c](https://github.com/karpathy/llama2.c/blob/master/build_msvc.bat)

Note that #19 is reproducible on Windows as well if /fp:fast is included.


karpathy:
This would need more explanation.

E.g. we're not using `clock_gettime`?

Also there's more changes, e.g. gitignore, and a change from const int to define.

I'd be interested in the minimal change needed to compile on Windows

ross-wheeler:
@karpathy I would humbly suggest using PR #118? There are no changes need to the source to compile and test on Windows. Thanks.

PR #107, opened 2024-04-13
warnings in train_gpt2.cu

zocterminal:


zocterminal:
Fixed a few warnings (debian 11 with cuda 12.0) in train_gpt2.cu:
- unused convenience variables (put them in comments)
- non-const char pointers to string literals
- a "%d" in a printf format that had no counterpart in the parameters

karpathy:
Thanks. Actually the `backward` pass stub that had a bunch of issues is only there because I pushed it accidentally in a previous commit when I tried to change one of the kernels. Ah well ok.

PR #108, opened 2024-04-13
Improved getting started from scratch

wbic16:
I had a new laptop, without any python tooling installed yet, and ran into the problems fixed by running this install script. Hopefully this helps more people get started with llm.c!


karpathy:
Looks a bit odd, there are requirements.txt already?

PR #109, opened 2024-04-13
Extracted common functions for CUDA dev

ngc92:
Extracts common parts out of the individual cuda kernel test files.
Makes it easier to write consistent tests (e.g., outputting time per repeat) and also some fixed, e.g., in some places, the old code was not properly free'ing malloced buffers; or claimed to test each block size, but in fact only tested one. 

Some challenges:
* When fusion the scale part into the attention kernel, we get different (unscaled) preatt results than the CPU kernel. Currently, this comparison is commented out. Similarly, checking the attention map doesn't seem to work.

Edit:
The error below was due to the test code not properly resetting the output memory. This is very important for testing backward, since there, we're generally adding instead of overwriting.

* With the new tests, we're testing different block sizes that weren't tested before, and I now get a failing test for cross-entropy backward:
> Mismatch of dlogits at 102926336: -0.041662 vs -0.083324

### DIFF
--- a/dev/cuda/attention_forward.cu
+++ b/dev/cuda/attention_forward.cu
@@ -39,7 +39,7 @@ uses a directly autoregressive softmax, and uses the online softmax algorithm.
 // CPU code reference
 
 void attention_forward_cpu(float* out, float* preatt, float* att,
-                       float* inp,
+                       const float* inp,
                        int B, int T, int C, int NH) {
     // input is (B, T, 3C) Q,K,V
     // preatt, att are (B, NH, T, T)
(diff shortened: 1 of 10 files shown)


karpathy:
wow, meaty one :) agree that i think this makes sense at this point...

karpathy:
@ngc92 it sounds like this PR is still in flux (per your edit comment etc.), will wait to full review until you think it looks ready.

ngc92:
I think it should be ready now. 
The remaining issues are not really a concern of the PR per-se, but more something  we need to keep in mind going forward: 
We may not be able to arbitrarily mix-and-match forward and backward kernel implementations for the same op, because they might want to store different intermediate results.

The TODO comments in this PR are thus less about something actually missing here, and more that I wanted to bring this to your attention.

karpathy:
made a few small changes/fixes and merged ty 👍

yuanrongxi:
nice

PR #112, opened 2024-04-13
HIP support

Avicted:
Generated with Hipify:
```bash
cd dev/hip
hipify-clang -I/opt/rocm-6.0.0/include/ ../cuda/crossentropy_forward.cu -o crossentropy_forward.hip.cpp
```

Compile example:
```bash
hipcc -O3 crossentropy_forward.hip.cpp -o crossentropy_forward
```

### DIFF
--- a/dev/hip/README.md
+++ b/dev/hip/README.md
@@ -1,0 +1,3 @@
+# dev/hip
+
+This directory is scratch space for developing HIP versions of the CUDA kernels.
--- a/dev/hip/common.h
+++ b/dev/hip/common.h
@@ -1,0 +1,129 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <hip/hip_runtime.h>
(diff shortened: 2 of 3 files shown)


karpathy:
I'd like to keep this code C/CUDA, but I'm happy to link to other ports in README.

ademeure:
I think even if we don't it'd be interesting to try to keep at least most things working with AMD's HIPIFY: https://github.com/ROCm/HIPIFY - at the very least, that would make maintaining a port easier (and potentially could justify including a script that runs HIPIFY in the main repo, but then that'd require regular testing, so it might still be too much work).

The only thing we have right now that definitely wouldn't work is cuBLASLt, but it's on their short-term roadmap AFAIK. @Avicted have you tried running HIPIFY on the other files, especially train_gpt2.cu? (ideally an older commit before the cuBLASLt integration)

PR #113, opened 2024-04-13
Fix cuBLAS to use TF32 with pre-initialised handle

ademeure:
Large perf gain as TF32 was only enabled for cuBLASLt and not cuBLAS by mistake.

Performance on A100:
- PyTorch: 23.6ms (version 2.1 so probably a bit faster on 2.2)
- Before: 48.2ms
- After: 25.1ms

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -412,10 +412,9 @@ void matmul_forward_cublas(float* out,
     const int sqrt_block_size = 32;
 
     cublasHandle_t handle; // cuBLAS context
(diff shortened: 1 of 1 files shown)


karpathy:
Gosh this is embarrassing, I missed this when I was refactoring the cublas handles. I'm on a box atm that doesn't have nvprof/ncu/etc so I'm "flying dark", I would have noticed this if I was profiling the code like a serious person. Have to fix, ok ty

PR #117, opened 2024-04-14
WIP: Fully fused classification layer

ngc92:
This fuses together all the pointwise operations that happen in the token classification layer. This essentially gives us the forward/backward for the cost of about just the forward pass, because this is entirely limited by memory transfers, and we save storing and loading the intermediate probabilities. 

Comes at the cost of losing a bit of flexibility regarding `dlosses`, because now, we already need to know  what we want to backprop. Shouldn't be much of a problem, I think, as long as  we're only considering averages of token-wise losses, with potential predetermined masking.


karpathy:
I like it, will incorporate into mainline shortly too.

PR #122, opened 2024-04-14
fix #114

### LINKED ISSUE #114: CUDA ERROR an illegal memory access was encountered
commit [`c02bae2ebc684a2e068c0dc59be00ff43167b44d`
](https://github.com/karpathy/llm.c/commit/c02bae2ebc684a2e068c0dc59be00ff43167b44d)
I got this error when running `train_gpt2cu`
```
[System]
Device 0: NVIDIA GeForce RTX 3060
enable_tf32: 1
[GPT-2]
max_seq_len: 1024
vocab_size: 50257
num_layers: 12
num_heads: 12
channels: 768
num_parameters: 124439808
train dataset num_batches: 74
val dataset num_batches: 8
batch size: 4
sequence length: 1024
val_num_batches: 10
num_activations: 2456637440
val loss 4.517170
step 0: train loss 4.367660 (took 140.356816 ms)
step 1: train loss 4.406325 (took 140.393840 ms)
step 2: train loss 4.484698 (took 140.358579 ms)
step 3: train loss 4.345248 (took 140.415392 ms)
step 4: train loss 4.043498 (took 140.366040 ms)
step 5: train loss 4.229475 (took 140.349640 ms)
step 6: train loss 4.175006 (took 140.337543 ms)
step 7: train loss 4.207677 (took 140.347408 ms)
step 8: train loss 4.127449 (took 140.839560 ms)
step 9: train loss 4.220684 (took 141.022061 ms)
val loss 4.517170
step 10: train loss 4.345275 (took 140.932235 ms)
step 11: train loss 4.245928 (took 141.035935 ms)
step 12: train loss 4.160665 (took 140.851916 ms)
step 13: train loss 3.989787 (took 140.892920 ms)
step 14: train loss 4.306040 (took 140.957500 ms)
step 15: train loss 4.340417 (took 141.032707 ms)
step 16: train loss 4.304399 (took 141.010662 ms)
step 17: train loss 4.424002 (took 140.930450 ms)
step 18: train loss 4.314507 (took 141.028123 ms)
step 19: train loss 4.287285 (took 140.992381 ms)
val loss 4.517170
[CUDA ERROR] at file train_gpt2.cu:1211:
an illegal memory access was encountered
```

### `nvidia-smi`
```
(.venv) ➜  llm.c git:(dev) ✗ nvidia-smi
Sun Apr 14 05:29:02 2024       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 550.54.15              Driver Version: 550.54.15      CUDA Version: 12.4     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 3060        Off |   00000000:01:00.0 Off |                  N/A |
|  0%   32C    P8             13W /  170W |       3MiB /  12288MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
                                                                                         
+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI        PID   Type   Process name                              GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
```

### `nvcc`
```
(.venv) ➜  llm.c git:(dev) ✗ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2024 NVIDIA Corporation
Built on Thu_Mar_28_02:18:24_PDT_2024
Cuda compilation tools, release 12.4, V12.4.131
Build cuda_12.4.r12.4/compiler.34097967_0
```

The file `train_gpt2.cu:1211:` indicates:
```C
cudaCheck(cudaMemcpy(cpu_probs, probs, model.config.vocab_size * sizeof(float), cudaMemcpyDeviceToHost));
```

ngc92:
In order to ensure that the sequence length is divisible by four, we do some padding.
We need to make sure that the padded tokens actually do have a valid token-id, so that they don't cause any out-of-bounds access.
Also added some asserts as part of the forward call, so that we hopefully can avoid similar problems in the future (or at least, they will be much easier to trace).

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -904,6 +904,11 @@ void gpt2_forward(GPT2 *model, int* inputs, int* targets, int B, int T) {
         }
     }
 
+    // validate inputs
+    for(int i = 0; i < B * T; ++i) {
+        assert(inputs[i] < V);
+    }
+
     // copy inputs/targets to the model
     cudaCheck(cudaMemcpy(model->inputs, inputs, B * T * sizeof(int), cudaMemcpyHostToDevice));
     if (targets != NULL) {
@@ -1224,7 +1229,11 @@ int main() {
 
         // once in a while do model inference to print generated text
         if (step > 0 && step % 20 == 0) {
-            gen_tokens[0] = GPT2_EOT; // the GPT-2 EOT token kicks off the generation
+            // the GPT-2 EOT token kicks off the generation
+            for(int i = 0; i < gen_max_length; ++i) {
+                gen_tokens[i] = GPT2_EOT;
+            }
+
             for (int t = 1; t < gen_max_length; t++) {
                 // note that inference is wasteful here because
                 // for each t, we re-compute all activations between 0 and t


msharmavikram:
I'm yet to check this fix but this will be slow in terms of perf. Wonder if this check can be added in the cuda kernel and assume artificial padding.

ngc92:
why would it be slow? It doesn't actually change anything inside the training loop, just the autoregressive test generation, which is  _atrocious_ in terms of performance anyway, because there is no caching  of previous tokens.

so the only thing that changes are  the additional asserts, which go away if you  compile with -DNDEBUG (which you should if you care about performance).

ngc92:
I agree that we might want to place the check inside the kernel, if this turns out to have significant performance impact on debug builds.

msharmavikram:
I think we can move this out of the for loop. The padding doesn't require to be done for each step. It needs to be done only once.

tnorman42:
This fix works for me. The issue is reproducible to me when any of the entries of gen_tokens are >= V (50257) due to not being initialized. Moving the initialization for loop to just after gen_tokens declaration (so it only needs to run once) also seems to solve the issue.

zocterminal:
Works for me.

A quicker version would probably be 

```
int gen_tokens[gen_max_length] = { 0, };
```

ngc92:
> Works for me.
> 
> A quicker version would probably be
> 
> ```
> int gen_tokens[gen_max_length] = { 0, };
> ```

I would argue strongly against that. Now I need to look up at cppreference whether this syntax  actually intializes the entire array, or just the first element. I need to check  if 0 == GPT2_EOT, which is no longer the obvious starting token in the code.

Finally, I don't think this is an operation that needs to be optimized anyway. We're running **quadratic time** inference for the  LLM here, currently, reprocessing all past tokens over and over again. Even if we fixed that, this is an extremely small operation (we're writing 64 ints to memory, nothing more!)

zocterminal:
It will initialize the whole array with zeros. The 0'th element will still be set to GPT2_EOT as in the current code line 1227 (GPT2_EOT is 50256).  It just avoids that the first call receives an array with incorrect values (after that it will be filled with values from the previous step%20 run).

A slightly more intuitive version would probably be:
```
int gen_tokens[gen_max_length] = { GPT2_EOT, 0, };
```

But this is just academic, bc yes I know that the loop you added is trivial in terms of time (based on how often that part is actually executed and based on the fact that it does a ton of printf). It would be just slightly cleaner code IMO.

ngc92:
Yes, it initializes with zero, but to me, this code always reads highly nonintuitively. Why would you mix explicit and implicit initialization when the implicit values are exactly the same as the explicit ones.
Why not write 
```
int gen_tokens[gen_max_length] = { GPT2_EOT, };
```
so that all zeros are implicit?

zocterminal:
Yes, agree.

PR #123, opened 2024-04-14
output of generated tokens as text via python tiktoken script

zocterminal:
Hacky function that takes the gen_tokens array that is created at every 20th step and generates a python script that will output the tokens as words via tiktoken (as demonstrated in the REDME.md).

For each of those steps it creates a `tiktoken_step_xx.py` file and prints a message that it can be used to show the output as words.

Updated .gitignore accordingly.


karpathy:
haha! too hacky. I think we should just do the correct thing. There is another PR for it, let me take a look :)

karpathy:
Done! :) see c165855cd331ac16de684e6d15507eabdc2f1a8d
closing

zocterminal:
> Done! :) see [c165855](https://github.com/karpathy/llm.c/commit/c165855cd331ac16de684e6d15507eabdc2f1a8d) closing

Mighty impressive ... awesome!

PR #124, opened 2024-04-14
Warnings about unused variables in test_gpt2.cu

zocterminal:
three unused variables moved into comments.
One unused array deleted.


zocterminal:
Closed this (this was a messy branch because I had accidentally checked files into it).

PR #126, opened 2024-04-14
Fix for resources going out of bound error, cleaning up warnings during fread call and removed unused variables. 

msharmavikram:
No change in correctness and performance. 

Before: 
```
step 38: train loss 4.941812 (took 25.920061 ms)
step 39: train loss 4.841427 (took 25.931963 ms)
val loss 5.121042
```

after (slightly faster but i am ignoring it):
```
step 38: train loss 4.941812 (took 25.774147 ms)
step 39: train loss 4.841427 (took 25.811688 ms)
val loss 5.121042
```


zocterminal:
I think if there are read errors we should also exit.

Alternately I believe you can get rid of the warnings by just casting the fread to void, to indicate to the compiler, that you're not interested in the return code:
`(void)fread(state_header, sizeof(int), 256, state_file);`

msharmavikram:
Added exit conditions. Voiding is not good practice.

train_gpt2.cu
```diff
@@ -425,7 +426,7 @@ void encoder_forward(float* out,
 void layernorm_forward(float* out, float* mean, float* rstd,
                        float* inp, float* weight, float* bias,
                        int B, int T, int C) {
-    const int block_size = 1024;
+    const int block_size = 256;
```
karpathy, line 429: You changed the block_size on layernorm

msharmavikram:
I will reopen this after i fix the merge conflict issue.

train_gpt2.cu
msharmavikram: I was previously running into resource unavailable error due to limited registers. To address this, I had reduced the block_size to 256.  However, I do not see this error in the current top of the tree. I am going to close this PR and raise a new PR that resolves only the warnings.

PR #128, opened 2024-04-14
Test updates

ngc92:
Several changes to the tests:
* Fixes printed timings, some were divided by the repeats twice
* Ensure that every kernel prints a more relative statistic (bandwidth, flops, time per token), and adjust number of digits so we don't spam meaningless decimals
* changed the softmax tests to use  `vocab_dim`  as the number of channels

The last one has several important implications:
  First, timings between the kernels might be changed now.
  More importantly, the old tests become meaningless, because with uniform logits, we get tiny probabilities for  each  output, and a kernel that just outputs zeros might actually pass the tests.
 This can  be fixed  by adding a few "outliers" to the input. With this, even the current kernels fail the tests.
 However, this is not because the current kernels are bad---In fact, by calculating partial sums in each thread, they are numerically more robust than the CPU implementation.  I've therefore changed the reference code to use  ` double`  for its sum accumulator, so that we can be assured that the ground-truth we're testing against is actually accurate.

### DIFF
--- a/dev/cuda/crossentropy_forward.cu
+++ b/dev/cuda/crossentropy_forward.cu
@@ -133,7 +133,7 @@ int main(int argc, char **argv) {
                                               kernel_num, d_out, d_probs, d_targets,
                                               B, T, V, block_size);
 
-        printf("block_size %4d | time %f ms\n", block_size, elapsed_time);
+        printf("block_size %4d | time %.4f ms | per token %.2f ns\n", block_size, elapsed_time, elapsed_time * 1'000'000 / (B*T));
     }
 
     // free memory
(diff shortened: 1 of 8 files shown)


karpathy:
These look nice! I'll merge when it's ready. It looks like it's still being iterated on?

ngc92:
I just checked all the existing kernels, and kernel 1 no longer passes  the stricter tests
what shall we do?  Remove it? Or use double for the sum here, too. That's probably not as bad for performance as you'd think, because the kernel spents almost all of its time waiting for memory, so even if you have only 1/64th of the flops available, that won't matter.

karpathy:
Yeah we should do the accumulates in higher precision imo. Agree RE performance hit being ~zero

ngc92:
OK, I've tested it, and on my RTX 4060 the difference is within the measurement noise. 
It seems there are some problems with other kernels, too.  The good news is, the fast ones, 6 and 7, do pass the tests. 
But kernel 2, for some reason, works for a block size of 32, but fails for 64.

karpathy:
😱 the race condition. beautiful find, merging.

Issue #94, opened 2024-04-12
Suggested to add a check for the return value of Malloc


dududuguo:
test_gpt2.c:55-58
train_gpt2.c:484 \531\ 634
Many implementations of malloc should include a checker to verify successful memory allocation. Suggested to add NULL checks after each malloc call and handle the memory allocation failures gracefully.
Thank you!

karpathy:
Fixed!

PR #129, opened 2024-04-15
Include none null assert for malloc

lancerts:
1. fix [issue](https://github.com/karpathy/llm.c/issues/94). Include an assert for malloc as a minimum check (do not intend to write a robust malloc with free and error check, etc. as it is not the focus of this repo)
2. Remove a duplicate `#include <assert.h>` in `train_gpt2.cu`.


karpathy:
need for this is now deprecated, use mallocCheck

PR #131, opened 2024-04-15
Cleaning up warnings during fread call and removed unused variables. (updated #126)

msharmavikram:
Making compilation a bit more reading-friendly.

### DIFF
--- a/test_gpt2.cu
+++ b/test_gpt2.cu
@@ -1,5 +1,6 @@
 #define TESTING
 #include "train_gpt2.cu"
+#define MODEL_HEADER 256
 
 // poor man's tensor checker
 int check_tensor(float *a, float *b, int n, char* label) {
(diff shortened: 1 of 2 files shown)


msharmavikram:
@karpathy can we merge this?

karpathy:
It looks really wordy. I want to think through how to make it nicer/shorter, maybe with macros or something.

msharmavikram:
Ok. Would something of this sort be useful?
```
#define HANDLE_ERROR(cond, msg) \
    do { \
        if (cond) { \
            fprintf(stderr, "%s\n", msg); \
            exit(EXIT_FAILURE); \
        } \
    } while(0)

#define CHECK_FREAD(ptr, size, count, stream) \
    HANDLE_ERROR(fread(ptr, size, count, stream) != count, \
        "Failed to read data: expected " #count " items.")
```

Example usage: 
```CHECK_FREAD(state_header, sizeof(int), 256, state_file);```

this would also create a mechanism to handle other errors in the same HANDLE_ERROR defn. If this is good, I would make the change and push.

karpathy:
ok here is what I came up with

https://github.com/karpathy/llm.c/commit/062ba5a30ee0d347355e31106f411af9fffbc5ce

fopenCheck, freadCheck, fcloseCheck, analogous to e.g. cudaCheck. Swap and replace, identical signatures.

Closing this issue as I think it is fixed by this commit, at least for the .cu file.

Next, we should probably refactor these common utilities to so both .c and .cu can use them.

PR #118, opened 2024-04-14
Visual Studio 2022 support - More const and some VS (Windows) build artifacts to ignore

ross-wheeler:
Adding Visual Studio build artifacts into the ignore file and adding in a few more const so Visual Studio can build these files. 2nd PR will have the Visual Studio (Windows) build changes. No source file changes needed after adding these const keywords in.

PR #134, opened 2024-04-15
Minimal changes to support Windows / Visual Studio

rosslwheeler:
const keyword additions and one new file in the platform/windows directory called unistd.h.


rosslwheeler:
@karpathy - this should be the minimal set of changes. Thanks.

karpathy:
Do we not have to include it?
Do we need to change the .cu as well?

rosslwheeler:
@karpathy - We'll need to include the fake unistd.h somewhere - I can change the makefile to create it "on the fly". We can add the fake one now and I can make the makefile change and delete it once those makefile changes are in?

The const keyword adds are (mostly/all?) the same as in the train_gpt2.c from a commit 2 days ago. We'll need these const in Windows because of a problem with OpenMP support. Adding these fixes it.

Makefile
```diff
@@ -35,29 +35,70 @@ else
...
-    $(warning OOPS Compiling without OpenMP support)
+    ifneq ($(OS), Windows_NT)
+      $(warning OOPS Compiling without OpenMP support)
+    endif
   endif
 endif
 
+ifeq ($(OS), Windows_NT)
+  CC := cl 
+  CFLAGS := /Iplatform\windows /Zi /nologo /W3 /WX- /diagnostics:column /sdl /Ox /Oi /Ot /GL /D _DEBUG /D _CONSOLE /D _UNICODE /D UNICODE /Gm- /EHsc /MD /GS /Gy /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /permissive- \
```
azret, line 46: @rosslwheeler the issue here is /fp:precise combined with /Ox. #19 is about this. What we want is /fp:fast with /Ox or /O2 to work without producing a nan.

rosslwheeler: @azret - open to suggestions on the optimizations - don't have a strong opinion on the compiler options. Let me know what you'd like to try here. Thanks.

dagelf:
@azret @rosslwheeler can either of you try just this single change around `gelu_backward` in `train_gpt2.c`?
```
#pragma float_control( precise, on, push ) // Save current setting and turn on /fp:precise

// code that needs more precision

#pragma float_control(pop) // Restore file's default settings
```

Seems like it might be the only required change.

dagelf:
In fact, first try to wrap only the `tanhf` call

Makefile
rosslwheeler: It's not failing - I think the issue per @azret was that we can tweak the optimizations better (faster). I'll need to look up the syntax to see what @azret is referring to above and give it a try.

rosslwheeler:
Sure - will report back. Thanks.

dagelf:
This `#pragma` is ignored by gcc, but breaks `clang`. I can add it, wrapped in yet another ifdef, if it makes `/fp:fast /Ox` work, without losing the performance gains in the other math calls. 

But check the performance of just wrapping `tanhf` vs wrapping the whole `gelu_backward`, and make sure to compile with `/fp:fast /Ox` - on `gcc` all the optimizations fell away when trying to wrap only `tanhf`. Not sure why, maybe `msvc` is more fine grained.

Makefile
dagelf: Basically, `train_gpt2` fails after the first step when compiled with `/fp:fast` (or `-ffast_math` on gcc). This should omit that optimization just for the function that causes the failure, while keeping the speedup for everything else.

dagelf: By "not failing" you mean it runs some training steps fine, even when compiled with `/fp:fast /Ox`?

azret:
**Test run 1**

**cl.exe /Ox /fp:fast /I. /I .\dev\win train_gpt2.c**

D:\Repos\llm.c>train_gpt2.exe
[GPT-2]
max_seq_len: 1024
vocab_size: 50257
num_layers: 12
num_heads: 12
channels: 768
num_parameters: 124439808
train dataset num_batches: 1192
val dataset num_batches: 128
num_activations: 73323776
val loss 5.252007
step 0: train loss 5.356174 (took 10266.000000 ms)
Error: must forward with targets before backward

**Test run 2** with 

```c
#pragma float_control( precise, on, push ) // Save current setting and turn on /fp:precise
void gelu_backward(float* dinp, float* inp, float* dout, int N) {
    for (int i = 0; i < N; i++) {
        float x = inp[i];
        float cube = 0.044715f * x * x * x;
        float tanh_arg = GELU_SCALING_FACTOR * (x + cube);
        float tanh_out = tanhf(tanh_arg);
        float coshf_out = coshf(tanh_arg);
        float sech_out = 1.0f / (coshf_out * coshf_out);
        float local_grad = 0.5f * (1.0f + tanh_out) + x * 0.5f * sech_out * GELU_SCALING_FACTOR * (1.0f + 3.0f * 0.044715f * x * x);
        dinp[i] += local_grad * dout[i];
    }
}
#pragma float_control(pop) // Restore file's default settings
```

**cl.exe /Ox /fp:fast /I. /I .\dev\win train_gpt2.c**

D:\Repos\llm.c>train_gpt2.exe
[GPT-2]
max_seq_len: 1024
vocab_size: 50257
num_layers: 12
num_heads: 12
channels: 768
num_parameters: 124439808
train dataset num_batches: 1192
val dataset num_batches: 128
num_activations: 73323776
val loss 5.252007
step 0: train loss 5.356174 (took 10156.000000 ms)
step 1: train loss 4.300983 (took 9938.000000 ms)
step 2: train loss 4.623276 (took 10109.000000 ms)
step 3: train loss 4.600286 (took 10172.000000 ms)
...

Wow! 10 seconds without OpenMP 👍 

Thanks @dagelf! Now convince Andrej to merge it in :)

azret:
Now re: **/openmp**:

```
D:\Repos\llm.c>cl.exe /Ox /fp:fast /openmp /I. /I .\dev\win train_gpt2.c
Microsoft (R) C/C++ Optimizing Compiler Version 19.39.33523 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

train_gpt2.c
train_gpt2.c(163): warning C4849: OpenMP 'collapse' clause ignored in 'parallel for' directive
train_gpt2.c(164): error C3015: initialization in OpenMP 'for' statement has improper form
train_gpt2.c(188): warning C4849: OpenMP 'collapse' clause ignored in 'parallel for' directive
train_gpt2.c(189): error C3015: initialization in OpenMP 'for' statement has improper form
train_gpt2.c(204): error C3015: initialization in OpenMP 'for' statement has improper form
train_gpt2.c(234): warning C4849: OpenMP 'collapse' clause ignored in 'parallel for' directive
train_gpt2.c(235): error C3015: initialization in OpenMP 'for' statement has improper form
train_gpt2.c(397): warning C4849: OpenMP 'collapse' clause ignored in 'parallel for' directive
train_gpt2.c(398): error C3015: initialization in OpenMP 'for' statement has improper form
```

(https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c3015?view=msvc-170)

Microsoft is always special. It wants the loops to be in this format. Notice the inline b and t were moved out.

```c
void matmul_forward(float* out,
                    float* inp, float* weight, float* bias,
                    int B, int T, int C, int OC) {
    // most of the running time is spent here and in matmul_backward
    // OC is short for "output channels"
    // inp is (B,T,C), weight is (OC, C), bias is (OC)
    // out will be (B,T,OC)
    int b;
    int t;
    #pragma omp parallel for collapse(2)
    for (b = 0; b < B; b++) {
        for (t = 0; t < T; t++) {
            float* out_bt = out + b * T * OC + t * OC;
            float* inp_bt = inp + b * T * C + t * C;
            for (int o = 0; o < OC; o++) {
                float val = (bias != NULL) ? bias[o] : 0.0f;
                float* wrow = weight + o*C;
                for (int i = 0; i < C; i++) {
                    val += inp_bt[i] * wrow[i];
                }
                out_bt[o] = val;
            }
        }
    }
}
```

After fixing all those

```
D:\Repos\llm.c>cl.exe /Ox /fp:fast /openmp /I. /I .\dev\win train_gpt2.c
Microsoft (R) C/C++ Optimizing Compiler Version 19.39.33523 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

train_gpt2.c
train_gpt2.c(165): warning C4849: OpenMP 'collapse' clause ignored in 'parallel for' directive
train_gpt2.c(194): warning C4849: OpenMP 'collapse' clause ignored in 'parallel for' directive
train_gpt2.c(242): warning C4849: OpenMP 'collapse' clause ignored in 'parallel for' directive
train_gpt2.c(407): warning C4849: OpenMP 'collapse' clause ignored in 'parallel for' directive
Microsoft (R) Incremental Linker Version 14.39.33523.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:train_gpt2.exe
train_gpt2.obj
```

Kaboom!

D:\Repos\llm.c>train_gpt2.exe
[GPT-2]
max_seq_len: 1024
vocab_size: 50257
num_layers: 12
num_heads: 12
channels: 768
num_parameters: 124439808
train dataset num_batches: 1192
val dataset num_batches: 128
num_activations: 73323776
**val loss inf**

I even removed the /Ox and /fp:fast

```
D:\Repos\llm.c>cl.exe /openmp /I. /I .\dev\win train_gpt2.c
Microsoft (R) C/C++ Optimizing Compiler Version 19.39.33523 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

train_gpt2.c
train_gpt2.c(165): warning C4849: OpenMP 'collapse' clause ignored in 'parallel for' directive
train_gpt2.c(194): warning C4849: OpenMP 'collapse' clause ignored in 'parallel for' directive
train_gpt2.c(242): warning C4849: OpenMP 'collapse' clause ignored in 'parallel for' directive
train_gpt2.c(407): warning C4849: OpenMP 'collapse' clause ignored in 'parallel for' directive
Microsoft (R) Incremental Linker Version 14.39.33523.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:train_gpt2.exe
train_gpt2.obj

D:\Repos\llm.c>train_gpt2.exe
[GPT-2]
max_seq_len: 1024
vocab_size: 50257
num_layers: 12
num_heads: 12
channels: 768
num_parameters: 124439808
train dataset num_batches: 1192
val dataset num_batches: 128
num_activations: 73323776
val loss inf
```

We're still not there for windows runs...

azret:
I removed parallelization on every layer except matmul_forward  and 7 seconds.

```
D:\Repos\llm.c>cl.exe /openmp:llvm /fp:fast /Ox /I. /I .\dev\win train_gpt2.c
Microsoft (R) C/C++ Optimizing Compiler Version 19.39.33523 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

train_gpt2.c
Microsoft (R) Incremental Linker Version 14.39.33523.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:train_gpt2.exe
train_gpt2.obj

D:\Repos\llm.c>train_gpt2.exe
[GPT-2]
max_seq_len: 1024
vocab_size: 50257
num_layers: 12
num_heads: 12
channels: 768
num_parameters: 124439808
train dataset num_batches: 1192
val dataset num_batches: 128
num_activations: 73323776
val loss 5.252007
step 0: train loss 5.356173 (took 7235.000000 ms)
step 1: train loss 4.300982 (took 7000.000000 ms)
step 2: train loss 4.623275 (took 7015.000000 ms)
step 3: train loss 4.600286 (took 8797.000000 ms)
step 4: train loss 4.616758 (took 10031.000000 ms)
step 5: train loss 4.231482 (took 9579.000000 ms)
```

Unfortunately enabling parallelization in attention and softmax errors out with inf. loss.

rosslwheeler:
@azret  when you say "errors out with inf. loss" what value metric are you talking about or is it crashing? Also, your compiler flags are still off a bit. Just an FYI. Not sure how it's building for you.

rosslwheeler:
@azret - adding /fp:fast definitely hangs it. But it runs fine about 3 seconds per step with /0x and the top of the trunk for me.

azret:
@rosslwheeler thank you very much. I have tried you project files. Very nice setup. They are very useful though when debugging. Unfortunately it's not about the projects files. There is only one file to build, really. You can just command line it.

**It is however about the compiler flags**. 

We want to match the performance on Windows in the same ballpark that they have on Mac and Linux. That means turning all runtime checking off. Turning all debug flags off. Turning all optimizations **on**. Turning fast floating points **on**. In addition, we want to parallelize as many layers as possible. So, when you finetune all these compiler flags on MSVC you get #19.
This is what we are trying to solve. And thanks to @dagelf we can now use /Ox +/fp:fast can just exclude the gelu layer see https://github.com/karpathy/llm.c/pull/134#issuecomment-2065415311

rosslwheeler:
@azret got it! Thanks for the clarity!

rosslwheeler:
@azret - adding /fp:fast with the pragma wrapper around gelu_backward took the OpenMP steps from ~3.5 seconds down to ~2.5 seconds. This is with zero code changes - about what you would expect based on your data above? Also, running WSL with GCC - would you expect it to be roughly the same speed as Windows (I know the compilers are different - just curious what you've seen)?

dagelf:
I don't think the ifdefs are too much... maybe we can even do a custom `tanhf` function to include. 

I have two questions about this pull request, 
1) Why do you want to create a header file in a Makefile rather than just adding it to the repo and including it where needed?
2) Does this pull actually fix OpenMP for `msvc` or is that still a work in progress?

rosslwheeler:
@dagelf 

1. Andrej was asking for **minimal** changes - discussion in https://github.com/karpathy/llm.c/pull/118 and check the top of this thread. I'm fine with a header in the repo or an auto-generated one.
2. OpenMP has been working in MSVC for about a week in my forks. I would not call it a fix per se - it's a work-around. I filed a bug report with Microsoft's compiler team today on it. That's why we need the const source fixes in the PR. Otherwise, all we would need is the Makefile changes at a minimum.

karpathy:
wow the diff on this PR. This is what is needed to compile C on Windows?

rosslwheeler:
@karpathy - we're running some compiler limitations with Windows.

1. OpenMP 5.2 - requires either the const changes OR changes in the actual C code to move some var initializers around. No other work-around that I'm aware of. I've filed a bug report with the MSVC team. So our choice is either fix the const's or fix the initializers in the OpenMP functions. 
2. unistd.h - Windows doesn't have this file. We can put unistd.h in dev? Would that work? The alternative is to auto-generate it in the makefile (that's what you see here in the PR).

Thoughts?

karpathy:
The const changes look good even as a general thing independent of Windows, I'd merge that regardless of Windows considerations.

The makefile echo I'm not a huge fan of.

Ideally we wouldn't introduce more files in the root of the repo. If there's any ways to hide them that could be a good idea.

As for where, I'm thinking of `dev` as a scratch space for dev. Maybe it makes sense to create a `platforms/win`, directory for any platform-specific code we might want and put things there.

rosslwheeler:
@karpathy - Not a fan of the Makefile header echo either. I'd rather move the header into dev or platform/win. Either works for me. Just let me know and I can generate a PR for this.

rosslwheeler:
Closing this - creating a document to cover building with different methods for Windows/MSVC.

PR #139, opened 2024-04-15
Easy Cloud Setup Script

dongreenberg:
Hi! Sometimes it's tricky or daunting to set up the hardware and environment for a script like this on a self-hosted cloud GPU. I was trying out llm.c on a fresh A10G on AWS, automating the setup and compilation via the library I work on (Runhouse), and figured I'd contribute it back in case you think it's useful to others. Feel free to close if not, or I can change to a notable fork.

Runhouse is built for rapid remote Python iteration, but the makefile actually takes care of this for C. If you make changes to the C code and rerun this script they'll be synced over and rebuilt quickly via the makefile. I'm also happy to change this script to strictly do the hardware setup in Python and then leave the user to compile and run in C via ssh.

### DIFF
--- a/cloud_gpt2_train.py
+++ b/cloud_gpt2_train.py
@@ -2,6 +2,7 @@ import runhouse as rh
 
 
 def train_cloud():
+    # This can be modified to use other instance types or other cloud providers.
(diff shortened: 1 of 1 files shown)


karpathy:
too specific to merge to master. possibly create a Discussion thread on instead.

PR #142, opened 2024-04-15
Update Makefile with -march=native for 25%+ CPU speedup

dagelf:
`-march=native` Results in a 25%+ speedup on all the platforms I've tried, for train_gpt2

```
model name      : Intel(R) Core(TM) i3-9100F CPU @ 3.60GHz
step 0: train loss 5.356172 (took 15197.876634 ms)
vs 
step 0: train loss 5.356185 (took 10418.548668 ms)

model name      : AMD Ryzen 5 3600 6-Core Processor
step 0: train loss 5.356173 (took 3599.359234 ms)
vs
step 0: train loss 5.356185 (took 2708.045790 ms)
```

### DIFF
--- a/train_gpt2.c
+++ b/train_gpt2.c
@@ -625,7 +625,7 @@ void gpt2_build_from_checkpoint(GPT2 *model, char* checkpoint_path) {
     model->param_sizes[14] = C; // lnfw
     model->param_sizes[15] = C; // lnfb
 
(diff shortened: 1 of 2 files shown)


karpathy:
doh, missed this flag when I was porting Makefile from llama2.c, nice find.
I don't get speedup on macOS but I do get the speedup on my linux box.
Thank you!

PR #143, opened 2024-04-15
enable padding for softmax computation and fix up the existing kernels

ngc92:
This doesn't help us as is, but going forward, its a first step towards padding the vocab dimension to a sane value that actually allows for fast implementations.

I haven't updated any of the softmaxes yet to actually make use of the fact that they're padded. I think the only kernel that would benefit directly from it (more on a LOC perspective than in terms of speed, I think), would be softmax7, as we could get rid of some of the ifs once we know ldd is a multiple of UNROLL_FACTOR.


ngc92:
rebased on  most recent master

ngc92:
superseded  by #264

PR #150, opened 2024-04-16
Optimised version of fused classifier + bugfixes(?)

ademeure:
This is a faster version of the cool new kernel from #117 (still /dev/cuda/ only). The biggest difference is it is optimised for doing one row per 1024-wide block rather than per 32-wide warp, which massively reduces the amount of data in-flight, and this allows us to basically always hit in the L2 cache for the 2nd read and this saves a huge amount of DRAM bandwidth. It also uses 128-bit loads which implicitly requires V to be a multiple of 4.

However, I believe the validation in classifier_fused.cu is currently not working correctly, possibly because the distribution of inputs results in all probabilities being ~zero, which is something I think ngc92 has a fix for.

Finally, I changed both old and new kernels to write to the gradient tensor, rather than modify it (read-modify-write) which costs us an extra DRAM access. I don't think accumulating gradients is required for this part of the network. All performance numbers below are with this change including Kernel1.

Performance:
Fused Kernel1 (old, block size invariant): ~6ms
Fused Kernel2 (new, 1024 threads per block): ~3.7ms
Forward Softmax Kernel7 (512 threads per block): 5.7ms (current default used in train_gpt2.cu)
Forward Softmax Kernel7 (1024 threads per block): ~3.7ms (optimal block size, got it wrong in my original PR for kernel7)

### DIFF
--- a/dev/cuda/classifier_fused.cu
+++ b/dev/cuda/classifier_fused.cu
@@ -154,33 +154,31 @@ __device__ float vec_at(const float4& vec, int index) {
     return reinterpret_cast<const float*>(&vec)[index];
 }
 
-__device__ SoftmaxParams prepare_softmax2(cg::thread_block_tile<32>& warp,
-                                         int idx, const float* inp, int V) {
+__device__ SoftmaxParams prepare_softmax_blockwide(cg::thread_block_tile<32>& warp,
+                                                   int idx, const float* inp, int V, int P) {
     // one row of inp, i.e. inp[idx, :] of shape (V,)
     // float4 to get 128-bit loads and memory level parallelism
-    // this is only possible if V is a multiple of 4
-    const float4* x_vec4 = reinterpret_cast<const float4*>(inp + idx * V);
+    const float4* x_vec4 = reinterpret_cast<const float4*>(inp + idx * P);
 
     float thread_maxval = -INFINITY;
     float thread_sumval = 0.0f;
     // do the loop in reverse to maximise probability of L2 cache hits
     // so even small L2s get some hits on the 2nd read of the same thread
-    for (int i = (V/4) + (threadIdx.x - blockDim.x); i >= 0; i -= blockDim.x) {
-        float4 v = x_vec4[i];
+    for (int i = (V+3)/4 + (threadIdx.x - blockDim.x); i >= 0; i -= blockDim.x) {
+        float4 v4 = x_vec4[i];
         #pragma unroll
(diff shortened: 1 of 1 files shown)


karpathy:
oops this PR now conflicts because I merged the other one. Sounds good, agree it is ok to skip += here, but I think it should come with a comment on this subtle issue. Please add a brief one, or I'm also ok adding after.

dev/cuda/classifier_fused.cu
```diff
@@ -169,6 +259,13 @@ void fused_classifier(int kernel_num, float* dlogits, float* losses,
         case 1:
             fused_classifier1(dlogits, losses, logits, dlosses, targets, B, T, V, block_size);
             break;
+        case 2:
+            if((V % 4) != 0) {
```
karpathy, line 263: I'd move this inside `fused_classifer2` at the top?

dev/cuda/classifier_fused.cu
```diff
@@ -139,15 +139,96 @@ __global__ void fused_classifier_kernel(float* dlogits, float* losses,
...
+    // this is only possible if V is a multiple of 4
+    const float4* x_vec4 = reinterpret_cast<const float4*>(inp + idx * V);
+
+    float thread_maxval = -INFINITY;
+    float thread_sumval = 0.0f;
+    // do the loop in reverse to maximise probability of L2 cache hits
+    // so even small L2s get some hits on the 2nd read of the same thread
+    for (int i = (V/4) + (threadIdx.x - blockDim.x); i >= 0; i -= blockDim.x) {
+        float4 v = x_vec4[i];
+        #pragma unroll
```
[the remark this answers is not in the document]
ademeure: Agreed, added a simple bounds check (even if we could theoretically memset it once and make sure nothing modifies it, not worth the risk of bugs...)

dev/cuda/classifier_fused.cu
ademeure: Removed it after ngc92's changes that split V and P (padded V) so it's implicitly guaranteed by the function declaration.

ademeure:
Should be ready to be merged again now (at least to /dev/cuda/)!

I made it more flexible so it can now also work as a forward-only option by passing a pointer to "probs" and making "dlogits" NULL instead.

Combined with the bounds checking it's making the code a bit longer than I'd like, and with a bit more overhead in terms of GPU instructions, but given it's massively DRAM limited it seems maybe worth it to only have one function doing everything?

Issue #106, opened 2024-04-13
[CUDA ERROR] at file \llm.c\train_gpt2.cu:405: too many resources requested for launch (old version does not have this issue - fyi)


ross-wheeler:
[System]
Device 0: NVIDIA RTX A5500
enable_tf32: 1
[GPT-2]
max_seq_len: 1024
vocab_size: 50257
num_layers: 12
num_heads: 12
channels: 768
num_parameters: 124439808
train dataset num_batches: 74
val dataset num_batches: 8
batch size: 4
sequence length: 1024
val_num_batches: 10
num_activations: 2456637440
[CUDA ERROR] at file \llm.c\train_gpt2.cu:405:
too many resources requested for launch

Switch back to the previous version?

msharmavikram:
Change the following lines

```
__global__  void layernorm_forward_kernel3(float* __restrict__ out, float* __restrict__ mean, float* __restrict__ rstd,
                                    const float*  __restrict__ inp, const float*  __restrict__ weight,
                                    const float* __restrict__ bias, int N, int C) {
```
to below
```
__global__ __launch_bounds__(1024,2) 
void layernorm_forward_kernel3(float* __restrict__ out, float* __restrict__ mean, float* __restrict__ rstd,
                                    const float*  __restrict__ inp, const float*  __restrict__ weight,
                                    const float* __restrict__ bias, int N, int C) {
```

I believe it is running out of registers. So need to forcefully do register spills.  This solved the error for me.

ross-wheeler:
Thank you, Vikram! That fixed it! Should this be a generic fix or is it card/gpu specific?

msharmavikram:
This may not work on all cards. I need to determine the correctness of this fix for different thread block size before submitting a PR. I didn't a chance to work on that yet.

msharmavikram:
Submitted a PR #126

msharmavikram:
@ross-wheeler I do not see this issue anymore on the top of the tree. Let me know if you still see the error without the proposed fix.

rosslwheeler:
@msharmavikram - confirmed - no issues now. Any idea what changed? I did see a lot of source changes.

rosslwheeler:
@msharmavikram - I still need your above fix. The top of tree did not fix my issue.

PR #151, opened 2024-04-16
simple fix for bug [#106]

msharmavikram:
@karpathy I have two solutions to fix the out-of-resource issue that some folks are observing (see https://github.com/karpathy/llm.c/issues/106). We are hitting this issue due to excessive register usage and the compiler is unable to determine it well. The two solutions to this are: 

a) keep the block_size = 1024 and add __launch_bounds__(1024, 2) as a prefix as shown below. This would force register spills to the memory subsystem. 

 ```
__global__ void layernorm_forward_kernel3(float* __restrict__ out, float* __restrict__ mean, float* __restrict__ rstd,
                                   const float* __restrict__ inp, const float* __restrict__ weight,
                                   const float* __restrict__ bias, int N, int C) {
```

to below 

```
__global__ __launch_bounds__(1024,2)
void layernorm_forward_kernel3(float* __restrict__ out, float* __restrict__ mean, float* __restrict__ rstd,
                                   const float* __restrict__ inp, const float* __restrict__ weight,
                                   const float* __restrict__ bias, int N, int C) {
``` 

This works great for GPUs with 2048 as their maximum resident number of threads. However, we some users (SM version: 8.6) have a maximum resident number of threads as 1536 and this solution will fail for them.  A potential alternative to keeping the block_size=1024 is reduced occupancy (up to 50% and at the cost of run time) using the below change 
```
__global__ __launch_bounds__(1024,1)
void layernorm_forward_kernel3(float* __restrict__ out, float* __restrict__ mean, float* __restrict__ rstd,
                                   const float* __restrict__ inp, const float* __restrict__ weight,
                                   const float* __restrict__ bias, int N, int C) {
``` 

I am not in favor of this either. 

b) Fundamental to the issue is we are launching too many resources and observing a launch failure. In this case, reducing the block size enables the compiler to optimize better. This is why I reduced the block size from 1024 to 512 (can be 256 too). Before I made this change, I checked for correctness and running `dev/cuda/layernorm_forward.cu`. block_size 256 or 512 makes about a 10% drop difference in performance (on an average run).  However, end-to-end layer time, I did not observe much difference! 

If you still want to keep block_size = 1024 then we need to add the `__CUDA_ARCH__` qualifier to make it conditioned for respective SM versions. Alternatively, we can think of ways to reduce register usage (not that easy given it uses cg.). (The last one is I need to check if there is a compiler flag that automatically allows register spill if it detects it. ) 

Let me know what you think!

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -594,7 +594,7 @@ void encoder_forward(float* out,
 void layernorm_forward(float* out, float* mean, float* rstd,
                        float* inp, float* weight, float* bias,
                        int B, int T, int C) {
-    const int block_size = 1024;
+    const int block_size = 512;
     const int N = B * T;
     const int grid_size = CEIL_DIV(N * 32, block_size);
     layernorm_forward_kernel3<<<grid_size, block_size>>>(out, mean, rstd, inp, weight, bias, N, C);


ngc92:
does the reduced occupancy on 1500-thread cards actually translate to any speed differences? This kernel is highly memory-bound anyway, so it might not actually matter.

msharmavikram:
It does not. We see about a `50us` difference between the two in the example kernel. So it is very tiny compared to 25ms.

karpathy:
Got it, thank you. I don't see these errors on my box so I appreciate your help / PRs to fix & document.

PR #163, opened 2024-04-17
Memory savings by reusing the same buffer across layers in backward

ngc92:
reuse memory buffers across layers during the backward pass

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -288,9 +288,9 @@ __global__ void permute_kernel_backward(float* dinp,
         int d_ = rest % d;
 
         int inp_idx = (b * N * 3 * NH * d) + (n * 3 * NH * d) + (0 * NH * d) + (nh_ * d) + d_;
-        dinp[inp_idx] += dq[idx];
-        dinp[inp_idx + NH * d] += dk[idx];
-        dinp[inp_idx + 2 * (NH * d)] += dv[idx];
+        dinp[inp_idx] = dq[idx];
+        dinp[inp_idx + NH * d] = dk[idx];
+        dinp[inp_idx + 2 * (NH * d)] = dv[idx];
     }
 }
 
@@ -323,7 +323,7 @@ __global__ void unpermute_kernel_backward(float* dinp, const float *dout, int B,
         int d_ = rest % d;
 
         int other_idx = (b * NH * N * d) + (n * NH * d) + (nh_ * d) + d_;
-        dinp[idx] += dout[other_idx];
+        dinp[idx] = dout[other_idx];
     }
 }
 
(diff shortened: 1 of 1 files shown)


train_gpt2.cu
```diff
@@ -601,12 +601,12 @@ __global__ void matmul_backward_bias_kernel_faster(float* dbias, const float* do
     }
     // write the final result (at thread 0) to global memory
     if (tid == 0) {
-        dbias[o] = shared[0];
+        dbias[o] += shared[0];
```
karpathy, line 604: every change is += -> =, except here = -> += ?

train_gpt2.cu
```diff
@@ -1533,23 +1567,25 @@ void gpt2_backward(GPT2 *model) {
...
-        float* dl_att = grads_acts.att + l * B * NH * T * T;
-        float* dl_v_accum = grads_acts.v_accum + l * B * T * C;
-        float* dl_attproj = grads_acts.attproj + l * B * T * C;
-        float* dl_residual2 = grads_acts.residual2 + l * B * T * C;
-        float* dl_ln2 = grads_acts.ln2 + l * B * T * C;
-        float* dl_fch = grads_acts.fch + l * B * T * 4*C;
-        float* dl_fch_gelu = grads_acts.fch_gelu + l * B * T * 4*C;
-        float* dl_fcproj = grads_acts.fcproj + l * B * T * C;
-        float* dl_residual3 = grads_acts.residual3 + l * B * T * C;
+        float* dl_ln1 = grads_acts.ln1;
```
karpathy, line 1570: wait if we're only doing it like this layer by layer, aren't we able to allocate much less for `grad_acts`? We don't need the multiplication by `L`?

train_gpt2.cu
```diff
@@ -1458,8 +1469,31 @@ void gpt2_backward(GPT2 *model) {
...
+        // allocate buffers for weight gradients
         model->grads_memory = malloc_and_point_parameters(&model->grads, model->param_sizes, 1);
-        model->grads_acts_memory = malloc_and_point_activations(&model->grads_acts, model->act_sizes);
+
+        // determine buffer size for activation gradients
+        // _mostly_ allocate as if we had just a single layer,
+        // because we're going to reuse memory
+        size_t bw_act_sizes[NUM_ACTIVATION_TENSORS];
+        GPT2Config cfg = model->config;
+        cfg.num_layers = 1;
```
karpathy, line 1480: ah, I just noticed this critical part. L = 1

train_gpt2.cu
ngc92: this is the bias kernel, right. So if we're doing gradient accumulation over multiple steps, we would want to add these up, right?
Technically, this isn't part of the memory changes, but when checking each kernel, I noticed this.

karpathy: But aren't we able to keep it as = like all the other kernels?

karpathy:
The danger here is that we have two forking points in our graph:

- residual layers create a fork
- `wte` has a fork - the same tensor is used both for token embeddings AND for final matmul

we have to make sure these two parts do a += correctly. I'm trying to make sure this happens in the code but it's not obvious yet.

karpathy:
1) ok `wte` will be ok because:
- matmul now does =
- but encoder backward still does += 
OK

Issue #147, opened 2024-04-16
use const properly, esp in function signatures


karpathy:
I wasn't diligent enough with my use of `float*` vs `const float*`. The former is used for outputs, the latter for inputs, to all functions. We'd like to refactor the function declarations to follow this paradigm, to be more defensive.

rosslwheeler:
This one?

int check_tensor(float *a, float *b, int n, const char* label) { this one generates a couple of warnings

test_gpt2.cu(139): warning #2464-D: conversion from a string literal to "char *" is deprecated
test_gpt2.cu(141): warning #2464-D: conversion from a string literal to "char *" is deprecated

There's also these but wasn't sure if there's some incoming changes to use these variables?

test_gpt2.cu(56): warning #177-D: variable "maxT" was declared but never referenced
test_gpt2.cu(57): warning #177-D: variable "L" was declared but never referenced
test_gpt2.cu(95): warning C4101: 'losses': unreferenced local variable

PR #166, opened 2024-04-17
const-fix crossentropy_softmax_backward.cu

lancerts:
https://github.com/karpathy/llm.c/issues/147


karpathy:
I'm deleting this file because of fused classifier, so closing

PR #169, opened 2024-04-18
Fix the bug that yields cpu, gpu results mismatch

lancerts:


karpathy:
we can't just malloc on repeat, without free. maybe memset to zero if needed?

lancerts:
> we can't just malloc on repeat, without free. maybe memset to zero if needed?

@karpathy good point... Always forget we are dealing with raw ptr instead of smart ptr : P. 
Uses the memset to 0 and the script executes all passed.

PR #172, opened 2024-04-18
fix(attention_forward.cu): Fix the number of elements to check

FeSens:
Fix the number of elements to check when verifying attention implementations.

This was causing me problems when I was trying to debug a new implementation on fewer tokens.


karpathy:
oops good find

PR #174, opened 2024-04-18
a little reorg :-p

zocterminal:
Reorganized source code to consolidate copied functions and structures:
- common stuff is now in train_common.c and train_common.h
- cuda version needs to define TRAIN_CUDA before including the header or the .c file.
- using the checkFopen(), mallocCheck() etc. functions in train_gpt2.c also.

It may be possible to consolidate the reading of the model a bit more by using ifdefs with the TRAIN_CUDA and TRAIN_CPU macros.

Enjoy.

### DIFF
--- a/train_common.c
+++ b/train_common.c
@@ -1,0 +1,279 @@
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <ctype.h>
+#include <math.h>
+#include <time.h>
+#include <assert.h>
+#include <float.h>
+#include <string.h>
+#include <unistd.h>
+#include <assert.h>
+#ifdef OMP
+#include <omp.h>
+#endif
+
+#ifdef TRAIN_CUDA
+#include <cublas_v2.h>
+#include <cuda_runtime.h>
+#include <cublasLt.h>
+#include <cooperative_groups.h>
+#include <cooperative_groups/reduce.h>
+#endif
+
+#include "train_common.h"
+
+
+
+#ifdef TRAIN_CUDA
+
+// CUDA error checking
+void cudaDoCheck(cudaError_t error, const char *file, int line) {
+  if (error != cudaSuccess) {
+    printf("[CUDA ERROR] at file %s:%d:\n%s\n", file, line,
+           cudaGetErrorString(error));
+    exit(EXIT_FAILURE);
+  }
+};
+
+// cuBLAS error checking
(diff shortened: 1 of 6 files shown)


train_common.c
```diff
@@ -0,0 +1,279 @@
...
+}
+
+
+// ----------------------------------------------------------------------------
+// Occasional gen_forward step
+
+void gen_forward(GPT2 *modelp, Tokenizer *tokenizerp, int genT, int B, int T  ) {
+
+    int* gen_tokens = (int*)mallocCheck(B * T * sizeof(int));
+#ifdef TRAIN_CUDA
```
karpathy, line 227: oh oh here we go. Instead of this maybe the `gen_forward` should be kept in the originals.

karpathy:
So I want to do this, I'm just not sure when :)

karpathy:
I certainly don't want to go down the IFDEF switching hell.

rosslwheeler:
Yes, it would be great if we can reduce / eliminate the use of #ifdef's in the source.

train_common.c
zocterminal: Putting the gen forward back would be easy (just delete the function and leave the original source in place).  And in that case the cudaChecks can go back as well.  I'm not at a build environment at the moment, but I can do that tomorrow.  Just tell me what you prefer to have changed. 

E.g. we could also eliminate the .h file and the changes in Makefile changes by just including the .c file in the other files (as you did in the test_gpt.cu, I just didn't know what style you prefer, so I did it my way (I'm used to compressing code and heavy use of ifdefs).

zocterminal:
What's hell for some is just a cozy summer vacation to others ;-)

dagelf:
Question is, really, what is the future of this project - how far can it go and how complex can it really get. Maybe split it out when it gets to the 4th target, 3 targets/flavors in one file doesn't mess with the elegant simplicity too much, does it? And exceptions for compilers wouldn't have any reason to blow up, would it? 

It is C after all...

dagelf:
Having most of a file covered in ifdefs sure messes with the elegant simplicity. Compiler quirks should be few, so its the targets that are the real concern, isn't it? But it is C after all, maybe the pragmatic approach would be to refactor/split things out to separate files if and when a 4th platform is added? 2 seems managable, 3 might even be. Four could become too much to keep track of. I am here to learn though.

zocterminal:
Closing (has hard to fix merge conflicts now, new version with different approach is https://github.com/karpathy/llm.c/pull/177).

PR #175, opened 2024-04-19
Fix the 0.0f initialization for double

lancerts:


zocterminal:
I'm wondering what the background for this change is.

The compiler certainly knows that the left side is a double and the conversion is made at compile time.

Is there any practical reason for this change or merely optics (which may be a good enough reason in itself ... just asking).

PR #177, opened 2024-04-19
Reorgsv2 😜

zocterminal:
2nd attempt to reorganize source code to consolidate copied functions:

Compared to the first "reorgs" pull request (https://github.com/karpathy/llm.c/pull/174) this is a minimalistic approach.
Only the most obvious candidates were put into a c-file and then pulled in from the main sources via `#include`.

- common stuff is now in train_common.c and train_common.h 
- train_common.h is not strictly necessary, but will make it easier to move away from the `#include` approach and compile separately (see the old "reorgs" (https://github.com/karpathy/llm.c/pull/174) how that would be done)
- made use the checkFopen(), mallocCheck() etc. functions in `train_gpt2.c` also.

It would be possible to move the GPT2 structs into `train_common.h` also.  The differences there between the files are marginal and would require only miniscule use of `#ifdefs`, but I didn't do that (based on the comments in the prior "reorgs" pull request).  The structs `GPT2Config` and `ParameterTensors` tensors are even identical between .c and .cu.

Enjoy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,18 +45,18 @@ endif
 # default target is all
 all: train_gpt2 test_gpt2 train_gpt2cu test_gpt2cu
 
-train_gpt2: train_gpt2.c
-	$(CC) $(CFLAGS) $(INCLUDES) $(LDFLAGS) $< $(LDLIBS) -o $@
+train_gpt2: train_gpt2.c train_common.c
+	$(CC) $(CFLAGS) $(INCLUDES) $(LDFLAGS) $^ $(LDLIBS) -o $@
 
-test_gpt2: test_gpt2.c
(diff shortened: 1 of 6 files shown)


zocterminal:
I'm fine either way.  

(I'm playing with your source on a superficial level because that helps me understand it more and helps me ingest it by osmosis ... and it even improves my own software (long story), so I'm just sharing the fruits of my work for what had my payoff already).

zocterminal:
Well, since you're speaking of the pain this is causing already. 

I'm very illiterate with AI, but I have decades of "C" experience under my belt and I can says this: Copying code is like a small cut. It first hurts just a little, then it hurts more and then it gets infested.  And the longer you ignore it, the worse the pain will be and the harder and more painful the cure will be.

PR #179, opened 2024-04-19
towards an even better backward attention kernel

ngc92:
Back to the drawing board, because I think the other kernels hit a local minimum, or at least the way the loops where organized made it very difficult to think about how to optimize this further.

I think there is quite a bit of room to optimize this version further, but for educational purposes, I think it is better to have a simpler version first, where the main idea is evident, and then go crazy with the optimizations in a second version.

This design limits T to multiples of the block_size. Incidentally, on my system at least, block size 64 turns out to be fastest, which actually is compatible with both our training at test scripts.

### DIFF
--- a/dev/cuda/attention_backward.cu
+++ b/dev/cuda/attention_backward.cu
@@ -623,14 +623,64 @@ __global__ void __launch_bounds__(BlockSize) softmax_autoregressive_backward_ker
                 }
             } else {
                 for (int t2i = 0; t2i < end; t2i++) {
-                    int t2 = t2b + t2i;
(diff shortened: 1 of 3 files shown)


dev/cuda/attention_backward.cu
```diff
@@ -560,6 +561,128 @@ __global__ void softmax_autoregressive_backward_kernel5(float* __restrict__ dpre
     }
 }
 
+
+// I want `BlockSize` to be statically known to the compiler, thus we get a template here.
```
karpathy, line 565: love this comment block

PR #184, opened 2024-04-19
Some more optimizations for the backward attention kernel

ngc92:
We're hitting diminishing returns, so this is probably going to be last update for this kernel from me for a while. I really like the simplicity of kernel 7, but  even kernel 8 isn't so bad, and we do get 5-10% speedup from this on my system.


karpathy:
I'm seeing a solid improvement on my end too, down to 77.9 ms/iter average.
Good call on cublasCheck too
Merging!

PR #186, opened 2024-04-19
Removed a few unnecessary heap allocations

jonathanmarvens:
This is just a minor update to removed a few unnecessary heap allocations.

PR #188, opened 2024-04-19
Updated a few variables to use exact width integer types

jonathanmarvens:
This is just a minor update to update a few variables to use exact width integer types.


jonathanmarvens:
@karpathy Given your comment in PR #186 (https://github.com/karpathy/llm.c/pull/186#issuecomment-2067294452), you mind telling me what issues you see with this types update? I think using the exact width types is the correct move here, but maybe I’m missing something.

PR #193, opened 2024-04-20
Minimal DDP implementation using MPI

chinthysl:
Simplest implementation of DDP ring all reduce algo. [https://pytorch.org/tutorials/intermediate/dist_tuto.html#our-own-ring-allreduce](url)
Have used MPI to exchange calculated gradients between processes just after the gpt2_backward function call.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,14 @@ LDFLAGS =
 LDLIBS = -lm
 INCLUDES =
 
+MPI_LDFLAGS = -L/usr/lib/x86_64-linux-gnu/openmpi/lib
+MPI_LDLIBS = -lmpi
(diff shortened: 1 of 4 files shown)


karpathy:
Not mergeable because we have to make it as optional / local and small and constrained as possible, but really cool PR, very eager to play with! 

E.g. we cannot modify the way we write the dataset, we'll have to modify the DataLoader in C if num_proc > 1, etc.

chinthysl:
@karpathy thank you for reviewing it. I'll try to make it constrained and small as possible.

chinthysl:
@karpathy please review the latest changes, I have made MPI optional and refactored somewhat.

karpathy:
We merged an MPI+NCCL version here:

https://github.com/karpathy/llm.c/commit/0c3e3e37f4109cdcc9b705c841f03d20a29e1472

So I am closing this PR. Thank you!

PR #195, opened 2024-04-20
profile and print out some nice stats

ngc92:
I think this should be more or less the script you wanted :)

The read/write and instruction numbers are mainly included so we can print out these cool stats at the end. If you wanted to identify bottlenecks, the bandwidth% would be much more helpful, but IMO, for that, you really should just open the file in nsight. Here, I just wanted to visualize how much work of which kind gets done in the different parts of the network.


karpathy:
nice :)

PR #196, opened 2024-04-20
Minimal DDP implementation using CUDA aware MPI 

chinthysl:
Simplest implementation of DDP ring all reduce algo in CUDA aware MPI. Simplified version of [ring-allreduce](https://pytorch.org/tutorials/intermediate/dist_tuto.html#our-own-ring-allreduce).
No complexity or additional dependencies added like NCCL. Pure MPI for easy understanding.

Related to the previous [PR](https://github.com/karpathy/llm.c/pull/193). Please use it for MPI make file changes required for nvcc.
I used MPI to exchange calculated gradients between processes just after the gpt2_backward function call.
When the MPI is installed with CUDA enabled, device buffers can be directly transferred between process using normal MPI send receive calls.
Implemented a naive kernel to sum the gradient buffer received in each iteration.


karpathy:
@chinthysl ty for the PRs btw, very eager to give it a go shortly. Spending a lot of cycles on the faster fp32 kernels still, and the fp16/bf16, but aiming to get around to it shortly.

karpathy:
We merged an MPI+NCCL version here:

https://github.com/karpathy/llm.c/commit/0c3e3e37f4109cdcc9b705c841f03d20a29e1472

So I am closing this PR. Thank you!

RE: NCCL or not, in this version of code we prefer speed, so we included the NCCL changes. Nice to see that the mpi version is quite small though!

PR #198, opened 2024-04-20
Support for vanilla build without CUDA tools

Ricardicus:
I tried to build on my mac and ran into a build error

```
$ make
NICE Compiling with OpenMP support
cc -Ofast -fno-finite-math-only -Wno-unused-result -march=native -Xclang -fopenmp -DOMP  -I/opt/homebrew/opt/libomp/include  -L/opt/homebrew/opt/libomp/lib train_gpt2.c -lm -lomp -o train_gpt2
clang: error: the clang compiler does not support '-march=native'
make: *** [train_gpt2] Error 1
```

It seems like the flag -march=native was not supported on my clang compiler.

I introduced the CFLAGS_COND as a list of CFlags that are included only if the CC supports it.

I also only build the nvcc targets if nvcc is installed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,8 @@
 data/
 gpt2_124M.bin
 gpt2_124M_debug_state.bin
+
+test_gpt2
+test_gpt2cu
+train_gpt2
+train_gpt2cu
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,27 @@
 CC ?= clang
-CFLAGS = -Ofast -fno-finite-math-only -Wno-unused-result -march=native
+CFLAGS = -Ofast -fno-finite-math-only -Wno-unused-result
 LDFLAGS =
 LDLIBS = -lm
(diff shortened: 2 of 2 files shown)


Ricardicus:
Should we perhaps add a .github workflow that check build status on some supported platforms and conditions?

karpathy:
@Ricardicus it's a good idea. I don't have that much experience with setting that up but would be happy to learn and accept PRs to that effect.

karpathy:
merged, ty for the PR

Ricardicus:
> @Ricardicus it's a good idea. I don't have that much experience with setting that up but would be happy to learn and accept PRs to that effect.

The workflow can run the test also. I can give it a go 👍

Ricardicus:
I have a PR now for this: #209 
With this in place, no commits to master or PR that breaks the builds or CPU training/testing will go unnoticed.

PR #199, opened 2024-04-20
Further memory reductions

ngc92:
With the fused classifier layer, there is actually no need to track logits and dlogits separately, and we can reuse that buffer.
Requires slight changes to the testing script, so we don't test dlogits.
Further, the helper buffers in the attention code aren't actually used during backward, so we really don't need on copy per layer.


karpathy:
Great finds! I'm seeing memory consumption go 14372MiB -> 10774MiB as a result, saving 3598MiB, **25% of the original**. Huge!

PR #201, opened 2024-04-20
Yet more memory reductions

ngc92:
Currently fails the test, because right now there is no way to get logits out of the model
Run as training => get  dlogits
Run for inference => get probs

The memory savings are quite substantial, though:
```
allocated 7127 MiB for activations
allocated 1421 MiB for activation gradients
=>
allocated 5946 MiB for activations
allocated 636 MiB for activation gradients
```

### DIFF
--- a/test_gpt2.cu
+++ b/test_gpt2.cu
@@ -95,7 +95,7 @@ int main(int argc, char *argv[]) {
     // at this point, target should be equal to expected_logits, let's compare
     // copy logits to CPU so we can compare them
     float* logits_cpu = (float*)mallocCheck(B * T * V * sizeof(float));
-    cudaMemcpy(logits_cpu, model.acts.logits, B * T * V * sizeof(float), cudaMemcpyDeviceToHost);
+    cudaMemcpy(logits_cpu, model.acts.output.logits, B * T * V * sizeof(float), cudaMemcpyDeviceToHost);
     int logits_ok = 1;
     for (int i=0; i<B*T*V; i++) {
         if(i < 3) {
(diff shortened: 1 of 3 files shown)


ngc92:
@karpathy I thought things would become more readable if I split the forward/backward activations into separate classes.
In a way, this worked, and it became much easier to track which memory was needed where, and  I could remove almost the entire thing!

At this point, I'm wondering whether we even should have a separate backward allocation, or whether we should just allocate the remaining three (!!) tensors with the forward allocations.

Here are the new numbers:
```
allocated 474 MiB for model parameters
allocated 5706 MiB for activations
allocated 474 MiB for parameter gradients
allocated 252 MiB for activation gradients
allocated 474 MiB for AdamW optimizer state m
allocated 474 MiB for AdamW optimizer state v
```
Total 7380 => I think we should be able to train on an 8GB GPU now  :)


Is there a reason why we're allocating the `m` and ` v` separately, instead  of one allocation + pointers as for the rest?

FeSens:
This is really impressive, but it's not clear to me from where those memory footprint reductions came from.
Was it all in the `fill_in_activation_sizes` change from `L * ...` to `B * ...`?

ngc92:
its an accumulation of many thing. For example, the logits tensor is several 100 MB, so by reusing this for the logits, logit gradients, and probabilities (which are never needed simultaneously) you get noticeable improvements.

Then, it is generally tracking when you read/write. For example, if you look at  the changes in this commit, it is based on the realization that the usage pattern in the backward pass generally is:
You have some buffer, and fill it with the gradient of the current operation. Then you backprop to the next, and read that buffer. Then you go to the next op, and this buffer sits unused. So if you later on have need for another buffer of the same (or smaller) size, you don't need separate memory, you can just overwrite the  values you were never going to read again anyway.

karpathy:
merging, this is really cool. 
i knew conceptually that backward only needs memory for the "frontier" of gradients (unlike the forward pass where caching is necessary), but the reductions are a lot more dramatic once you actually get in there and start slashing things.

agree that we want to look into eliminating the grads altogether as a separate structure in future commits.

PR #208, opened 2024-04-21
train_gpt2.py: fix for conditional use of fused and for computing gradients

Ricardicus:
The train_gpt2.py is broken at the moment.
This is a fix for that. 
These things I found to be wrong and have fixed:

- gradients are not computed which results in 

```
  File "/Users/runner/work/llm.c/llm.c/train_gpt2.py", line 403, in <module>
    write_state(model, x, y, logits, loss, "gpt2_124M_debug_state.bin")
  File "/Users/runner/work/llm.c/llm.c/train_gpt2.py", line 27[9](https://github.com/Ricardicus/llm.c/actions/runs/8775693842/job/24078315777#step:6:10), in write_state
    grads = {name: param.grad.cpu() for name, param in model.named_parameters()}
                   ^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'cpu'
```

- fused is used when the device is not cuda which breaks down on cpu and mps 

- attempting to write data that is not on the cpu to disk

- torch.cuda.reset_peak_memory_stats() only works if cuda is available

### DIFF
--- a/train_gpt2.py
+++ b/train_gpt2.py
@@ -214,7 +214,7 @@ class GPT(nn.Module):
 
 # a few utilities for saving params/grads/activations to files for loading in C
 def write_fp32(tensor, file):
-    file.write(tensor.detach().numpy().astype("float32").tobytes())
+    file.write(tensor.detach().cpu().numpy().astype("float32").tobytes())
 
 def write_tensors(model_tensors, L, file):
     write_fp32(model_tensors["transformer.wte.weight"], file) # (V, C)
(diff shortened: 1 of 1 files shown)


Ricardicus:
I stumbled upon this when I worked on the workflow, PR #209 
That PR includes this fix. I thought it would look good with a premier workflow run that is successful, so I wanted that fix in there.. If you want that PR, then this PR can be dismissed.

PR #209, opened 2024-04-21
Adding a github workflow

Ricardicus:
This is a start, but this workflow does the following:

- builds and tests on CPU with and without openMP (macos and ubuntu)
- builds the CUDA files (does not test them)

Ideally one would like to test the CUDA also, but it is hard to do this.
I think it would be possible to use the virtual GPU drivers from NVIDIA, but I don't know
how to do this and I think that rabit hole can be left uninvestigated for now. This is a start!

It will run on PRs and pushes to the master branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,62 @@
+name: Build and test
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build-and-test-cpu:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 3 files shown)


Ricardicus:
This PR includes the fix presented in this PR:

https://github.com/karpathy/llm.c/pull/208

So If you want to merge this, then you can forget about PR #208

Ricardicus:
Here is a link to the logs of how this workflow was running on my fork:
https://github.com/Ricardicus/llm.c/actions/runs/8776172376

karpathy:
very cool, let's test it out!

Ricardicus:
Yey!

PR #210, opened 2024-04-21
Added shared memory for the atomic additions for the layernorm_back

ChrisDryden:
This cr was made to address the issue found in the profiler that the atomic operations in the final loop of this kernel were causing a bunch of warp stalls. By doing the atomic operation on shared memory instead of global memory and then copying the shared memory afterwards it signifigantly reduces the time by almost a factor of 4 for the largest block sizes:

Before:
```
block_size   32 time 0.3854 ms
block_size   64 time 0.4663 ms
block_size  128 time 0.4643 ms
block_size  256 time 0.4715 ms
block_size  512 time 0.4740 ms
block_size 1024 time 0.4769 ms
```

After:
```
block_size   32 time 0.3658 ms
block_size   64 time 0.2301 ms
block_size  128 time 0.1782 ms
block_size  256 time 0.1526 ms
block_size  512 time 0.1426 ms
block_size 1024 time 0.1431 ms
```

### DIFF
--- a/dev/cuda/layernorm_backward.cu
+++ b/dev/cuda/layernorm_backward.cu
@@ -175,6 +175,20 @@ __global__ void layernorm_backward_kernel2(float* dinp, float* dweight, float* d
     const float mean_bt = mean[b * T + t];
     const float rstd_bt = rstd[b * T + t];
 
+    extern __shared__ float shared[];
+
+    float* dbias_shared = shared;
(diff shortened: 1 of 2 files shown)


karpathy:
This will be very nice; Can you please also fix the formatting (it looks buggy almost with the indents) and make the style a bit more consistent?

PR #213, opened 2024-04-22
Custom matmul attention

ngc92:
My own implementation of (lower-triangular) matrix multiplication. 
It is not as efficient as CuBLAS, but since we only calculate half as many numbers, it is a net win.

Cannot get rid of `permute` yet, because  `v` is still needed later in the kernel. Will look into replacing the second matmul too, so we can capitalize on not needing permutes (might require changes to backward, too).


ngc92:
this was never intended to be actually merged, just to demonstrate how the custom matmul would look like. Now that it is full of conflicts, let's keep the number of open PRs to a reasonable level.

PR #214, opened 2024-04-22
added triangular matrix multiplication kernel

ngc92:
Companion to #213, adding a file specifically for the development of this matmul.
Also shows different intermediate kernels on the way towards efficiency. 

To give a break from all the maths and indexing in the code, the development of these is described as a story. 
Some of the metaphors are stretched quite a bit, so feel free to make adjustments, but I  hope that overall, this might be easier to follow than just "indexing with this formula to achieve coalesced access".

Currently, the reads in the inner loop still cause 2-way bank conflicts, so there is still room for improvement.

Timings on my machine:
>  time 1.40 ms vs 2.37 ms for CuBLAS

Given that we're doing only half the work, that leaves us still 20% less efficient than cuBLAS.

### DIFF
--- a/dev/cuda/common.h
+++ b/dev/cuda/common.h
@@ -86,8 +86,8 @@ void validate_result(T* device_result, const T* cpu_reference, const char* name,
         if (i < 5) {
             printf("%f %f\n", cpu_reference[i], out_gpu[i]);
         }
-        // ensure correctness for all elements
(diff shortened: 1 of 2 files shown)


karpathy:
Wow, you really had a lot of fun with the TriMatlon 😂 😂 😂 
The most incredible fusion of art and engineering I've seen yet :D

PR #200, opened 2024-04-20
Fixes -Ofast optimizations breaking model by skipping them for gelu_backward

dagelf:
Please see https://github.com/karpathy/llm.c/pull/168 for details and work being done on optimizing the activation function on CPU.

This PR results in a 10-23% speedup on CPU and has been tested on `msvc` and `gcc` on x86 and fixes #19

PR #216, opened 2024-04-22
Make msvc compile by adding back the ifdefs in gelu_backward

dagelf:
Fixes #200 breaking `msvc`.

### DIFF
--- a/train_gpt2.c
+++ b/train_gpt2.c
@@ -364,8 +364,10 @@ void gelu_forward(float* out, float* inp, int N) {
 }
 
 // we want to use -Ofast optimization, but sadly GeLU breaks, so disable this flag just for it (#168)
-#pragma float_control(precise, on, push) // On msvc /fp:fast is a lot faster, but the expf inside coshf breaks the model
-__attribute__((optimize("no-finite-math-only"))) // same for gcc -Ofast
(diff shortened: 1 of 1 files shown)


dagelf:
Also resolves #217

dagelf:
Chances are it was the GPT2 download, maybe I didn't stare at the log hard enough... but:

https://status.huggingface.co/
image

dagelf:
Of course you're right. This is why #215 failed:

image

dagelf:
Still, so far it's one less download costing just 1M ... and it fixes `msvc` compile.

Should I move all the output `.bin` files to `data/` too? Currently they clutter the root.

.gitignore
```diff
@@ -1,8 +1,10 @@
 # dot files and such
 .vscode/
+.venv/
```
azret, line 4: add
 
.vs/

as well please for visual studio

karpathy:
I don't want to clutter the repo with shakespeare if we can avoid it. It's a pet peeve to git clone a repo and have to wait long seconds/minutes for the download.

karpathy:
Let's not move the bin files right now, let's try to keep this PR focused and minimal

dagelf:
Impressive that this repo is still under 2M, including all history. New PR incoming.

Issue #144, opened 2024-04-16
[todo] Accumulate in double instead of float


karpathy:
I have not kept good hygiene on using `double` for accumulates everywhere that you have local register variables. Accumulate should be in `double`, and then read/write in `float`, todo to fix everywhere, part by part. Doing this most likely results in much better precision/accuracy at near zero cost of additional compute/memory.

Example accumulate that should have been a double instead of float:

```c
 float m = 0.0f;
for (int i = 0; i < C; i++) {
    m += x[i];
}
m = m/C;
```

ross-wheeler:
Are you talking about these warnings?
train_gpt2.c(232,17): warning C4244: 'initializing': conversion from 'double' to 'float', possible loss of data
train_gpt2.c(304,31): warning C4244: 'argument': conversion from 'int' to 'float', possible loss of data
train_gpt2.c(304,17): warning C4244: 'initializing': conversion from 'double' to 'float', possible loss of data
train_gpt2.c(362,43): warning C4305: 'argument': truncation from 'double' to 'float'
train_gpt2.c(370,26): warning C4305: 'argument': truncation from 'double' to 'float'
train_gpt2.c(374,77): warning C4305: 'argument': truncation from 'double' to 'float'
train_gpt2.c(934,47): warning C4244: 'argument': conversion from 'int' to 'float', possible loss of data
train_gpt2.c(935,47): warning C4244: 'argument': conversion from 'int' to 'float', possible loss of data
train_gpt2.cu(728): warning C4244: 'argument': conversion from 'int' to 'float', possible loss of data
train_gpt2.cu(728): warning C4244: 'initializing': conversion from 'double' to 'float', possible loss of data

syoyo:
Kahan summation is well used technique to reduce numerical error in summation computation

https://en.m.wikipedia.org/wiki/Kahan_summation_algorithm

Its rather simple to implement.

ChrisDryden:
Putting in here a summary of the discussion of the result of this CR: https://github.com/karpathy/llm.c/pull/221 There is a signifigant slowdown on the order of magnitude of 32x for using doubles compared to floats on gaming gpu's and by using doubles instead of floats for the summation in this location led to a 2x slowdown even though the variance in error was reduced. Will have to be done on a case by case basis looking at the tradeoff of CPU vs Memory bound kernels.

PR #221, opened 2024-04-22
Faster `matmul_backward_bias` using coalesced reads and shared memory in the kernel

al0vya:
This kernel seems to offer a <4x runtime improvement over `matmul_backward_bias_kernel2` on an RTX 2070 Super GPU, runtime comparison shown below:

```
matmul_backward_bias_kernel2:
block_size 32 time 0.9027 ms
block_size 64 time 4.0396 ms
block_size 128 time 3.4077 ms
block_size 256 time 1.4159 ms
block_size 512 time 0.7302 ms
block_size 1024 time 0.9947 ms

matmul_backward_bias_kernel4:
block_size 32 time 0.6875 ms
block_size 64 time 0.4235 ms
block_size 128 time 0.2925 ms
block_size 256 time 0.2419 ms
block_size 512 time 0.2714 ms
block_size 1024 time 0.2519 ms
```

The kernel passes `test_gpt2cu` without any issue, and also the tests in `dev/cuda/matmul_backward_bias.cu` *after increasing the tolerance from 1e-3 to 5e-3*: with the tolerence set to 1e-3, the weights printed to the console are very similar to the CPU reference weights, so I am hazarding that this is an acceptable change -- I've seen a tolerance of up to 1e-2 in other parts of the code...

In `test_gpt2cu`, the runtime seems to reduce from ~42 ms/iter to ~39 ms/iter (again, using an RTX 2070 Super GPU).

I've included some comments at the start of the kernel to explain the philosophy behind it.

Happy to receive feedback to improve the PR, sorry if I've missed something obvious.

### DIFF
--- a/dev/cuda/matmul_backward_bias.cu
+++ b/dev/cuda/matmul_backward_bias.cu
@@ -194,7 +194,7 @@ void matmul_backward_bias3(float* dinp, float* dweight, float* dbias,
 void matmul_backward_bias4(float* dinp, float* dweight, float* dbias,
                       float* dout, float* inp, float* weight, float* ones,
                       int B, int T, int C, int OC, int block_size) {
-    const int grid_size = OC / 32;
+    const int grid_size = OC / 32; // for now, OC must be divisible by 32 for this kernel to work
     matmul_backward_bias_kernel4<<<grid_size, block_size, block_size * sizeof(float)>>>(dbias, dout, B, T, OC);
 }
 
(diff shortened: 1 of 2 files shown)


ChrisDryden:
Was thinking that the weight tolerance could be fixed by setting the summation variable to a double https://github.com/karpathy/llm.c/issues/144

karpathy:
@ChrisDryden good idea i'm working on the kernel right now, will check

PR #226, opened 2024-04-22
gitignore .* and *.bin

dagelf:
Is this acceptable to keep it simple and make `msvc` peeps happy?


rosslwheeler:
I still show 71 files unstaged with the above but I'm testing the solution/project files currently which generates a lot more than the Windows Makefile version does. Let me reply back with that Makefile amount - I'm sure it's smaller. We will want to ignore .obj's and .exe's at a minimum.

dagelf:
Isn't there any option we can set to make it generate all it's artifacts in a .tmp directory or such?

PR #231, opened 2024-04-23
Added ref to the zig port 

Saimirbaci:
Added a reference implementation in zig of the llm.c


karpathy:
I'm so glad I didn't hit merge quickly as I usually do for these

Saimirbaci:
Oops good catch, sorry about that. Fixing it now

Saimirbaci:
Now it is fixed, sorry again.

PR #233, opened 2024-04-23
feat(attention_forward.cu): Gentle introduction to flash attention 2

FeSens:
This is a very, very gentle introduction to Flash Attention 2 with CuTe (Cutlass v3).
It's gentle because it's not fully working, and it's not finished.

What I've got so far:
* Work partitioned between Query block, Batch, and Head (as is Flash attention 2 to the best of my knowledge);
* Efficiently copying tiles of Q and K using CuTe;

Where I'm stuck:
* Using  cute::gemm is not working, I'm not able to make it compile.

I will be hanging out on the CudaMode discord if anyone wants to pair or has a better understanding of CuTe and wants to help =).


FeSens:
Ok, I've figured out that I need to make the layouts of sQ and sK static for now if I want to use gemm(sQ, sK, ...).

ericauld:
Is it compiling now?

FeSens:
It is! Not to the performance I wish it had, but it definitely is compiling.
Thanks for the Cutlass class @ericauld !
Any tips on how to speed up this kernel?

karpathy:
@FeSens can you post what kind of perf you're seeing for this?

FeSens:
This is at 65% of the speed of Cublas `cublasSgemmStridedBatched`

ngc92:
I believe you're running into the same trap I did. We currently don't enable tensor cores in this dev file, so `cublasSgemmStridedBatched` will be much slower than it is going to be in the real model.

PR #236, opened 2024-04-24
Windows changes for Makefile

rosslwheeler:
Trimmed down the Makefile changes earlier for Windows. Tested with Windows and with Ubuntu 22 (WSL) - no issues. I noticed the Cuda performance has improved by 2x which was cool. Once this is in, then I'll use this to update the CI and we'll have coverage for Windows. Let me know if you see anything problematic. Was thinking we could reorganize this file into 3 discrete sections: OS check, Compiler check and Build. Just a thought...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,24 +4,42 @@ LDFLAGS =
 LDLIBS = -lm
 INCLUDES =
 CFLAGS_COND = -march=native
-
-# Find nvcc
-NVCC := $(shell which nvcc 2>/dev/null)
+SHELL_UNAME = $(shell uname)
+REMOVE_FILES = rm -f
+OUTPUT_FILE = -o $@
 
 # NVCC flags
 NVCC_FLAGS = -O3 --use_fast_math
 NVCC_LDFLAGS = -lcublas -lcublasLt
 
-# Function to test if the compiler accepts a given flag.
-define check_and_add_flag
+ifneq ($(OS), Windows_NT)
+  NVCC := $(shell which nvcc 2>/dev/null)
+
+  # Function to test if the compiler accepts a given flag.
+  define check_and_add_flag
     $(eval FLAG_SUPPORTED := $(shell printf "int main() { return 0; }\n" | $(CC) $(1) -x c - -o /dev/null 2>/dev/null && echo 'yes'))
     ifeq ($(FLAG_SUPPORTED),yes)
         CFLAGS += $(1)
     endif
-endef
+  endef
 
-# Check each flag and add it if supported
-$(foreach flag,$(CFLAGS_COND),$(eval $(call check_and_add_flag,$(flag))))
+  # Check each flag and add it if supported
+  $(foreach flag,$(CFLAGS_COND),$(eval $(call check_and_add_flag,$(flag))))
+else
+  CFLAGS :=
+  REMOVE_FILES = del
+  SHELL_UNAME := Windows
+  NVCC := $(shell where nvcc 2> nul)
+  CC := cl
+  CFLAGS = /Idev /Zi /nologo /Wall /WX- /diagnostics:column /sdl /O2 /Oi /Ot /GL /D _DEBUG /D _CONSOLE /D _UNICODE /D UNICODE /Gm- /EHsc /MD /GS /Gy /fp:fast /Zc:wchar_t /Zc:forScope /Zc:inline /permissive- \
+   /external:W3 /Gd /TP /wd4996 /FC /openmp:llvm
+  LDFLAGS :=
+  LDLIBS :=
+  INCLUDES :=
+  NVCC_FLAGS += -I"dev"
+  WIN_CUDA_RENAME = rename $@.exe $@
+	OUTPUT_FILE = /link /OUT:$@
+endif
 
 # Check if OpenMP is available
 # This is done by attempting to compile an empty file with OpenMP flags
@@ -34,7 +52,7 @@ ifeq ($(NO_OMP), 1)
   $(info OpenMP is manually disabled)
 else
   # Detect if running on macOS or Linux
-  ifeq ($(shell uname), Darwin)
+  ifeq ($(SHELL_UNAME), Darwin)
     # Check for Homebrew's libomp installation in different common directories
     ifeq ($(shell [ -d /opt/homebrew/opt/libomp/lib ] && echo "exists"), exists)
       # macOS with Homebrew on ARM (Apple Silicon)
@@ -54,19 +72,21 @@ else
       $(warning OpenMP not found, skipping OpenMP support)
     endif
   else
-    # Check for OpenMP support in GCC or Clang on Linux
-    ifeq ($(shell echo | $(CC) -fopenmp -x c -E - > /dev/null 2>&1; echo $$?), 0)
-      CFLAGS += -fopenmp -DOMP
-      LDLIBS += -lgomp
-      $(info OpenMP found, compiling with OpenMP support)
-    else
-      $(warning OpenMP not found, skipping OpenMP support)
+    ifneq ($(OS), Windows_NT)
+      # Check for OpenMP support in GCC or Clang on Linux
+      ifeq ($(shell echo | $(CC) -fopenmp -x c -E - > /dev/null 2>&1; echo $$?), 0)
+        CFLAGS += -fopenmp -DOMP
+        LDLIBS += -lgomp
+        $(info OpenMP found, compiling with OpenMP support)
+      else
+        $(warning OpenMP not found, skipping OpenMP support)
+      endif
     endif
   endif
 endif
 
-# PHONY means these targets will always be executed
-.PHONY: all train_gpt2 test_gpt2 train_gpt2cu test_gpt2cu train_gpt2fp32cu test_gpt2fp32cu
+# DEFAULT means these targets will always be executed
+.DEFAULT: all 
 
 # Add targets
 TARGETS = train_gpt2 test_gpt2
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -54,19 +72,21 @@ else
...
+        $(info OpenMP found, compiling with OpenMP support)
+      else
+        $(warning OpenMP not found, skipping OpenMP support)
+      endif
     endif
   endif
 endif
 
-# PHONY means these targets will always be executed
-.PHONY: all train_gpt2 test_gpt2 train_gpt2cu test_gpt2cu train_gpt2fp32cu test_gpt2fp32cu
```
Ricardicus, line 69: .PHONY is very convenient, it rebuilds the targets even if they exist. Why is it removed?

Makefile
```diff
@@ -4,24 +4,42 @@ LDFLAGS =
...
+  NVCC := $(shell where nvcc 2> nul)
+  CC := cl
+  CFLAGS = /Idev /Zi /nologo /Wall /WX- /diagnostics:column /sdl /O2 /Oi /Ot /GL /D _DEBUG /D _CONSOLE /D _UNICODE /D UNICODE /Gm- /EHsc /MD /GS /Gy /fp:fast /Zc:wchar_t /Zc:forScope /Zc:inline /permissive- \
+   /external:W3 /Gd /TP /wd4996 /FC /openmp:llvm
+  LDFLAGS :=
+  LDLIBS :=
+  INCLUDES :=
+  NVCC_FLAGS += -I"dev"
+  WIN_CUDA_RENAME = rename $@.exe $@
+	OUTPUT_FILE = /link /OUT:$@
```
Ricardicus, line 41: spaces/tabs here

Makefile
```diff
@@ -76,31 +96,36 @@ ifeq ($(NVCC),)
...
     $(info nvcc found, including CUDA builds)
-    TARGETS += train_gpt2cu test_gpt2cu train_gpt2fp32cu test_gpt2fp32cu
+    TARGETS += train_gpt2cu test_gpt2cu train_gpt2fp32cu test_gpt2fp32cu profile_gpt2cu
 endif
 
 all: $(TARGETS)
 
 train_gpt2: train_gpt2.c
-	$(CC) $(CFLAGS) $(INCLUDES) $(LDFLAGS) $< $(LDLIBS) -o $@
+	$(CC) $(CFLAGS) $(INCLUDES) $(LDFLAGS) $< $(LDLIBS) $(OUTPUT_FILE)
```
Ricardicus, line 105: My personal opinion is that I think the OUTPUT_FILE variable obfuscates a little more than it helps. It is also not used for the *cu recipies.

Makefile
Ricardicus, line 69: .PHONY is very convenient. It rebuilds the files even if they exists. Why is it removed? I would like it back..

rosslwheeler: Ah - was trying to avoid that! My preference is know if a target has been built and to not rebuild it. I've normally seen .PHONY only have special commands and not targets. What's the rationale for wanting to build something that's already successfully been built? And no problem putting it back. Thought it was a mistake...no worries...

Makefile
rosslwheeler: Got it - will fix that. Having trouble with Makefiles in VS Code with it using the wrong format.

Ricardicus:
Since windows is such a world of its own, there will be a question about maintaining it. In order to make it easier during development, should we perhaps introduce a flag to the preprocessor that is only active when building on windows? I can image people writing whole sections of code that is POSIX compliant with zero care in the world for the Windows case. The CI will break and then the question is "can you rewrite" etc.. or "should we stop supporting Windows"? With a flag it is possible to get around it, but this can cause the code to look strange over time.

Makefile
```diff
@@ -4,24 +4,42 @@ LDFLAGS =
...
-$(foreach flag,$(CFLAGS_COND),$(eval $(call check_and_add_flag,$(flag))))
+  # Check each flag and add it if supported
+  $(foreach flag,$(CFLAGS_COND),$(eval $(call check_and_add_flag,$(flag))))
+else
+  CFLAGS :=
+  REMOVE_FILES = del
+  SHELL_UNAME := Windows
+  NVCC := $(shell where nvcc 2> nul)
+  CC := cl
+  CFLAGS = /Idev /Zi /nologo /Wall /WX- /diagnostics:column /sdl /O2 /Oi /Ot /GL /D _DEBUG /D _CONSOLE /D _UNICODE /D UNICODE /Gm- /EHsc /MD /GS /Gy /fp:fast /Zc:wchar_t /Zc:forScope /Zc:inline /permissive- \
```
Ricardicus, line 34: Perhaps add a flag that is only active during windows build..
/D WIN_BUILD or something like that.

Makefile
rosslwheeler: Yes, that's because the Cuda linker is different (it's runs like a 2-pass tool) and could not locate the switch to control the name of the output file. I assume there is one - let me know if you know what the syntax is for that.

Makefile
rosslwheeler: There's plenty of predefined macros by the compiler to know that we're building on Windows. Use those?

Makefile
Ricardicus: The comment above PHONY is misleading, perhaps rewrite it. PHONY means that Make will not check if the file exists before deciding to build it, which makes it seem like they are always executed which is why the comment is partly correct. But since you introduced DEFAULT, there is a little bit of confusion.. By default, without the DEFAULT, the first recipe definition is used. In this case it is "all".

Makefile
```diff
@@ -54,19 +72,21 @@ else
...
+        $(warning OpenMP not found, skipping OpenMP support)
+      endif
     endif
   endif
 endif
 
-# PHONY means these targets will always be executed
-.PHONY: all train_gpt2 test_gpt2 train_gpt2cu test_gpt2cu train_gpt2fp32cu test_gpt2fp32cu
+# DEFAULT means these targets will always be executed
+.DEFAULT: all
```
Ricardicus, line 89: I have though about it and I think you should reinstate .PHONY but correct the comment above it so that it isn't misleading. In this case, our first recipe definition is already "all" which we intend it to be, so there is no need for DEFAULT. The reason as to why .PHONY is used is not this, it is so that Make does not care if the file exist or not. If it exists, it will be rebuilt anyway. The CI actually relies on this. Otherwise, one would have to throw in "make clean" in order to trigger new builds after for example one toggles OpenMP on/off.

Makefile
rosslwheeler: Yes, that confused me.

Makefile
rosslwheeler: Got it - I was just trying to avoid the rebuild if it's already built. But I can put it back - no worries.

Makefile
Ricardicus: Yes, but since they are so many, and perhaps differ on different compiler versions (who knows), and since we control the build, we can easily just declare one that we know will be used in this codebase.

Makefile
Ricardicus: Yes, nvcc is doing a lot of work under the hood. With --verbose one can see it. It relies on the compilers and linkers already installed... Good thing that NVCC is OK with the -o flag then.

Makefile
```diff
@@ -76,31 +96,36 @@ ifeq ($(NVCC),)
...
 	$(NVCC) $(NVCC_FLAGS) $< $(NVCC_LDFLAGS) -o $@
+	$(WIN_CUDA_RENAME)
 
 test_gpt2fp32cu: test_gpt2_fp32.cu
 	$(NVCC) $(NVCC_FLAGS) $< $(NVCC_LDFLAGS) -o $@
+	$(WIN_CUDA_RENAME)
 
 profile_gpt2cu: profile_gpt2.cu
 	$(NVCC) $(NVCC_FLAGS) -lineinfo $< $(NVCC_LDFLAGS) -o $@
+	$(WIN_CUDA_RENAME)
```
Ricardicus, line 128: Even if the specify  -o to NVCC, it does in classic Window fashion still names its output with the .exe-suffix.
Maybe we should do the same? And stkip the WIN_CUDA_RENAME step. It also looks strange when we are not on Windows to have that line there.

rosslwheeler: Yes, that's what I was trying to fix. The problem is worse if we use the .exe name though. Then all the target names need a suffix. So you end up having something like:

profile_gpt2cu.$(EXE): profile_gpt2.cu 

I thought the above was worse and went with the WIN_CUDA_RENAME macro....

Makefile
rosslwheeler: Probably _WIN32 or _WIN64. I've seen _MSC_VEW used a bit too.

rosslwheeler:
@Ricardicus - will update the file shortly. Thanks for all of the helpful comments!

Makefile
Ricardicus: I think it is better we set a standard in our own code for this in the codebase to just distinguish between Windows and non-Windows builds for clarity and control.

Makefile
rosslwheeler: Yes, that surprised me that the -o worked. I will try to find a better solution.

Makefile
rosslwheeler: I would opt for the predefined _WIN64 but don't have a strong preference.

Ricardicus: I think that only works if we compile for the 64-bit case right?
Maybe we don't need to distinguish between those (yet).
_WIN32 is defined for both 32-bit and 64-bit environments when compiling on Windows, so it is more universal.
So if we don't use a custom macro, then I would opt for the _WIN32. 
But I would still like to see our own custom macro, because of the reasons i mentioned.

rosslwheeler: Got it - _WIN32 works and another TBD name-wise.

Makefile
```diff
@@ -85,8 +85,8 @@ else
   endif
 endif
 
-# DEFAULT means these targets will always be executed
-.DEFAULT: all 
+# Force the rebuild of targets - CI relies on this.
```
Ricardicus, line 88: Perhaps skip the "CI relies on this", it does rely on it now but we can easily change it so it doesn't.
We do not have to set this practice in stone. Maybe just "# Force the rebuild of targets".

rosslwheeler: Sure - one sec.

Ricardicus: Also, what do you think about letting Windows have its .exe-suffixes?
If the codebase grows, it would be time consuming to always force rebuilds, but here we seem to be OK with it. And then the renaming thing is a non-issue, so perhaps we don't have to bother?

rosslwheeler: I'd like to see if we can figure out the Cuda Linker output file issue - I haven't given up hope yet. Will report back on the investigation either in this PR or the CI issue we have.

Makefile
rosslwheeler: I've got this fixed - will update shortly.

rosslwheeler:
Changes made in https://github.com/karpathy/llm.c/pull/256

PR #238, opened 2024-04-24
CUDA precision configurable via build

Ricardicus:
Defaults to BF16 just like it is today.

One can configure the precision via the build now, for example like this:

```
ENABLE_BF16=1 ENABLE_FP16=0 make train_gpt2cu
```


Ricardicus:
Perhaps this removes the need to the _fp32.cu files?

karpathy:
I currently want to keep the fp32 version as a simple checkpoint in root. We'll continue developing the mixed version and I think it will drift and probably complexify more.

For mixed version we should set the best defaults but allow optional overrides. Current master does not allow overrides, so this PR adds them, which is nice. Trying it out.

karpathy:
This branch breaks `make traingpt2cu` for me:

```
make train_gpt2cu
OpenMP found, compiling with OpenMP support
nvcc found, including CUDA builds
CUDA flags: -O3 --use_fast_math -Xcompiler -DENABLE_BF16
/usr/bin/nvcc -O3 --use_fast_math -Xcompiler -DENABLE_BF16 train_gpt2.cu -lcublas -lcublasLt -o train_gpt2cu
train_gpt2.cu(793): error: argument of type "floatN *" is incompatible with parameter of type "__nv_bfloat16 *"
          atomicAddX(&dbias[i], (Tparams)dbias_shared[i]);
```

Ricardicus:
Ok! I went a little too far with the defines. I think it should be fixed now.

Ricardicus:
If this is interesting I can add these build variants to the CI before merging also

Issue #202, opened 2024-04-21
[.gitignore] gitignore does not ignore all binary files


ankitaggarwal011:
`gpt2_tokenizer.bin` and other generated binary files `train_gpt2`, `test_gpt2`, etc are not ignored in the working git directory. One way can be to output all generated binary files to `bin/` directory and add `bin/` to `.gitignore`, though adding binary files individually to `.gitignore` might be sufficient?

rosslwheeler:
If you're looking for a list which includes MSVC build artifacts, you can find them here: https://github.com/karpathy/llm.c/pull/134/files

dagelf:
Would be better to update the paths to `data/` which is already ignored: `sed -i s@gpt2_@data/gpt2_@g *`

dagelf:
Lets see if *.bin is ok #226

PR #239, opened 2024-04-24
Add to .gitignore: fp32 binaries, *.exe, *.obj, .venv

dagelf:
Has anyone ever had any issues just ignoring .*?

Also added minimal msvc ignores.


karpathy:
I don't love ignoring .* and then exceptioning things. Scared for that strategy.

dagelf:
I'm sure you have good reason, would love to know why!

I have researched it and not found much written about it. IMHO .* are hidden because you don't really want to see them. 

Seems so much cleaner than having a lot of lines in .gitignore?

I personally don't care about it because I don't use vscode and all the addons, yet...  but having tried it, I can see the appeal of having a good .gitignore and being able to run `git status`.

dagelf:
But the '!.gitignore' looks ugly... so for aesthetical reasons... I can see that!

dagelf:
Closes #202

PR #240, opened 2024-04-24
Implemented float4 dtype for parrallel memory reads for gelu_forward

ChrisDryden:
This PR was implemented as part of this working session: https://youtu.be/eOOjKTlLY-s

The kernel is currently memory limited in the main training program and it was suggested to use float4's to be able to do concurrent reads from memory in a single operation to reduce the amount of warp stalls. When tested with the main training program it caused around a .5/ms reduction in iter speed

### DIFF
--- a/dev/cuda/gelu_forward.cu
+++ b/dev/cuda/gelu_forward.cu
@@ -18,6 +18,10 @@ version 1 is naive port from CPU code to kernel
 #include <cuda_runtime.h>
 #include "common.h"
 
+__device__ float& vec_at(float4& vec, int index) {
+    return reinterpret_cast<float*>(&vec)[index];
+}
+
 // ----------------------------------------------------------------------------
 // CPU code reference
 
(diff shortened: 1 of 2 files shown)


train_gpt2_fp32.cu
```diff
@@ -977,8 +986,8 @@ void residual_forward(float* out, float* inp1, float* inp2, int N) {
 
 void gelu_forward(float* out, const float* inp, int N) {
     const int block_size = 128;
-    const int grid_size = CEIL_DIV(N, block_size);
-    gelu_forward_kernel<<<grid_size, block_size>>>(out, inp, N);
+    const int grid_size = CEIL_DIV(N/4, block_size);
```
karpathy, line 989: can we assert N % 4

dev/cuda/gelu_forward.cu
```diff
@@ -44,6 +48,25 @@ __global__ void gelu_kernel(float* out, const float* inp, int N) {
     }
 }
 
+// Optimized GELU kernel using float4
+__global__ void gelu_kernel2(float4* out, const float4* inp, int N) {
+    int idx = (blockIdx.x * blockDim.x + threadIdx.x); 
+    int i = idx * 4; // Each thread handles 4 floats
+    if (i < N) {
```
karpathy, line 55: if we assert N % 4 we can remove both if guards, which are annoying and ugly and create identation.

PR #248, opened 2024-04-24
[Multi-GPU] llm.c now runs on multiple GPUs with NCCL

PeterZhizhin:


karpathy:
Very exciting! Looking forward to step through this in detail tomorrow.

train_gpt2.cu
```diff
@@ -1898,6 +2024,41 @@ void gpt2_backward(GPT2 *model) {
...
+}
+
+// Averages out the loss and gradients across all GPUs. No-op when multi-GPU is disabled.
+void gpt2_mutli_gpu_accumulate(GPT2* model, MultiGpuConfig* multi_gpu_config) {
+    // Average all losses.
+    model->accumulated_mean_loss = multi_gpu_cpu_float_mean(model->mean_loss, multi_gpu_config);
+#ifdef MULTI_GPU
+    // Average all gradients.
+    char* grads_memory_iterator = (char*)model->grads_memory;
+    for (int i = 0; i < NUM_PARAMETER_TENSORS; ++i) {
```
karpathy, line 2046: ugh really sad to have this loop here :( with @ngc92 changed this will be just two calls. but ideally it would be 1.

karpathy:
Sadly I'm not able to test these changes yet because Lambda brought down my box during a re-image :(. I expect I'll get it back later today and will run & take a closer look.

train_gpt2.cu
PeterZhizhin: I agree it is ugly. Looking forward to the rearrangements.

I think we would need at least 2 calls, as long as we have both 16 and 32 bit gradients.

ademeure:
640K tokens/s on 8xA100 with -np 8 -b4 (BF16 mode)! Perf drops from about ~82K to ~80K per GPU when going multi-GPU which seems really good for a first implementation.

I needed to fix 2 small issues to get it working:

1.  "all_hostsname_hashes" is not initialised before the MPI_Allgather() ==> I added `all_hostsname_hashes[process_rank] = hostname_hash;` but not 100% sure that's the correct way to initialise it.

2. It was running inference on all GPUs for the last step (which generates a confusing mess repeating every word 8x) because of the priority of || vs && in conditionals, need to use () like this: `if (multi_gpu_config.process_rank == 0 && (step > 0 && (step % sample_every) == 0 || last_step)) {`

rosslwheeler:
@PeterZhizhin - ran into an interesting problem with the multi-gpu change above in train_gpt2.cu. Designated initializers (been in GCC C for a while now) are not supported in C++ (e.g. Cuda) until C++ 20. Microsoft's C++ compiler errors out the code below UNLESS you specify C++ 20 which is arguably correct. 

Question: would you be okay if we change this code to simple C code (see diff below)? This keeps us from forcing the NVCC builds on Windows to C++ 20.

image

I can do the PR if you're okay with the change? Thanks!

Issue #242, opened 2024-04-24
CI Mac issue with resources for Python?


rosslwheeler:
@Ricardicus - I've been seeing this for a few days now. Is this a GithHub limitation with the Mac runner? Below is a partial log from the CI run. I was just updating a branch in my fork and saw this again. Figured I'd drop a note in case it's new.
```

2024-04-24T17:47:08.7084440Z Installing collected packages: mpmath, urllib3, typing-extensions, tqdm, sympy, safetensors, regex, pyyaml, packaging, numpy, networkx, MarkupSafe, idna, fsspec, filelock, charset-normalizer, requests, jinja2, torch, tiktoken, huggingface-hub, tokenizers, transformers
2024-04-24T17:47:27.0972560Z Successfully installed MarkupSafe-2.1.5 charset-normalizer-3.3.2 filelock-3.13.4 fsspec-2024.3.1 huggingface-hub-0.22.2 idna-3.7 jinja2-3.1.3 mpmath-1.3.0 networkx-3.3 numpy-1.26.4 packaging-24.0 pyyaml-6.0.1 regex-2024.4.16 requests-2.31.0 safetensors-0.4.3 sympy-1.12 tiktoken-0.6.0 tokenizers-0.19.1 torch-2.3.0 tqdm-4.66.2 transformers-4.40.1 typing-extensions-4.11.0 urllib3-2.2.1
2024-04-24T17:47:27.6409360Z ##[group]Run python prepro_tinyshakespeare.py
2024-04-24T17:47:27.6409790Z [36;1mpython prepro_tinyshakespeare.py[0m
2024-04-24T17:47:27.6618450Z shell: /bin/bash -e {0}
2024-04-24T17:47:27.6618710Z ##[endgroup]
2024-04-24T17:47:29.2269560Z Downloading https://raw.githubusercontent.com/karpathy/char-rnn/master/data/tinyshakespeare/input.txt to data/tiny_shakespeare.txt...
2024-04-24T17:47:29.2283770Z 
2024-04-24T17:47:29.2415120Z data/tiny_shakespeare.txt:   0%|          | 0.00/425k [00:00<?, ?iB/s]
2024-04-24T17:47:29.2420860Z data/tiny_shakespeare.txt: 1.06MiB [00:00, 71.0MiB/s]                 
2024-04-24T17:47:29.3844710Z Saved 32768 tokens to data/tiny_shakespeare_val.bin
[...]

PR #251, opened 2024-04-25
Minor device settable edit

Ricardicus:
In the CI, we run train_gpt2.py on Mac.
But lately some odd error appears. It looks like this:

```
2024-04-24T17:47:45.8970760Z RuntimeError: MPS backend out of memory (MPS allocated: 0 bytes, other allocations: 0 bytes, max allowed: 7.93 GB). Tried to allocate 147.24 MB on private pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).
2024-04-24T17:47:46.2765760Z ##[error]Process completed with exit code 1.
```

Before then we had this situation:

```
/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/torch/backends/mps/init.py:22: UserWarning: Skipping device Apple Paravirtual device that does not support Metal 2.0 (Triggered internally at /Users/runner/work/pytorch/pytorch/pytorch/aten/src/ATen/mps/MPSDevice.mm:101.)
return torch._C._mps_is_available()
Running pytorch 2.2.2
using device: cpu
```

It means that we could not really use Metal in the virtual CI workflow, our device became selected to cpu.
Now, it seems pytorch in some cases lets us use Metal, and we end up in that situation where when the API calls are done by Pytorch; we get errors.

So in order to specify that we actually want to run on the CPU (even whn MPS is what appears to be falsely "marked" as available) I made this little change.

I know you have a history with argparse, so I only did this via environment variables.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
         run: python prepro_tinyshakespeare.py
 
       - name: Train model
-        run: python train_gpt2.py
+        run: python train_gpt2.py --device cpu
 
       - name: Compile training and testing program
         run: make test_gpt2 train_gpt2
--- a/train_gpt2.py
+++ b/train_gpt2.py
@@ -328,6 +328,8 @@ if __name__ == "__main__":
     parser.add_argument("--num_iterations", type=int, default=10, help="number of iterations to run")
     parser.add_argument("--batch_size", type=int, default=4, help="batch size")
     parser.add_argument("--sequence_length", type=int, default=64, help="sequence length")
+    parser.add_argument("--device", type=str, default=None, help="device to use (e.g., 'cpu', 'cuda:0')")
+
     args = parser.parse_args()
     B, T = args.batch_size, args.sequence_length
     assert 1 <= T <= 1024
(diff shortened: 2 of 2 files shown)


Ricardicus:
It does doesn't it?

It is something that happens when we run this in the workflow action, see the logs here:
https://github.com/Ricardicus/llm.c/actions/runs/8827724123/job/24235441372

@rosslwheeler also noticed the same on his fork, see this issue: https://github.com/karpathy/llm.c/issues/242. Might be a matter of time before it happens here? There is nothing wrong with the code, one should be able to ask torch if the MPS backend is available.

Ricardicus:
In the builds that succeed, the mps available return False and we get this warning: "Skipping device Apple Paravirtual device that does not support Metal 2.0". It is interesting, it seems like on our forks this check returns True and we go ahead using metal even if it isn't really possible.

karpathy:
And why can't we just add a new flag --device to our argparse? that overrides an autodetect?

Ricardicus:
I threw it in as a --device argument.

karpathy:
pushed the fix latest commit.

Ricardicus:
One use case for --device (that is not just to get around a CI bug) is if you want to compare CPU performance with the accelerator at hand.

PR #252, opened 2024-04-25
reorder weights according to their precision

ngc92:
Simplify our logic by keeping weights of the same precision close together.

(If we want to go with this, we also need to update the fp32 network to match; hence, for now this is a Draft PR)

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -1325,25 +1325,29 @@ typedef struct {
 } GPT2Config;
 
 // the parameters of the model
-#define NUM_PARAMETER_TENSORS 16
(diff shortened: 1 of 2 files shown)


karpathy:
Random thought I was already going to update how we write mixed precision tensors to file, reading inside mixed precision program. We could couple this change with that one, then we don't have to update the fp32 part of the code.

This is not just a code refactor, afaict this would speed up the code because our adam kernels run over more elements at a time?

karpathy:
seems like we should be able to directly load the version 2 files saved from .py file now on master! will check tomorrow.

PR #257, opened 2024-04-26
train_gtp2.c accesses parameters as arrays

zocterminal:
Would you be interested in this patch, where I changed many of the functions that receive multidimensional arrays as parameters.  I believe, since the C code seems mainly as an introduction to understand the process, this would make things much easier to read.

So, instead of doing the multiplications like in the original source, I'm declaring pointers to multidimensional arrays from the parameter, so that the source can use C array syntax like: 

```
old:      float* out_bt = out + b * T * C + t * C;

new:      float* out_bt = &outBTC[b][t][0];
```

or 

```
old:          int ix = inp[b * T + t];

new:          const int ix = inpBT[b][t];
```

There are a few still missing, but if you are interested, I would complete this.

Markus

### DIFF
--- a/train_gpt2.c
+++ b/train_gpt2.c
@@ -21,27 +21,43 @@ There will be other versions of this code that specialize it and make it fast.
 #include <omp.h>
 #endif
 
+#define DECL_ARRAYPTR2(type,name,D1,D2,init) \
+                        type (*name)[D2]= (type (*)[D2])init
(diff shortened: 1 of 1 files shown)


karpathy:
I don't super like it :) and I think it might look a bit less educational even because it obfuscates the fact that this is all just simple pointer arithmetic

zocterminal:
No worries then, I understand everyone has his/her own style and perception of code.

PR #264, opened 2024-04-27
enable padding in model export/import for nicer shapes

ngc92:
a  new attempt at this. Less ugliness on the C side because we just pad from python.


karpathy:
referenced this PR in detail, but made a lot of changes as well, esp considering the other commit merge from this morning.
resulting push is in one commit here:
https://github.com/karpathy/llm.c/commit/835060e18d1ceeb6c19ad1c214b9f03cc13daf16
closing this one.

PR #265, opened 2024-04-27
load bf16 directly, and some "quality of life" handling of fp32/fp16/bf16 precisions

karpathy:
Code to load bf16 weights directly, and also re-wire the position of tensors to put the layernorms (which are in fp32) at the end. the training loop seems to work ok, and the tests pass and the loss and optimization looks ok, but the gradients don't match. which can't be right. so there is a bug, but it's a bit too late in the day for me to debug right now, creating a PR and going to sleep, will fix tomorrow

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,7 +128,7 @@ train_gpt2fp32cu: train_gpt2_fp32.cu
 	$(NVCC) $(NVCC_FLAGS) $< $(NVCC_LDFLAGS) $(NVCC_INCLUDES) $(NVCC_LDLIBS) $(NVCC_LDFLAGS) -o $@
 
 test_gpt2cu: test_gpt2.cu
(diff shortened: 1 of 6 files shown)


karpathy:
oh and afaik this is an "inline merge" of #252 . if this merges, #252 can close.

karpathy:
I marked this from DRAFT to PR because it is technically done afaik, and possibly could merge.
I'm doing a de-risk comparing fp32 and bf16 on 1 epoch of TinyStories with this branch and will post results before merging.

karpathy:
the only new functionality technically now is that .py writes bf16 file directly, and C loads it directly, if in bf16.
this means that we can load/store our models and checkpoints in half the size => 2X faster reads and smaller file sizes on disk

sadly, test_gpt2.cu I had to 3X some of the tolerances for reasons I don't understand, as this change should be a total noop, I just re-shuffled the memory around. This now makes me feel a bit uncomfortable again and like there is still some bug lurking...

the layernorms remain in their old places

PR #266, opened 2024-04-27
Add a local convenience Makefile for dev/cuda/

austinvhuang:
Adds a local convenience Makefile for dev/cuda/ with individual targets, `all`, `run`, and `clean`. The compiler invocations largely follow the header comments for each target. One minor exception is `matmul_backward_bias.cu` (removed the `-lineinfo` flag in the Makefile).

While building the `all` target, there were two minor warnings for `encoder_backward.cu` and `layernorm_backward.cu` which are also patched here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ dev/cuda/*_backward
 dev/cuda/classifier_fused
 dev/cuda/adamw
 dev/cuda/matmul_backward_bias
+dev/cuda/nccl_all_reduce
 *.obj
 *.exe
 
--- a/dev/cuda/Makefile
+++ b/dev/cuda/Makefile
@@ -1,0 +1,88 @@
+NVCC := $(shell which nvcc 2>/dev/null)
+
+ifeq ($(NVCC),)
+		$(error nvcc not found.)
+endif
+
+TARGETS = adamw attention_backward attention_forward classifier_fused crossentropy_forward crossentropy_softmax_backward encoder_backward encoder_forward gelu_forward layernorm_backward layernorm_forward matmul_backward matmul_backward_bias matmul_forward nccl_all_reduce residual_forward softmax_forward trimat_forward
+
+CFLAGS = -O3 --use_fast_math
+
+all: $(TARGETS)
+
+adamw: adamw.cu
+	$(NVCC) $(CFLAGS) adamw.cu -o adamw
+
+attention_backward: attention_backward.cu
+	$(NVCC) $(CFLAGS) attention_backward.cu -o attention_backward -lcublas
+
+attention_forward: attention_forward.cu
+	$(NVCC) $(CFLAGS) attention_forward.cu -o attention_forward -lcublas
+
+classifier_fused: classifier_fused.cu
+	$(NVCC) $(CFLAGS) classifier_fused.cu -o classifier_fused
+
+crossentropy_forward: crossentropy_forward.cu
+	$(NVCC) $(CFLAGS) crossentropy_forward.cu -o crossentropy_forward
+
+crossentropy_softmax_backward: crossentropy_softmax_backward.cu
+	$(NVCC) $(CFLAGS) crossentropy_softmax_backward.cu -o crossentropy_softmax_backward
+
+encoder_backward: encoder_backward.cu
+	$(NVCC) $(CFLAGS) encoder_backward.cu -o encoder_backward
+
+encoder_forward: encoder_forward.cu
+	$(NVCC) $(CFLAGS) encoder_forward.cu -o encoder_forward
+
+gelu_forward: gelu_forward.cu
+	$(NVCC) $(CFLAGS) gelu_forward.cu -o gelu_forward
+
+layernorm_backward: layernorm_backward.cu
+	$(NVCC) $(CFLAGS) layernorm_backward.cu -o layernorm_backward
+
+layernorm_forward: layernorm_forward.cu
+	$(NVCC) $(CFLAGS) layernorm_forward.cu -o layernorm_forward
+
+matmul_backward: matmul_backward.cu
+	$(NVCC) $(CFLAGS) -Xcompiler -fopenmp matmul_backward.cu -o matmul_backward -lcublas
+
+matmul_backward_bias: matmul_backward_bias.cu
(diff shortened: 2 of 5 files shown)


rosslwheeler:
@austinvhuang - you can use a Make pattern rule here shortening and simplifying the Makefile. I can do an even shorter version of the whole file if you're interested. This is just the middle part as an example. I assume linking cublas is okay. The linker should not link libraries if they are not needed. 

```
%: %.cu
	$(NVCC) $(CFLAGS) $< -o $@ -lcublas

adamw: adamw.cu
classifier_fused: classifier_fused.cu
crossentropy_forward: crossentropy_forward.cu
crossentropy_softmax_backward: crossentropy_softmax_backward.cu
encoder_backward: encoder_backward.cu
encoder_forward: encoder_forward.cu
gelu_forward: gelu_forward.cu
layernorm_backward: layernorm_backward.cu
layernorm_forward: layernorm_forward.cu
residual_forward: residual_forward.cu
softmax_forward: softmax_forward.cu
matmul_backward_bias: matmul_backward_bias.cu
attention_forward: attention_forward.cu
trimat_forward: trimat_forward.cu

matmul_backward: matmul_backward.cu
	$(NVCC) $(CFLAGS) -Xcompiler -fopenmp matmul_backward.cu -o matmul_backward -lcublas

matmul_forward: matmul_forward.cu
	$(NVCC) $(CFLAGS) -Xcompiler -fopenmp matmul_forward.cu -o matmul_forward -lcublas -lcublasLt

nccl_all_reduce: nccl_all_reduce.cu
	$(NVCC) -lmpi -lnccl -I/usr/lib/x86_64-linux-gnu/openmpi/include -L/usr/lib/x86_64-linux-gnu/openmpi/lib/ nccl_all_reduce.cu -o nccl_all_reduce

clean:
	rm -f $(TARGETS)
```

karpathy:
Great!

- I like the @rosslwheeler ref
- I'd like to group these semantically, maybe the forward kernels, backward kernels, updates, nccls
- Some comments, or I am also happy to add them after

austinvhuang:
Updated the Makefile to use the suggested pattern rule, grouped kernels into forward/backward/updates/nccls. Feel free to make additional changes - the targets could be further organized (eg according to roughly where they appear in the model) but for now I just used alphabetical ordering within each category. classifier_fused.cu has both forward and backward code in it so doesn't quite work with this grouping. For now I just lumped it in with the forward kernels.

Updated the dev/cuda/README.md as well.

rosslwheeler:
@austinvhuang - looks good! If you're interested in an even more optimized - fewer lines version - it is below. It takes advantage of the consistent pattern in the directory. Added bonus - this will allow for a new cu file to be introduced and tested using the Makefile without Makefile changes if it uses the same compilation recipe as the majority of cu files (CU_FILES).  

```
# Makefile for building dev/cuda kernels

# Find nvcc
NVCC := $(shell which nvcc 2>/dev/null)
ifeq ($(NVCC),)
		$(error nvcc not found.)
endif

# Compiler flags
CFLAGS = -O3 --use_fast_math
MPI_PATHS = -I/usr/lib/x86_64-linux-gnu/openmpi/include -L/usr/lib/x86_64-linux-gnu/openmpi/lib/ 

SPECIAL_CU_FILES = matmul_forward.cu matmul_backward.cu nccl_all_reduce.cu
SPECIAL_CU_TARGETS := $(SPECIAL_CU_FILES:%.cu=%)
CU_FILES := $(filter-out $(SPECIAL_CU_FILES), $(wildcard *.cu))
CU_TARGETS := $(CU_FILES:%.cu=%)

all: $(CU_TARGETS) $(SPECIAL_CU_TARGETS)

$(CU_TARGETS): %: %.cu
	$(NVCC) $(CFLAGS) $< -o $@ -lcublas

# Forward kernels - attention_forward classifier_fused crossentropy_forward encoder_forward gelu_forward layernorm_forward residual_forward softmax_forward trimat_forward matmul_forward
matmul_forward: matmul_forward.cu
	$(NVCC) $(CFLAGS) -Xcompiler -fopenmp matmul_forward.cu -o matmul_forward -lcublas -lcublasLt

# Backward kernels - attention_backward crossentropy_softmax_backward encoder_backward layernorm_backward matmul_backward matmul_backward_bias matmul_backward
matmul_backward: matmul_backward.cu
	$(NVCC) $(CFLAGS) -Xcompiler -fopenmp matmul_backward.cu -o matmul_backward -lcublas

# Update kernels - adamw

# NCCL - nccl_all_reduce

nccl_all_reduce: nccl_all_reduce.cu
	$(NVCC) -lmpi -lnccl $(MPI_PATHS) nccl_all_reduce.cu -o nccl_all_reduce

run_all: all
	@for target in $(TARGETS); do \
		echo "\n========================================"; \
		echo "Running $$target ..."; \
		echo "========================================\n"; \
		./$$target; \
	done

clean:
	rm -f $(CU_TARGETS) $(SPECIAL_CU_TARGETS)


```

ngc92:
I'm not sure of the usefulness of `run_all` as it stands now---because this will just test kernel version 1 for each kernel, which is generally quite slow and not really relevant except for providing easy-to-read reference code.

Maybe we should just change the default from executing kernel version 1 to executing the most relevant kernel.

austinvhuang:
I have a slight bias for being explicit over a `$(wildcard *.cu)` because the surface area becomes open ended in a way that could break in implicit way eg if a new source file gets added to the repo that requires a specific library to build - `make all` then breaks in spite of no changes having been made to the Makefile. Also discoverability is a little worse with the implicitness. That said, in this case the universe of kernels seems is mostly fixed and small so I'm okay with going ahead and incorporating this.

re: run_all - we could have a convention where the default no args runs the known best kernel or (heuristically) the highest number kernel. I'll leave that out of scope for this PR unless others really want to see that here.

One other change I'll make in the next commit is to add `-lineinfo` per @ngc92 's suggestion in the discord.

ngc92:
+1 for being explicit.  I occasionally have  other  .cu files in that directory

karpathy:
agree with discussion above merging

rosslwheeler:
Yep, agree - being explicit is a help.

PR #267, opened 2024-04-27
Possible refactor to reduce CUDA boilerplate

ademeure:
I was trying to think of how to simplify the FP32 file, but the main thing I came up with is generic and applicable to both paths. If you really hate either macros or C++ templates, you might not like it, but it does get rid of a fair bit of boilerplate, mostly through 2 changes:

1) Macros to initialise idx in a standard way for every kernel:

`#define THREAD_INIT(n)       int idx = blockIdx.x * blockDim.x + threadIdx.x; \`
`                             if (idx >= n) { return; }`
`#define THREAD_REVERSE(n)    int idx = (gridDim.x-blockIdx.x-1) * blockDim.x + threadIdx.x; \`
`                             if (idx >= n) { return; }`
`#define WARP_MODE_FORWARD(n) int idx = blockIdx.x * blockDim.x/32 + threadIdx.x/32; \`
`                             if (idx*32 >= n) { return; }`
`#define WARP_MODE_REVERSE(n) int idx = (gridDim.x-blockIdx.x-1) * blockDim.x/32 + threadIdx.x/32; \`
`                             if (idx*32 >= n) { return; }`

And used in a kernel function:
`__global__ void residual_forward_kernel(int N, float* out, float* inp1, float* inp2) {`
`    THREAD_INIT(N)`
`    out[idx] = __ldcs(&inp1[idx]) + __ldcs(&inp2[idx]);`
`}`
Combined with the 2nd change below, the input to these macros is always N, but the changes are really orthogonal, and you could pass "B * NH" for example.

2) New templated functions to call CUDA kernels (note the << >> calling scheme in CUDA is already C++ templates anyway...)

`template <int block_size = 256, typename KernelFunc, typename... Args>`
`void launch(KernelFunc kernel, int num_threads, Args... args) {`
`    const int grid_size = CEIL_DIV(num_threads, block_size);`
`    kernel<<<grid_size, block_size>>>(num_threads, args...);`
`    cudaCheck(cudaGetLastError());`
`}`
`template <int block_size = 256, typename KernelFunc, typename... Args>`
`void launch_shared(KernelFunc kernel, int num_threads, int shared_memory, Args... args) {`
`    const int grid_size = CEIL_DIV(num_threads, block_size);`
`    kernel<<<grid_size, block_size, shared_memory>>>(num_threads, args...);`
`    cudaCheck(cudaGetLastError());`
`}`

Which can then be used this way directly without a boilerplate function per-kernel:

`launch<512>(residual_forward_kernel, B*T*C, l_residual3, l_residual2, l_fcproj);`

3) Also, I reordered the function arguments to have integers first, then outputs, then inputs - which looks cleaner when the first argument is the number of threads, but led to a lot of lines of code being slightly changed and increasing the size of the diff a lot...

Possibly an argument against these changes are that they bring everything further away from "raw CUDA" in a way that isn't standardised like the CUDA C++ libraries. Also, it's a bit late doing this kind of thing after splitting the FP32 path from the other one, since a lot of this should apply to both, and that's going to be "fun"...

Possibly just a smaller diff doing only (1) would make sense, since that just feels better than what we have now, where different kernels randomly do either "if (idx >= N) return;" or "if (idx < N) { ... }" etc... I don't think there's a good way to do this without macros though. If we wanted, it would also allow defining some helper types like warp_id and lane_id automatically for every kernel.

### DIFF
--- a/train_gpt2_fp32.cu
+++ b/train_gpt2_fp32.cu
@@ -27,40 +27,60 @@ the layernorms are connected to the residuals so we += in layernorm backward.
 #include <cooperative_groups.h>
 #include <cooperative_groups/reduce.h>
 
+namespace cg = cooperative_groups;
 // ----------------------------------------------------------------------------
 // CUDA utils
+// These helpers initialise the index of the element being computed
+// Warp mode allows one "element" per threadblock using threadIdx.x internally
+// Reverse modes are used to try to maximise cache hits from previous forward mode kernels
+#define THREAD_INIT(n)       int idx = blockIdx.x * blockDim.x + threadIdx.x; \
+                             if (idx >= n) { return; }
+#define THREAD_REVERSE(n)    int idx = (gridDim.x-blockIdx.x-1) * blockDim.x + threadIdx.x; \
+                             if (idx >= n) { return; }
+#define WARP_MODE_FORWARD(n) int idx = blockIdx.x * blockDim.x/32 + threadIdx.x/32; \
+                             if (idx*32 >= n) { return; }
+#define WARP_MODE_REVERSE(n) int idx = (gridDim.x-blockIdx.x-1) * blockDim.x/32 + threadIdx.x/32; \
+                             if (idx*32 >= n) { return; }
+#define CEIL_DIV(M, N) (((M) + (N)-1) / (N)) // convenience for grid/block dimensions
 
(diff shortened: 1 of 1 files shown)


karpathy:
Per discussion on Discord closing this Draft PR. Prefer to keep things readable, simple, even if it's a little bit more bloated and boilerplate-y.

PR #270, opened 2024-04-27
clang-tidy

ngc92:
Adds a clang-tidy file and  clang-tidy target to the make file. 
Since the .cu files are in  flux right now, this is just looking at gpt2.c

I'm not quite sure which checks we should enable, but I think the ones that I've selected should be a good starting point.
In particular, there is one warning about using the result of 32-bit matrix multiplications as an offset in a pointer operation -- that seems like a good one to have, to prevent stuff like the high batch-size crashes.

Regarding  the "fixes" I've made to ensure we don't get these warnings, I'm not 100% sure about those, and even more so about using ptrdiff_t vs size_t. 

(using size_t for the local B, T, etc variables gets us 64-bit offset calculations, but also implementation-defined narrowing once we call the actual kernels)

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,1 @@
+Checks: '-*,bugprone-*,-bugprone-easily-swappable-parameters,cert-*-c,misc-confusable-identifiers,misc-const-correctness,performance-*'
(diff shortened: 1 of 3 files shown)


dagelf:
Wow, this is neat. Might as well add all the *.c* files!

PR #272, opened 2024-04-27
Full BF16 including layernorms by default (minimising number of BF16 atomics)

ademeure:
I added 4 different new versions of layernorm_backward_kernel, performance is best for:

- Kernel 4 (using atomicCAS, no scratch, but rounding many times so probably worse numerical accuracy
- Kernel 6 ==> new default (FP32 scratchpad, do one final rounding to BF16 at the end, could be made stochastic in the future).

We probably just want to integrate Kernel 6 but might want to add all of them to /dev/cuda/ in the future. I haven't fixed the BF16 atomics in encoder_backward yet but the performance penalty of that is much much smaller, it might be worth doing it manually using atomicCAS so we can implement stochastic rounding there as well though.

Performance on my RTX 4090 is a tiny bit faster (potentially noise) than with the previous mixed FP32/BF16.

### DIFF
--- a/dev/cuda/common.h
+++ b/dev/cuda/common.h
@@ -76,19 +76,19 @@ float* make_ones_float(size_t N) {
 // ----------------------------------------------------------------------------
 // testing and benchmarking utils
 
-template<class T>
-void validate_result(T* device_result, const T* cpu_reference, const char* name, std::size_t num_elements, T tolerance=1e-4) {
-    T* out_gpu = (T*)malloc(num_elements * sizeof(T));
-    cudaCheck(cudaMemcpy(out_gpu, device_result, num_elements * sizeof(T), cudaMemcpyDeviceToHost));
+template<class D, class T>
+void validate_result(D* device_result, const T* cpu_reference, const char* name, std::size_t num_elements, T tolerance=1e-4) {
+    D* out_gpu = (D*)malloc(num_elements * sizeof(T));
+    cudaCheck(cudaMemcpy(out_gpu, device_result, num_elements * sizeof(D), cudaMemcpyDeviceToHost));
     int nfaults = 0;
     for (int i = 0; i < num_elements; i++) {
         // print the first few comparisons
         if (i < 5) {
-            printf("%f %f\n", cpu_reference[i], out_gpu[i]);
+            printf("%f %f\n", cpu_reference[i], (T)out_gpu[i]);
         }
         // ensure correctness for all elements. We can set an "ignore" mask by writing NaN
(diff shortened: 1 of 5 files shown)


karpathy:
I'm assuming this is not meant to merge as-is? Would it make sense to put most of this into dev/cuda and then cherry-pick the layernorm we want to use, as usual, into train_gpt2.cu?

train_gpt2.cu
```diff
@@ -940,6 +944,399 @@ __global__ void layernorm_backward_kernel2(Tdinp* dinp, Tparams* dweight, Tparam
...
+    float* dbias_shared = shared;
+    float* dweight_shared = shared + C;
+
+    // init shared memory to zero
+    #pragma unroll
+    for(int i = threadIdx.x; i < C; i+= blockDim.x){
+       dbias_shared[i] = 0.0f;
+       dweight_shared[i] = 0.0f;
+    }
+    uint *tmp_flag = (uint*)(shared + C*2);
```
ngc92, line 971: is this actually used in this kernel?

train_gpt2.cu
```diff
@@ -940,6 +944,399 @@ __global__ void layernorm_backward_kernel2(Tdinp* dinp, Tparams* dweight, Tparam
...
```
ngc92, line 1048: same here

ademeure:
Thanks! Fixed those bugs, and only kept Kernel6 in train_gpt2.cu, adding all the kernels to /dev/cuda/

I think this might be the first /dev/cuda change with BF16, so I added support for comparing CPU and GPU data of different types (converting GPU data to CPU type before comparison) in validate_result().

dev/cuda/common.h
```diff
@@ -76,19 +76,19 @@ float* make_ones_float(size_t N) {
 // ----------------------------------------------------------------------------
 // testing and benchmarking utils
 
-template<class T>
-void validate_result(T* device_result, const T* cpu_reference, const char* name, std::size_t num_elements, T tolerance=1e-4) {
-    T* out_gpu = (T*)malloc(num_elements * sizeof(T));
-    cudaCheck(cudaMemcpy(out_gpu, device_result, num_elements * sizeof(T), cudaMemcpyDeviceToHost));
+template<class D, class T>
+void validate_result(D* device_result, const T* cpu_reference, const char* name, std::size_t num_elements, T tolerance=1e-4) {
+    D* out_gpu = (D*)malloc(num_elements * sizeof(T));
```
ngc92, line 81: sizeof(D)?

PR #273, opened 2024-04-28
Add NSight Compute ranges, use CUDA events for timings

PeterZhizhin:
- CUDA events allow for more accurate timings (as measured by a GPU)
- `nvtxRangePush`/`nvtxRangePop` Adds simple stack traces to NSight Systems:
  - Sample run command: `nsys profile mpirun --allow-run-as-root -np 2 ./train_gpt2cu`
  - Then inspect the resulting profile on a local machine with NVidia NSight Systems.

You'll see something like this in NSight for GPU profile:
NSight Systems sample profile

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -33,6 +33,8 @@ This reads & runs in fp32, B=4, T=64, LR=1e-4, val/sample never (200),
 -a 1 is "overfit single batch", -x 10 is 10 iterations, and -f 0 disables tf32
 */
 
+#include <string>
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdarg.h>
@@ -45,11 +47,14 @@ This reads & runs in fp32, B=4, T=64, LR=1e-4, val/sample never (200),
 #include <assert.h>
 // GPU / CUDA related
 #include <cublas_v2.h>
+#include <cuda_profiler_api.h>
 #include <cuda_runtime.h>
 #include <cublasLt.h>
 #include <cuda_bf16.h>
 #include <cooperative_groups.h>
 #include <cooperative_groups/reduce.h>
+#include <nvtx3/nvToolsExt.h>
+
 // Multi-GPU related
 #ifdef MULTI_GPU
 #include <mpi.h>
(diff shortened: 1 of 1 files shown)


ademeure:
This is really nice!

It's a fair bit of boilerplate, maybe it could be wrapped into a tiny tiny class with a constructor/destructor calling push/pop, since the push/pop are nearly always at { }. Not sure what Andrej's stylistic preferences would be here.

karpathy:
agree quite eager to get this in on a high level, but is there a way to skip all the added LOC around push/pop?

ademeure:
> agree quite eager to get this in on a high level, but is there a way to skip all the added LOC around push/pop?

what I mentioned above with NVTX3_FUNC_RANGE() means you only need 1 line for the push, instead of 2 lines (push+pop). I don't think we can get it to anything less than that. Also maybe the string manipulation could be done more cleanly in less code with C++ rather than sprintf.

karpathy:
@PeterZhizhin are you able to incorporate the above into this PR? I can also take a look if no. Also this seems to have merge conflicts now bleh

karpathy:
(Otherwise pretty eager to merge this!)

PeterZhizhin:
I can, spend my time working on a video yesterday. Will do it today.

PeterZhizhin:
Hello, @karpathy! I've rebased the PR and now I use a C++ class with a constructor-desctructor.

Using native NVTX3_FUNC_RANGE probably requires us to use CMake (https://github.com/NVIDIA/NVTX?tab=readme-ov-file#cmake). Seems like it's not bundled in with CUDA?

On my installation I get this error:
```
train_gpt2.cu:48:10: fatal error: nvtx3/nvtx3.hpp: No such file or directory
   48 | #include <nvtx3/nvtx3.hpp>
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
```

rosslwheeler:
@PeterZhizhin - we can pull that header down from the GitHub repo (or others) in the Makefile. Do you have a list of headers that you need from that repo?

karpathy:
oh i see, i didn't realize. in that case i would be ok with just keeping the two calls like before, that sounds gnarly, no pulling of anything

ademeure:
> oh i see, i didn't realize. in that case i would be ok with just keeping the two calls like before, that sounds gnarly, no pulling of anything

@PeterZhizhin I think your latest PR solves the 2 calls without needing any additional headers, right? So we should be able to use that, it seems clean enough to me (since we control the name I guess we could rename it something nicer than "NVTX_RANGE_FN" if we wanted to as well).

PeterZhizhin:
Rebased and fixed the build errors. Good to merge.

PR #275, opened 2024-04-28
Removing Atomic Adds and adding memory coalescion

ChrisDryden:
This PR is ontop of the GELU memory coalescion PR and is essentially just a rewrite of the backwards encoder to use shared memory instead of atomic adds and then using the Packed struct to do coalesced adds. This reduces the runtime for me by around 1ms.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -51,6 +51,21 @@ mpirun -np 4 ./train_gpt2cu -b 8 -v 200 -s 200 -i data/TinyStories
 // ----------------------------------------------------------------------------
 // CUDA precision settings
 
+// 128 Memory Read Packing data structure
+template<class ElementType>
+struct alignas(16) Packed128 {
+    __device__ ElementType& operator[](int index) {
+        return reinterpret_cast<ElementType*>(&payload)[index];
+    }
+
+    __device__ const ElementType& operator[](int index) const {
+        return reinterpret_cast<const ElementType*>(&payload)[index];
+    }
+
+    int4 payload;
+    static constexpr const size_t size = sizeof(int4) / sizeof(ElementType);
+};
+
 // turn on bf16 as default, done up here for now
 #define ENABLE_BF16
 
@@ -685,26 +700,37 @@ __global__ void residual_forward_kernel(TOut* out, T1* inp1, T2* inp2, int N) {
 }
 
 #define GELU_SCALING_FACTOR sqrtf(2.0f / M_PI)
-__global__ void gelu_forward_kernel(floatX* out, const floatX* inp, int N) {
+__global__ void gelu_forward_kernel(Packed128<floatX>* out, const Packed128<floatX>* inp, int N) {
     int i = blockIdx.x * blockDim.x + threadIdx.x;
-    if (i < N) {
-        float xi = (float)inp[i];
-        float cube = 0.044715f * xi * xi * xi;
-        out[i] = (floatX)(0.5f * xi * (1.0f + tanhf(GELU_SCALING_FACTOR * (xi + cube))));
+    if(i < N){ 
+        Packed128<floatX> data = inp[i];
+        Packed128<floatX> local_out;
+        for(int j = 0; j < Packed128<floatX>::size; j++){
+            float xi = (float)data[j];
+            float cube = 0.044715f * xi * xi * xi;
+            local_out[j] = (floatX)(0.5f * xi * (1.0f + tanhf(GELU_SCALING_FACTOR * (xi + cube))));
+        }
+        out[i] = local_out;
     }
 }
 
-__global__ void gelu_backward_kernel(floatX* dinp, const floatX* inp, const floatX* dout, const int N) {
+__global__ void gelu_backward_kernel(Packed128<floatX>* dinp, const Packed128<floatX>* inp, const Packed128<floatX>* dout, const int N) {
     int i = blockIdx.x * blockDim.x + threadIdx.x;
     if (i < N) {
-        float x = (float)inp[i];
-        float cube = 0.044715f * x * x * x;
-        float tanh_arg = GELU_SCALING_FACTOR * (x + cube);
-        float tanh_out = tanhf(tanh_arg);
-        float coshf_out = coshf(tanh_arg);
-        float sech_out = 1.0f / (coshf_out * coshf_out);
-        float local_grad = 0.5f * (1.0f + tanh_out) + x * 0.5f * sech_out * GELU_SCALING_FACTOR * (1.0f + 3.0f * 0.044715f * x * x);
-        dinp[i] = (floatX)(local_grad * (float)dout[i]);
+        Packed128<floatX> inp_packed = inp[i];
+        Packed128<floatX> dout_packed = dout[i];
+        Packed128<floatX> local_out;
+        for(int j = 0; j < Packed128<floatX>::size; j++){
+            float x = (float)inp_packed[j];
+            float cube = 0.044715f * x * x * x;
+            float tanh_arg = GELU_SCALING_FACTOR * (x + cube);
+            float tanh_out = tanhf(tanh_arg);
+            float coshf_out = coshf(tanh_arg);
+            float sech_out = 1.0f / (coshf_out * coshf_out);
+            float local_grad = 0.5f * (1.0f + tanh_out) + x * 0.5f * sech_out * GELU_SCALING_FACTOR * (1.0f + 3.0f * 0.044715f * x * x);
+            local_out[j] = (floatX)(local_grad * (float)dout_packed[j]);
+        }
+        dinp[i] = local_out;
     }
 }
 
@@ -1329,15 +1355,15 @@ void residual_forward(TOut* out, T1* inp1, T2* inp2, int N) {
 
 void gelu_forward(floatX* out, const floatX* inp, int N) {
     const int block_size = 128;
-    const int grid_size = CEIL_DIV(N, block_size);
-    gelu_forward_kernel<<<grid_size, block_size>>>(out, inp, N);
+    const int grid_size = CEIL_DIV(N/Packed128<floatX>::size, block_size);
+    gelu_forward_kernel<<<grid_size, block_size>>>((Packed128<floatX>*)out, (Packed128<floatX>*)inp, N);
     cudaCheck(cudaGetLastError());
 }
 
 void gelu_backward(floatX* dinp, const floatX* inp, const floatX* dout, const int N) {
     const int block_size = 128;
-    const int grid_size = CEIL_DIV(N, block_size);
-    gelu_backward_kernel<<<grid_size, block_size>>>(dinp, inp, dout, N);
+    const int grid_size = CEIL_DIV(N/Packed128<floatX>::size, block_size);
+    gelu_backward_kernel<<<grid_size, block_size>>>((Packed128<floatX>*)dinp, (Packed128<floatX>*)inp, (Packed128<floatX>*)dout, N);
     cudaCheck(cudaGetLastError());
 }
 
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -487,6 +487,8 @@ template <typename Type, typename Tdout>
 __global__ void encoder_backward_kernel(Type* dwte, Type* dwpe,
                                         const Tdout* dout, const int* inp,
                                         int B, int T, int C) {
+    extern __shared__ Type shared_mem[];
+
     int idx = blockIdx.x * blockDim.x + threadIdx.x;
     int N = B * T * C;
 
@@ -502,8 +504,21 @@ __global__ void encoder_backward_kernel(Type* dwte, Type* dwpe,
         Type* dwte_ix = dwte + ix * C + c;
         Type* dwpe_tc = dwpe + t * C + c;
 
-        atomicAddX(dwte_ix, (Type)*dout_btc);
-        atomicAddX(dwpe_tc, (Type)*dout_btc);
+        // Aggregating in shared memory
+        shared_mem[threadIdx.x] = (Type)*dout_btc;
+
+        __syncthreads(); // Wait for all threads to write to shared memory
+
+        // Perform reduction in shared memory
+        if (threadIdx.x == 0) {
+            Type sum = 0;
+            for (int i = 0; i < blockDim.x; i++) {
+                sum += shared_mem[i];
+            }
+            // Only one thread writes back to global memory
+            *dwte_ix += sum;
+            *dwpe_tc += sum;
+         }
     }
 }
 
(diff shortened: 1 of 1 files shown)


ngc92:
I'm not sure if these are actually correct right now. shouldn't the N paramter to the kernel also be divided by the packet size?

ademeure:
I’ll take a look later at the rest but since I’ve been working on those atomics yesterday, I don’t think that change is functionally correct (or can be made to worn reliably) :(

The global memory addresses that kernel is writing to are dependent on threadIdx.x, specifically/especially “c” which will be different for every thread in a warp, but this is accumulating then adding it all to the same address I think?

Fundamentally I cannot think of a way to get rid of atomicAdd there without losing nearly all the parallelism and running very slowly on the GPU. It’s effectively a sparse scatter: a small number of values are being added to a large number of memory locations. e.g. with a vocabulary size of ~50000 and 1024 tokens * 8 batch size, at least ~42000(*C) entries won’t be touched at all. But you don’t know what tokens will affect what until you read it, so unless you use atomics, threads in one threadblock on one SM might be trying to modify the same data as another thread in another SM at the same time, and its change gets lost/overwritten & the read/write are not atomic.

Thankfully those are very nicely coalesced atomics, notice how ‘c’ is effectively the index. And if the vocabulary is padded at 128 byte boundaries (64 elements), then all the 32 ‘c’ being processed by one warp in one instruction fit on the same 128 bytes cacheline (and it’s not too bad if it fits on 2 either). So the atomic performance should be very good, to the point where we should just be DRAM limited. The problem is more related to BF16 atomics being slow but I am currently thinking we could just use atomicCAS doing the addition manually in the kernel, which has the extra benefit of allowing us to do stochastic rounding.

There might be clever ways to get rid of the atomic completely by splitting the work across threads/SMs in a very specific way, but I suspect they’d cause more problems (e.g. lack of parallelism) than they’d solve unfortunately…

ngc92:
Agree re token encodings, but for the positional encodings, I think we should be able to get rid of the atomics

ademeure:
I spent a whole bunch of time looking into alternatives to this approach for memory coalescing, and as I mentioned on Discord, NVIDIA's automatic coalescing for 128-bit loads is *incredibly* finicky, and you need some horrible manually unrolled operations to make it work reliably, e.g.:
```template <typename TypeI, typename TypeO, typename TypeT>
__device__ void transfer8(TypeO* out, TypeI *inp, TypeT *tmp) {
    inp = static_cast<TypeI*>(__builtin_assume_aligned(inp, 16));
    out = static_cast<TypeO*>(__builtin_assume_aligned(out, 16));
    tmp[0] = inp[0]; tmp[1] = inp[1]; tmp[2] = inp[2]; tmp[3] = inp[3];
    out[0] = tmp[0]; out[1] = tmp[1]; out[2] = tmp[2]; out[3] = tmp[3];
    tmp[4] = inp[4]; tmp[5] = inp[5]; tmp[6] = inp[6]; tmp[7] = inp[7];
    out[4] = tmp[4]; out[5] = tmp[5]; out[6] = tmp[6]; out[7] = tmp[7];
}
template <typename TypeI, typename TypeO>
__device__ void fetch8(TypeO* out, TypeI *inp) {
    typename std::remove_const<TypeI>::type tmp[8];
    transfer8(out, inp, tmp);
}
template <typename TypeI, typename TypeO>
__device__ void write8(TypeO *out, TypeI* inp) {
    TypeO tmp[8];
    transfer8(out, inp, tmp);
}
```

So I think the approach in this PR is possibly just as good or better than that abomination.

But we should be able to massively reduce the amount of boilerplate and changes to function declarations in the PR, while also making the float<->floatX conversions in the current code cleaner. Here's an example of what I would propose at the cost of a slightly bigger Packed128 class declaration but simpler/cleaner kernels:

```template<class ElementType>
struct alignas(16) Packed128 {
    __device__ Packed128() {}
    __device__ Packed128(const ElementType* data, size_t offset128=0) {
        payload = reinterpret_cast<const int4*>(data)[offset128];
    }
    __device__ ElementType& operator[](int index) {
        return reinterpret_cast<ElementType*>(&payload)[index];
    }
    __device__ const ElementType& operator[](int index) const {
        return reinterpret_cast<const ElementType*>(&payload)[index];
    }
    __device__ float fp32(int index) {
        return static_cast<float>(reinterpret_cast<ElementType*>(&payload)[index]);
    }
    __device__ void write(ElementType* data, size_t offset128=0) {
        reinterpret_cast<int4*>(data)[offset128] = payload;
    }
    static constexpr const size_t size = sizeof(int4) / sizeof(ElementType);
    int4 payload;
};
typedef Packed128<floatX> f128;

__global__ void residual_forward_kernel(floatX* out, floatX* inp1, floatX* inp2, int N) {
    int idx = blockIdx.x * blockDim.x + threadIdx.x;
    if (idx >= N) { return; }

    f128 inp1_packed(inp1, idx);
    f128 inp2_packed(inp2, idx);
    f128 out_packed;
    for(int i = 0; i < f128::size; i++){
        out_packed[i] = inp1_packed.fp32(i) + inp2_packed.fp32(i);
    }
    out_packed.write(out, idx);
}

#define GELU_SCALING_FACTOR sqrtf(2.0f / M_PI)
__global__ void gelu_forward_kernel(floatX* out, const floatX* inp, int N) {
    int idx = blockIdx.x * blockDim.x + threadIdx.x;
    if(idx >= N) { return; }

    f128 inp_packed(inp, idx);
    f128 out_packed;
    for(int i = 0; i < f128::size; i++){
        float xi = inp_packed.fp32(i);
        float cube = 0.044715f * xi * xi * xi;
        out_packed[i] = (0.5f * xi * (1.0f + tanhf(GELU_SCALING_FACTOR * (xi + cube))));
    }
    out_packed.write(out, idx);
}
```
Let me know what you think! @karpathy does this look okay to you stylistically? It definitely makes the GPU go brrrrrrrr: the SASS for residual_forward_kernel is a thing of simple beauty after this, there's not a single wasted instruction as far as I can tell!

ademeure:
Here's an updated version that supports __ldcs, if we wanted to support __ldcg, then we'd just turn that boolean into an enum:

```cuda
constexpr bool CACHE_DEFAULT = false;
constexpr bool CACHE_STREAMING = true;

template<class ElementType>
struct alignas(16) Packed128 {
    __device__ Packed128() {}
    __device__ Packed128(const ElementType* data, size_t offset128=0, bool streaming=CACHE_DEFAULT) {
        load(data, offset128, streaming);
    }
    __device__ ElementType& operator[](int index) {
        return reinterpret_cast<ElementType*>(&payload)[index];
    }
    __device__ const ElementType& operator[](int index) const {
        return reinterpret_cast<const ElementType*>(&payload)[index];
    }
    __device__ float fp32(int index) {
        return static_cast<float>(reinterpret_cast<ElementType*>(&payload)[index]);
    }

    __device__ void load(const ElementType* data, size_t offset128=0, bool streaming=CACHE_DEFAULT) {
        const int4* ptr = &reinterpret_cast<const int4*>(data)[offset128];
        if (streaming) {
            payload = __ldcs(ptr);
        } else {
            payload = *ptr;
        }
    }
    __device__ void store(ElementType* data, size_t offset128=0, bool streaming=CACHE_DEFAULT) {
        int4* ptr = &reinterpret_cast<int4*>(data)[offset128];
        if (streaming) {
            __stcs(ptr, payload);
        } else {
            *ptr = payload;
        }
    }

    static constexpr const size_t size = sizeof(int4) / sizeof(ElementType);
    int4 payload;
};
typedef Packed128<floatX> f128;

__global__ void residual_forward_kernel(floatX* out, floatX* inp1, floatX* inp2, int N) {
    int idx = blockIdx.x * blockDim.x + threadIdx.x;
    if (idx >= N) { return; }

    f128 inp1_packed(inp1, idx, CACHE_STREAMING);
    f128 inp2_packed(inp2, idx, CACHE_STREAMING);
    f128 out_packed;
    for(int i = 0; i < f128::size; i++){
        out_packed[i] = inp1_packed.fp32(i) + inp2_packed.fp32(i);
    }
    out_packed.store(out, idx);
}

#define GELU_SCALING_FACTOR sqrtf(2.0f / M_PI)
__global__ void gelu_forward_kernel(floatX* out, const floatX* inp, int N) {
    int idx = blockIdx.x * blockDim.x + threadIdx.x;
    if(idx >= N) { return; }

    // alternative init using load() rather than the constructor if you prefer it being explicit
    f128 inp_packed, out_packed;
    inp_packed.load(inp, idx, CACHE_STREAMING);

    for(int i = 0; i < f128::size; i++){
        float xi = inp_packed.fp32(i);
        float cube = 0.044715f * xi * xi * xi;
        out_packed[i] = (0.5f * xi * (1.0f + tanhf(GELU_SCALING_FACTOR * (xi + cube))));
    }
    // this shouldn't actually be streaming, just putting this here as an example of how it's used
    out_packed.store(out, idx, CACHE_STREAMING);
}

void residual_forward(floatX* out, floatX* inp1, floatX* inp2, int N) {
    N /= f128::size;
    const int block_size = 256;
    const int grid_size = CEIL_DIV(N, block_size);
    residual_forward_kernel<<<grid_size, block_size>>>(out, inp1, inp2, N);
    cudaCheck(cudaGetLastError());
}

void gelu_forward(floatX* out, const floatX* inp, int N) {
    N /= f128::size;
    const int block_size = 128;
    const int grid_size = CEIL_DIV(N, block_size);
    gelu_forward_kernel<<<grid_size, block_size>>>(out, inp, N);
    cudaCheck(cudaGetLastError());
}
```

ademeure:
And yet another even simpler version that uses a template for "streaming" and a "f128_cs" type:

```cuda
template<class ElementType, bool streaming=false>
struct alignas(16) Packed128 {
    __device__ Packed128() {}
    __device__ Packed128(const ElementType* data, size_t offset128=0) {
        load(data, offset128);
    }
    __device__ ElementType& operator[](int index) {
        return reinterpret_cast<ElementType*>(&payload)[index];
    }
    __device__ const ElementType& operator[](int index) const {
        return reinterpret_cast<const ElementType*>(&payload)[index];
    }
    __device__ float fp32(int index) {
        return static_cast<float>(reinterpret_cast<ElementType*>(&payload)[index]);
    }

    __device__ void load(const ElementType* data, size_t offset128=0) {
        const int4* ptr = &reinterpret_cast<const int4*>(data)[offset128];
        if (streaming) {
            payload = __ldcs(ptr);
        } else {
            payload = *ptr;
        }
    }
    __device__ void store(ElementType* data, size_t offset128=0) {
        int4* ptr = &reinterpret_cast<int4*>(data)[offset128];
        if (streaming) {
            __stcs(ptr, payload);
        } else {
            *ptr = payload;
        }
    }
    
    static constexpr const size_t size = sizeof(int4) / sizeof(ElementType);
    int4 payload;
};
typedef Packed128<floatX, false> f128;
typedef Packed128<floatX, true>  f128_cs;

__global__ void residual_forward_kernel(floatX* out, floatX* inp1, floatX* inp2, int N) {
    int idx = blockIdx.x * blockDim.x + threadIdx.x;
    if (idx >= N) { return; }

    f128    out_packed;
    f128    inp1_packed(inp1, idx);
    f128_cs inp2_packed(inp2, idx);
    for(int i = 0; i < f128::size; i++){
        out_packed[i] = inp1_packed.fp32(i) + inp2_packed.fp32(i);
    }
    out_packed.store(out, idx);
}

#define GELU_SCALING_FACTOR sqrtf(2.0f / M_PI)
__global__ void gelu_forward_kernel(floatX* out, const floatX* inp, int N) {
    int idx = blockIdx.x * blockDim.x + threadIdx.x;
    if(idx >= N) { return; }

    // alternative init using load() rather than the constructor if you prefer it being explicit
    f128_cs inp_packed, out_packed;
    inp_packed.load(inp, idx);

    for(int i = 0; i < f128::size; i++){
        float xi = inp_packed.fp32(i);
        float cube = 0.044715f * xi * xi * xi;
        out_packed[i] = (0.5f * xi * (1.0f + tanhf(GELU_SCALING_FACTOR * (xi + cube))));
    }
    out_packed.store(out, idx); // this will be a cache streaming store since it's f128_cs
}

void residual_forward(floatX* out, floatX* inp1, floatX* inp2, int N) {
    N /= f128::size;
    const int block_size = 256;
    const int grid_size = CEIL_DIV(N, block_size);
    residual_forward_kernel<<<grid_size, block_size>>>(out, inp1, inp2, N);
    cudaCheck(cudaGetLastError());
}

void gelu_forward(floatX* out, const floatX* inp, int N) {
    N /= f128::size;
    const int block_size = 128;
    const int grid_size = CEIL_DIV(N, block_size);
    gelu_forward_kernel<<<grid_size, block_size>>>(out, inp, N);
    cudaCheck(cudaGetLastError());
}
```

ngc92:
Lets put all the options next to each other, then:
```c++

template<class ElementType>
struct alignas(16) Packed128 {
    __device__ __forceinline__ Packed128() = default;
    __device__ __forceinline__ explicit Packed128(int4 bits) {
        static_assert(sizeof(bits) == sizeof(payload), "Size mismatch.");
        memcpy(&payload, &bits, sizeof(bits));
    }

    __device__ __forceinline__ ElementType& operator[](int index) {
        return payload[index];
    }
    __device__ __forceinline__ const ElementType& operator[](int index) const {
        return payload[index];
    }
    __device__ __forceinline__ float fp32(int index) {
        return static_cast<float>(payload[index]);
    }

    __device__ __forceinline__ int4 get_bits() const {
        int4 bits;
        static_assert(sizeof(bits) == sizeof(payload), "Size mismatch.");
        memcpy(&bits, &payload, sizeof(bits));
        return bits;
    }

    static constexpr const size_t size = sizeof(int4) / sizeof(ElementType);

    ElementType payload[size];
};


// load a Packet128 from an aligned memory address
template<class ElementType>
__device__ __forceinline__ Packed128<ElementType> load_aligned(const ElementType* address) {
    return Packed128<ElementType>{*reinterpret_cast<const int4*>(address)};
}

// load a Packet128 from an aligned memory address with streaming cache hint
template<class ElementType>
__device__ __forceinline__ Packed128<ElementType> load_aligned_cs(const ElementType* address) {
    return Packed128<ElementType>{__ldcs(reinterpret_cast<const int4*>(address))};
}

// store a Packet128 to an aligned memory address
template<class ElementType>
__device__ __forceinline__ void store_aligned(ElementType* target, Packed128<ElementType> value) {
    *reinterpret_cast<int4*>(target) = value.get_bits();
}

// store a Packet128 to an aligned memory address with streaming cache hint
template<class ElementType>
__device__ __forceinline__ void store_aligned_cs(ElementType* target, Packed128<ElementType> value) {
    __stcs(reinterpret_cast<int4*>(target), value.get_bits());
}

__global__ void gelu_kernel2(float* out, const float* inp, int N) {
    using packet_t = Packed128<float>;
    int i = (blockIdx.x * blockDim.x + threadIdx.x) * packet_t::size;
    if (i < N) {
        packet_t packet_out;
        packet_t packet_in = load_aligned_cs(inp + i);
        for(int k = 0; k < packet_in.size; ++k) {
            float xi = packet_in[k];
            float cube = 0.044715f * xi * xi * xi;
            packet_out[k] = 0.5f * xi * (1.0f + tanhf(GELU_SCALING_FACTOR * (xi + cube)));
        }
        store_aligned(out + i, packet_out);
    }
}
```

ChrisDryden:
llm.c with a different approach here:
https://github.com/karpathy/llm.c/pull/298

Issue #246, opened 2024-04-24
WikiText 103 evaluation


karpathy:
I've seen some repos use WikiText-103 as the dataset they use to eval GPT-like models, e.g.:

https://github.com/tysam-code/hlb-gpt/tree/main

Add prepro script to download and preprocess and tokenize WikiText-103 just like tiny shakespeare / tiny stories, following this repo. Adapt the mainline training script `train_gpt2.cu` to report the validation performance on this set.

Add python code that does the same, evaluates on WikiText-103, and reports performance for all the GPT-2 model sizes. This is our baseline to reach, training from scratch init.

Optionally help research other ways that people have evaluated GPT-2 models, or attempted to reproduce them in the past.

karpathy:
We are abandoning WikiText103 because it's a total mess. We'll instead look at one/few of ARC Easy / Challenge, Squad, Hellaswag, TriviaQA, LAMBADA. Closing.

PR #276, opened 2024-04-28
Adding WikiText-103 dataset preprocessing and tokenization

joeshmoe0112358:
This PR enables the preprocessing and tokenization of the WikiText-103 dataset in response to  #246.

A few notes I would like to make on my preprocessing of the text:
- I considered two ways of downloading the dataset
  1. The dataset can be found on hugging face at [https://huggingface.co/datasets/wikitext](url) where going into the 
```wikitext-103-raw-v1``` folder shows 4 different parquet files (2 for training, 1 for test, 1 for val). I initially attempted to use this method to download the dataset but I quickly ran into trouble realizing that unless I used pandas, an extra dependency, this would look very messy (please correct me if I am wrong about this I am kinda new to this stuff so maybe there is actually a way to handle parquet files smoothly without pandas). Because of this I opted for the other option.
  2. As [https://github.com/tysam-code/hlb-gpt/tree/main](url) was mentioned, you can use the same download link used in this code which gives you a zipped folder containing three files of raw text (wiki.train.raw, wiki.val.raw, wiki.test.raw). I am not entirely sure but I got the impression that the text in these files was some sort of translation from the original parquet files from the hugging face version because there were these weird headers/titles that were enclosed in like "=" on both sides. I thought that this download option would be the best choice because it would go smoother based on my knowledge.

- To deal with the awkward headers/titles, I made the effort of removing these and leaving just the paragraphs of text/descriptions and separating what would have been paragraphs for a given topic into their own sections with the use of <|endoftext|>. I believe my code takes care of most, if not all, of these headers so that the model can just focus on the text itself and not be possibly confused by the unusual headers that would have had to be encapsulated in <|endoftext|> tokens anyways leaving the model in a position of uncertainty of whether the text to follow after an <|endoftext|> is a header or an actual paragraph. But now that I am writing this I realize that this actually is not that big of a problem because of it decides to do a header then it will follow with the paragraph anyways so I am now thinking that making this effort is pointless. I will just leave it as is in case you don't care although I am pretty sure you would probably prefer to just do simpler preprocessing so that benchmarking is more standardized. In that case, please let me know and I can fix it.

- There are possibly better ways of doing this than the way that I did it. If there are, please let me know, and again I can fix it.

That pretty much sums it up. On a rough skim, ```train_gpt2.cu``` looked like it could already handle using this dataset right away just by using the i flag and typing ```data/wikitext-103``` but I may have missed something. I thought before diving too deep into other parts that I would at least just get this code for the dataset in.

### DIFF
--- a/prepro_wikitext-103.py
+++ b/prepro_wikitext-103.py
@@ -73,6 +73,7 @@ def tokenize():
     train_data_filename = os.path.join(DATA_CACHE_DIR, "wikitext-103/wikitext-103-raw/wiki.train.raw")
     train_text = open(train_data_filename, 'r').read()
 
+    print("Cleaning training data (this should take about 1 minute)...")
     # cleanup the training text
     train_text = train_text.strip() # remove leading and trailing whitespace
     train_text = train_text.replace(" \n \n ", '\n<|endoftext|>') # injecting special token in between sections
@@ -89,11 +90,13 @@ def tokenize():
     train_text = "<|endoftext|>".join(train_split[i] for i in range(len(train_split)))
     train_tokens = encode(train_text)
     train_tokens_np = np.array(train_tokens, dtype = np.int32)
+    print("Training data cleaned")
 
     # now repeat same cleanup process but for validation text
     val_data_filename = os.path.join(DATA_CACHE_DIR, "wikitext-103/wikitext-103-raw/wiki.valid.raw")
     val_text = open(val_data_filename, 'r').read()
 
+    print("Cleaning validation data...")
     val_text = val_text.strip() 
     val_text = val_text.replace(" \n \n ", '\n<|endoftext|>')
     val_text = "<|endoftext|>" + val_text
@@ -106,14 +109,18 @@ def tokenize():
     val_text = "<|endoftext|>".join(val_split[i] for i in range(len(val_split)))
     val_tokens = encode(val_text)
     val_tokens_np = np.array(val_tokens, dtype = np.int32)
+    print("Validation data cleaned")
 
     # now just dump the encoded tokens into binary files
     train_filename = os.path.join(DATA_CACHE_DIR, "wikitext-103_train.bin")
     val_filename = os.path.join(DATA_CACHE_DIR, "wikitext-103_val.bin")
     with open(train_filename, "wb") as f:
-       f.write(train_tokens_np.tobytes())
+        for chunk in tqdm([train_tokens_np[i : i + 1024] for i in range(0, len(train_tokens_np), 1024)], desc = "Writing train data to wikitext-103_train.bin", unit = "iB"):
+            f.write(chunk.tobytes())
+
     with open(val_filename, "wb") as f:
-        f.write(val_tokens_np.tobytes())
+        for chunk in tqdm([val_tokens_np[i : i + 1024] for i in range(0, len(val_tokens_np), 1024)], desc = "Writing validation data to wikitext-103_val.bin", unit = "iB"):
+            f.write(chunk.tobytes())
     
     print(f"Saved {len(val_tokens_np)} tokens to {val_filename}")
     print(f"Saved {len(train_tokens_np)} tokens to {train_filename}")
--- a/prepro_wikitext-103.py
+++ b/prepro_wikitext-103.py
@@ -1,0 +1,123 @@
+"""
+Downloads and tokenizes the WikiText-103 dataset.
+- The download is from "https://wikitext.smerity.com/wikitext-103-raw-v1.zip"
+ following https://github.com/tysam-code/hlb-gpt/tree/main
+- The tokenization is GPT-2 tokenizer with tiktoken
+
+The output is written to a newly created data/ folder.
+The script prints:
+
+Saved 241185 tokens to data/wikitext-103_val.bin
+Saved 114933466 tokens to data/wikitext-103_train.bin
+
+And runs in 3-4 minutes (~1.5 minutes to download data 
++ ~2 minutes to preprocess) depending on your internet
+connection and computer. The .bin files are raw byte
+streams of int32 numbers indicating the token ids.
+"""
+
+import os
+import re
+import requests
+from tqdm import tqdm
+
+import tiktoken
+import numpy as np
+
+DATA_CACHE_DIR = "data"
+enc = tiktoken.get_encoding("gpt2")
+encode = lambda s: enc.encode(s, allowed_special={'<|endoftext|>'})
+
+def download_file(url: str, fname: str, chunk_size=1024):
+    """Helper function to download a file from a given url"""
+    resp = requests.get(url, stream=True)
+    total = int(resp.headers.get("content-length", 0))
+    with open(fname, "wb") as file, tqdm(
+        desc=fname,
+        total=total,
+        unit="iB",
+        unit_scale=True,
+        unit_divisor=1024,
+    ) as bar:
+        for data in resp.iter_content(chunk_size=chunk_size):
+            size = file.write(data)
+            bar.update(size)
+
+def download():
+    """Downloads the WikiText-103 dataset to DATA_CACHE_DIR"""
+    os.makedirs(DATA_CACHE_DIR, exist_ok=True)
+
+    # download the WikiText-103 dataset, unless it's already downloaded
+    data_url = "https://wikitext.smerity.com/wikitext-103-raw-v1.zip"
+    data_filename = os.path.join(DATA_CACHE_DIR, "WikiText-103.zip")
+    if not os.path.exists(data_filename):
+        print(f"Downloading {data_url} to {data_filename}...")
+        download_file(data_url, data_filename)
+    else:
+        print(f"{data_filename} already exists, skipping download...")
+
+    # unzip the file
+    data_dir = os.path.join(DATA_CACHE_DIR, "wikitext-103")
+    if not os.path.exists(data_dir):
+        os.makedirs(data_dir, exist_ok=True)
+        print(f"Unzipping {data_filename}...")
+        os.system(f"unzip {data_filename} -d {data_dir}")
+    else:
+        print(f"{data_dir} already exists, skipping unzipping...")
+
+def tokenize():
+    # special token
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by azret** [comment]
This is cool.

This fails for me. Is there a cross platform lib that could do the unzip instead of calling an external process?

```
 # unzip the file
    data_dir = os.path.join(DATA_CACHE_DIR, "wikitext-103")
    if not os.path.exists(data_dir):
        os.makedirs(data_dir, exist_ok=True)
        print(f"Unzipping {data_filename}...")
        os.system(f"unzip {data_filename} -d {data_dir}")
    else:
        print(f"{data_dir} already exists, skipping unzipping...")
```

**Comment by joeshmoe0112358** [comment]
I am not actually sure why it's not working for you. I just pulled all changes so I am on the latest version of the repo and then I just ran ```python prepro_wikitext-103.py``` and the script ran for me producing this output:
```
Downloading https://wikitext.smerity.com/wikitext-103-raw-v1.zip to data/WikiText-103.zip...
data/WikiText-103.zip: 100%|████████████████████████████████████████████████████████| 183M/183M [00:09<00:00, 20.5MiB/s]
Unzipping data/WikiText-103.zip...
Archive:  data/WikiText-103.zip
   creating: data/wikitext-103/wikitext-103-raw/
  inflating: data/wikitext-103/wikitext-103-raw/wiki.test.raw
  inflating: data/wikitext-103/wikitext-103-raw/wiki.valid.raw
  inflating: data/wikitext-103/wikitext-103-raw/wiki.train.raw
Saved 241185 tokens to data/wikitext-103_val.bin
Saved 114933466 tokens to data/wikitext-103_train.bin
```

So it definitely worked for me. Can you tell me more about where it is failing for you. I see that you made mention of an error message regarding failing to decode. Are there any other error messages or parts that fail or is it only this one? By the way I like your idea about adding a progress bar and am working on implementing this right now. Thanks for the feedback.

File: prepro_wikitext-103.py [hunk 1]
```diff
@@ -0,0 +1,130 @@
...
+        download_file(data_url, data_filename)
+    else:
+        print(f"{data_filename} already exists, skipping download...")
+
+    # unzip the file
+    data_dir = os.path.join(DATA_CACHE_DIR, "wikitext-103")
+    if not os.path.exists(data_dir):
+        os.makedirs(data_dir, exist_ok=True)
+        print(f"Unzipping {data_filename}...")
+        os.system(f"unzip {data_filename} -d {data_dir}")
```
azret, line 64: This will fail if the "unzip" is not found. e.g. not available on Windows. Is there no python library that will do this?

prepro_wikitext-103.py
```diff
@@ -0,0 +1,130 @@
...
+    else:
+        print(f"{data_dir} already exists, skipping unzipping...")
+
+def tokenize():
+    # special token
+    eot = enc._special_tokens['<|endoftext|>']
+
+    # fetch training text
+    train_data_filename = os.path.join(DATA_CACHE_DIR, "wikitext-103/wikitext-103-raw/wiki.train.raw")
+    train_text = open(train_data_filename, 'r').read()
```
azret, line 74: here maybe add encoding="utf8" in the open() call. Fails on Windows without it.

prepro_wikitext-103.py
```diff
@@ -0,0 +1,130 @@
...
+    
+    # now join the remaining chunks via eot
+    train_text = "<|endoftext|>".join(train_split[i] for i in range(len(train_split)))
+    train_tokens = encode(train_text)
+    train_tokens_np = np.array(train_tokens, dtype = np.int32)
+    print("Training data cleaned")
+
+    # now repeat same cleanup process but for validation text
+    val_data_filename = os.path.join(DATA_CACHE_DIR, "wikitext-103/wikitext-103-raw/wiki.valid.raw")
+    val_text = open(val_data_filename, 'r').read()
```
azret, line 97: Same here encoding="utf8"

prepro_wikitext-103.py
joeshmoe0112358: Oh I understand what you mean now. I found that there is a python library called zipfile that should be able to do this. I am implementing this now.

joeshmoe0112358:
I have just updated it and added the changes that you have suggested. Please let me know if it now works for you or if you are still finding problems with it. Thanks again for your feedback.

azret:
Thank you for the update. It now works on Windows as is.

karpathy:
Useful reference that I found with a quick search:
- https://www.reddit.com/r/MachineLearning/comments/oye64h/r_struggling_to_reproduce_perplexity_benchmarks/h7ucco2/
- https://huggingface.co/docs/transformers/perplexity
- https://github.com/huggingface/transformers/issues/483
- https://github.com/openai/gpt-2/issues/78

Ok so odeally we would reproduce the numbers in Table 3 in the paper, but these numbers use some weird detokenizers. So we have two options:

1. Try to find which detokenizer was used and exactly reproduce Table 3
2. Ignore the detokenizer and reproduce the perplexity numbers posted by Alec in the Reddit thread on raw WikiText103, with its weirdness and all.

I'm ok with either of these, maybe slight preference to (1) but ok with (2). As part of this PR and before merging, we should try to reproduce either (1) and (2) as close as possible, to give confidence that we did things right.

joeshmoe0112358:
I am looking into this now and I will update you or ask questions as needed. Thanks for the guidance.

joeshmoe0112358:
Okay after reading on this carefully here are my takeaways:
- I am seeing a lot of different numbers on perplexity being reported across the board and many people having difficulty replicating/matching the numbers reported in Table 3 of the GPT-2 paper. Because of this, I think it would be best to first establish a baseline on the raw WikiText-103 text using the table by Alec as a stable anchor point. Once this is achieved, then we can consider further work on figuring out how to preprocess and detokenize the text to reproduce the numbers reported in Table 3.
- Since a large focus of this repo is simplicity, attempting to preprocess and reproduce the GPT-2 paper's Table 3 might increase complexity and make things messier. So maybe we should just ignore this and only do (2)? That is unless it is not an issue that the data preprocessing scripts get tricky because really the main priority is in keeping the actual C and CUDA code as clean as possible.
- When calculating the perplexity scores we should use a stride length of 32 for evaluation with a window size of 1024 because this is what was used to evaluate GPT-2 and also to produce Alec's table, according to Alec. 
- Although I have not looked at the main code for training and evaluation, I do not recall anything about computing perplexity scores so this may be more lines of code to set up.

In summary, what I plan to do:
- no preprocessing (including no <|endoftext|> injection) and raw WikiText-103 as a baseline
- reproduce Alec's number for WikiText-103 GPT-2 perplexity score using a 1024 sliding window and 32 stride length
- mess around with hugging face data loader stuff because they were able to replicate the number for gpt2-large on wikitext-2

If anything I have said here is wrong or you disagree with, please correct me.

joeshmoe0112358:
Okay there are some contradictions caused by Alec's table and information I think.

**Claims/Information:**
1. Alec claims to have produced a table of values for perplexity scores evaluated on the raw wikitext-103 and wikitext-2 without any detokenizers, using a stride length of 32 and a window size equal to the context length. The numbers in question are a perplexity score of 37.50 for wikitext-103 and 34.63 for wikitext-2.
2. The GPT-2 paper's table has a perplexity score of 37.50 on the preprocessed wikitext-103 and 29.41 on wikitext-2.
3. My tests using the hugging face script produce a perplexity score of 31.04 on the raw wikitext-103 and wikitext-2 val split, and 29.94 on wikitext-103 and wikitext-2 test split.
4. The huggingface perplexity scores tutorial reproduces the number for gpt2-large on raw wikitext-2 using a stride length of 1024, and using a stride length of 512 drops the perplexity score from 19.44 (the number reported in the paper is 19.93) to 16.45, showing how big of a difference the stride length can make and also verifying/reinforcing that GPT-2 was evaluated with stride length equal to context length (but also a bit skeptical here because huggingface used the raw text and got the same number?).

**Issues:**
- Under the assumption that the GPT-2 paper ran its evaluation on preprocessed text (also likely on the training split and I will explain why in just a bit), we expect that the GPT-2 paper scores are lower than those of Alec's raw text, which in this case seems to be true for wikitext-2 but not for wikitext-103 which makes me skeptical of Alec's number for wikitext-103 perplexity score.
- The test and validation splits of the wikitext-103 dataset are identical to the test and validation splits of the wikitext-2 dataset (they are literally the same text), this means that the perplexity scores reported for wikitext-2 and wikitext-103 should be identical. However, they are not. This could mean a few things:
  1. If they did actually evaluate on test or validation splits, they must have randomly sampled from the splits in order to produce this difference in scores (but even then it is highly unlikely that the gap should be this large unless they used a questionably small sample size that created this variation). 
  2. If they did not randomly sample chunks of text from val/test, they may have run their evaluations on the training splits of each dataset.
- The huggingface script got the same perplexity score (using stride length equal to context length) as the paper for gpt2-large but they evaluated on the raw wikitext-2 and the paper evaluated on preprocessed text.

**My take on this:**
- I am quite skeptical and actually a bit unsure of what to make of this because of the many contradictions here (e.g. varying stride lengths, same numbers but different methods, different numbers on same text)
- Reproducing these perplexity scores has proved to be a considerably challenging task that will require extensive testing.
- It may be best to use the numbers that we produce using the huggingface library as an anchor point instead for reproducing with the mainline implementation because we at least know for sure where our numbers will be coming from and we can make sure that we are testing llm.c with the same exact methods.

**What I am going to try next:**
- I am going to try doing some tests on the code I wrote that preprocesses wikitext-103 to see what perplexity scores result from evaluating on the val and test splits from this. If the numbers are similar to those from the paper, we can have some confidence that the preprocessing I did is close to that done for the paper's tests. However, I kinda doubt this because we still do not know what text the paper tested on other than that it was tested on "WikiText103," but I still think there is a reasonable chance that this may work.

I would like your judgement on this. Please let me know if I have made any mistakes or overlooked any details in my report on this issue.

karpathy:
We are abandoning WikiText103 because it's a total mess. We'll instead look at one/few of ARC Easy / Challenge, Squad, Hellaswag, TriviaQA, LAMBADA. Closing.

PR #278, opened 2024-04-28
moved checked helper functions into a separate file

ngc92:
function that are (or should be)  shared between our three training scripts moved into a common header file

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -49,6 +49,9 @@ mpirun -np 4 ./train_gpt2cu -b 8 -v 200 -s 200 -i data/TinyStories
 #include <nccl.h>
 #endif
 
+// our own utilities
+#include "utils.h"
+
 // ----------------------------------------------------------------------------
 // CUDA precision settings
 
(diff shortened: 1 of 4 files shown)


train_gpt2.cu
```diff
@@ -266,79 +269,6 @@ __device__ __forceinline__ void stochastic_rounding(float in, float *out, unsign
...
-FILE *fopen_check(const char *path, const char *mode, const char *file, int line) {
-    FILE *fp = fopen(path, mode);
-    if (fp == NULL) {
-        fprintf(stderr, "Error: Failed to open file '%s' at %s:%d\n", path, file, line);
-        fprintf(stderr, "Error details:\n");
-        fprintf(stderr, "  File: %s\n", file);
-        fprintf(stderr, "  Line: %d\n", line);
-        fprintf(stderr, "  Path: %s\n", path);
-        fprintf(stderr, "  Mode: %s\n", mode);
-        fprintf(stderr, "---> HINT: try to re-run `python train_gpt2.py`\n");
```
karpathy, line 282: :( this gets deleted. hmmm....

ngc92: should just be in the utils header, or am I missing something?

PR #281, opened 2024-04-29
Removed templates from kernels to clarify data types and simplify code

ChrisDryden:
When doing the conversion of 32 to 16 bit data types adding the templates helped clarify which inputs needed to share data types but is no longer needed and this makes it easier for developers now to tell what data types are being passed into the kernels.

Had this mostly written up before the chat in the discord about the decision of whether to keep mixed precision so thought I might as well make the PR


karpathy:
looks much cleaner ty

PR #282, opened 2024-04-29
Reused shared memory usage for sum and max

ChrisDryden:
When profiling this kernel I realized that these two shared memory is actually not used at the same time and can be reused.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -856,8 +856,7 @@ __device__ SoftmaxParams prepare_softmax_blockwide_nofloat4(cg::thread_block_til
     // two reductions of up to 1024 threads:
     // 1) inside warp (shuffle), 2) cross-warp (shared memory), 3) inside warp (shuffle)
     // this results in much cleaner assembly than a multi-warp cg::reduce
-    __shared__ float shared_maxval[32];
-    __shared__ float shared_sumval[32];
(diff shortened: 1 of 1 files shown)


ademeure:
There's actually a reason I did that which I should really have documented (so it's a good thing you pointed this out - this kind of thing shouldn't be implicit!) which leads to a subtle bug in your code: if you reuse the same shared memory, you need an additional __synchthreads() between these two lines:

```warp_maxval = (lane_id < num_warps) ? shared_maxval_sumval[lane_id] : -FLT_MAX;
[...]
if (lane_id == 0) { shared_maxval_sumval[warp_id] = warp_sumval; }```

Otherwise, you might have a warp in the block overwrite the data in shared_maxval_sumval before another warp has read the result for that previous instruction :( Since that is dynamic behaviour, you might get correct results on your GPU, but suffer from rare random errors that would be very difficult to track down, it's one the most annoying kinds of CUDA bugs in my opinion...

This kernel is so DRAM bandwidth limited that I don't think it will make any difference but it'd be somewhat interesting to see what the impact is on a more compute limited kernel doing a block reduction.

I think we probably want to close this PR but we'll want to revamp that code so it's nicely wrapped up and potentially no longer dependent on cooperative groups, so it might be worth revisiting this in the future though: https://github.com/karpathy/llm.c/issues/292

PR #290, opened 2024-04-29
Adding the example kernels to CI

Ricardicus:
Might as well build the example kernels up here in the CI!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,8 +49,8 @@ jobs:
       - name: Execute testing program (No OpenMP)
         run: ./test_gpt2
 
-  build-with-cuda-fp32:
-    runs-on: ubuntu-latest  # Host OS, Docker will run on top of this
+  build-cuda-fp32:
+    runs-on: ubuntu-latest
     container:
       image: nvidia/cuda:12.4.1-devel-ubuntu22.04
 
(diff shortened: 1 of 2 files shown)


rosslwheeler:
@Ricardicus - there was a discussion a few days ago and they wanted to keep those Makefile lines in there for visibility. I assume that might still be the case?

Ricardicus:
Oh ok! I has missed that. What kind of visibility? I think the names of the files are pretty good. This doesn't change what gets built and what not. I can leave the makefile as it is if it was a philosophy behind it..

Ricardicus:
If we want some kind of explanation of the translation units in there, maybe we should add a Readme under there instead of putting that doc in makefile

Ricardicus:
If we want some kind of explanation of the translation units in there, my suggestion is that we add a Readme here under instead of putting that doc in the makefile. The comments in the makefile I removed group the files in forward and backward pass, but the files already have _backward and _forward in them.

Ricardicus:
... and there is no wildcard *.cu being built, only the targets defined in the TARGETS variable (expanded in the all:) decides what gets built and what not..

karpathy:
I'm ok adding these to CI I think, but I wouldn't remove them from Makefile. There are great reasons to want to use these locally during dev. How long does this compile for / how much time does it add to CI

Ricardicus:
All of them are still in the makefile, they are still built. It can check the time it took.

rosslwheeler:
@Ricardicus - can you build them in CI with make -j?

Ricardicus:
Oh that would speed it up considerably ... can fiddle with it when i get back.. also; the reason i changed the makefile was because i considered some lines as unnecessary. This change does not change the makefiles functionality . The case for leaving these lines would be to make the makefile a bit more explanatory (with the comments)

Ricardicus:
I removed the changes to the Makefile and I added a -j4 that makes the build step faster.

PR #293, opened 2024-04-29
yet another gelu

ngc92:
more complicated Packet128 for cleaner kernels

### DIFF
--- a/dev/cuda/gelu_forward.cu
+++ b/dev/cuda/gelu_forward.cu
@@ -18,11 +18,8 @@ version 1 is naive port from CPU code to kernel
 #include <cuda_runtime.h>
 #include "common.h"
 
-// Finally, we define a wrapper type that contains 128 bits of whatever underlying type we want
-// we store the actual data in an int4 vector, and reinterpret its bits, because int4 gets nvcc to
-// reliably produce 128-bit instructions
-// TODO do we really need this here, or can we get away with the int4 trick just inside the  load/store functions
-// we allow individual element access with [], and provide a convenience accessor to get the data converted to
+// We define a wrapper type that contains 128 bits of whatever underlying type we want
+// We allow individual element access with [], and provide a convenience accessor to get the data converted to
 // a regular float for mixed-precision operations.
 
 template<class ElementType>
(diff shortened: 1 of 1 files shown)


ngc92:
This is how it would look like if we moved all the casting into the load/store functions:
```c++
template<class ElementType>
struct alignas(16) Packed128 {
    __device__ ElementType& operator[](int index) {
        return payload[index];
    }
    __device__ const ElementType& operator[](int index) const {
        return payload[index];
    }
    __device__ float fp32(int index) {
        return static_cast<float>(payload[index]);
    }
    static constexpr const size_t size = sizeof(int4) / sizeof(ElementType);

    ElementType payload[size];
};

// use this function to load a Packet128 from an aligned memory address
template<class ElementType, ELoadMode Mode=ELoadMode::CA>
__device__ __forceinline__ Packed128<std::remove_const_t<ElementType>> load_aligned(ElementType* address, load_mode_t<Mode> mode = {}) {
    int4 bits = generic_load(reinterpret_cast<const int4*>(address), mode);
    Packed128<std::remove_const_t<ElementType>> result;
    static_assert(sizeof(bits) == sizeof(result), "Size mismatch.");
    memcpy(&result, &bits, sizeof(bits));
    return result;
}

// use this function to store a Packet128 to an aligned memory address
template<class ElementType, EStoreMode Mode=EStoreMode::WB>
__device__ void store_aligned(ElementType* target, Packed128<ElementType> value, store_mode_t<Mode> mode = {}) {
    int4 bits;
    static_assert(sizeof(bits) == sizeof(value), "Size mismatch.");
    memcpy(&bits, &value, sizeof(bits));
    generic_store(reinterpret_cast<int4*>(target), bits, mode);
}
```

dev/cuda/gelu_forward.cu
```diff
@@ -44,6 +103,22 @@ __global__ void gelu_kernel(float* out, const float* inp, int N) {
...
+    int i = (blockIdx.x * blockDim.x + threadIdx.x) * packet_t::size;
+    if (i < N) {
+        packet_t packet_out;
+        packet_t packet_in = load_aligned_cs(inp + i);
+        for(int k = 0; k < packet_in.size; ++k) {
+            float xi = packet_in[k];
+            float cube = 0.044715f * xi * xi * xi;
+            packet_out[k] = 0.5f * xi * (1.0f + tanhf(GELU_SCALING_FACTOR * (xi + cube)));
+        }
+        store_aligned(out + i, packet_out);
```
karpathy, line 118: any reason we loadcs but store without cs?

ngc92: for this kernel, we read exactly once and store once, so these hints don't gain us anything. But by not keeping the input in cache, but the output there, _maybe_ the next kernel can be a bit faster. This is just guesswork though, I haven't actually measured this.

karpathy: got it! that makes sense actually

ngc92:
close by #298

PR #295, opened 2024-04-29
Remove FloatN & simplify adam/reduce with BF16 LayerNorms

ademeure:
The MULTI_GPU path is untested, but everything else seems to work fine. I kept the per-tensor "param_sizeof" as it's used in test_gpt2.cu for example, it's not much code and may be useful again in the future.

This will create merge conflicts with #289 but hopefully not hard to fix.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -59,6 +59,7 @@ mpirun -np 4 ./train_gpt2cu -b 8 -v 200 -s 200 -i data/TinyStories
 
 enum PrecisionMode {
     PRECISION_FP32,
+    PRECISION_FP16,
     PRECISION_BF16
 };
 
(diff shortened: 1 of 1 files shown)


train_gpt2.cu
```diff
@@ -1871,45 +1851,18 @@ void gpt2_update(GPT2 *model, float learning_rate, float beta1, float beta2, flo
         cudaCheck(cudaMalloc((void**)&model->v_memory, model->num_parameters * sizeof(float)));
         cudaCheck(cudaMemset(model->m_memory, 0, model->num_parameters * sizeof(float)));
         cudaCheck(cudaMemset(model->v_memory, 0, model->num_parameters * sizeof(float)));
-        printf0("allocated %d MiB for AdamW optimizer state m\n", (int)round(model->num_parameters * sizeof(float) / (1024 * 1024)));
-        printf0("allocated %d MiB for AdamW optimizer state v\n", (int)round(model->num_parameters * sizeof(float) / (1024 * 1024)));
+        printf("allocated %zu MiB for AdamW optimizer state m\n", (model->num_parameters * sizeof(float)) >> 20);
```
karpathy, line 1854: fixing this in my local pr before push, should be print0

PR #296, opened 2024-04-29
separate out our Tokenizer

karpathy:


tokenizer.h
```diff
@@ -0,0 +1,93 @@
+/*
+Defines the GPT-2 Tokenizer.
+Only supports decoding, i.e.: tokens (integers) -> strings
```
ngc92, line 3: Only supports **decoding**, i.e.: tokens (integers) -> strings
This is all we need for unconditional generation.
If we wanted to later prompt the model, we'd have to add **decoding**  :)

karpathy: hehe oops

karpathy:
I did better than just move it, I am now saving it in the Tokenizer file in the header, which is a lot better.

train_gpt2.py
```diff
@@ -351,8 +351,9 @@ def write_tokenizer(enc, filename):
     n = enc.max_token_value + 1
     header = torch.zeros(256, dtype=torch.int32)
     header[0] = 20240328 # magic
-    header[1] = 1 # tokenizer version = 1
+    header[1] = 2 # tokenizer version = 2 (1 -> 2: includes EOT token)
     header[2] = n # number of tokens
+    header[3] = enc.eot_token # EOT token
```
azret, line 356: Yep. Perfect! Thank you.

azret:
What about the C/CPU version?

Issue #292, opened 2024-04-29
delete use of cooperative groups in kernels


karpathy:
We use a lot of cooperative groups functionality in our kernels. This is an additional dependency that is likely mildly convenient, but it is also likely that the code could be written without them, without too much added complexity, and just as fast. As a general feature ideally llm.c is very careful in the "dependency surface" of its code, which would make it very portable, easy to skim/read even if slightly longer, and easy to run or port to any hardware, old/new/edge/exotic/ or otherwise unthought of.

I would accept PRs that develop cooperative-groups-free kernels in `dev/cuda` that:

1) aren't too much more complex or more LOC
2) have the same speed

On top of `dev/cuda` I'd be happy to merge these into "mainline" `train_gpt2.cu` and the fp32 version `train_gpt2fp32.cu`.

ChrisDryden:
Just posting some notes here on my research of how to remove all of the CG related code to remove the dependency:
```
    sum = cg::reduce(warp, sum, cg::plus<float>{});
```
Can be replaced with the following

```
__device__ float warpReduceSum(float val) {
    for (int offset = 16; offset > 0; offset /= 2) {
        val += __shfl_xor_sync(0xFFFFFFFF, val, offset);
    }
    return val;
}

sum = warpReduceSum(sum)
```

Without the need for any thread syncs. 


Also the other variables that are used can be replaced by the following:
```
int warpSize = 32;
int laneId = threadIdx.x % warpSize;
int warpId = threadIdx.x / warpSize;
int warpsPerBlock = (blockDim.x / warpSize);
```

```
warp.thread_rank() == laneId
warp.size() == warpSize
warp.meta_group_size() == warpsPerBlock
warp.meta_group_rank() == warpId
```

I have replaced most of the kernel to test for performance improvement and I was not able to see any noticable change by removing the cooperative groups.

ngc92:
in many  cases, I also find it quite convenient to just have a blockSize of 32 in x direction, and the rest in y direction.
Then threadIdx.x corresponds to laneId and threadIdx.y is warpId. Doesn't work when the block naturally already uses the other block dims.

PR #297, opened 2024-04-29
Remove the cooperative groups in matmul_backward_bias.cu

lancerts:
https://github.com/karpathy/llm.c/issues/292
Given the fastest one is kernel 4 which is also used in train_gpt2
1. Remove the old kernels 2 and 3, which are using cg
2. Re-index the kernel4 to kernel2

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -634,7 +634,7 @@ __global__ void gelu_backward_kernel(floatX* dinp, const floatX* inp, const floa
 // the idea is to employ one block to reduce along several columns,
 // where each block has a width of 32 columns to ensure coalesced access.
 // at the end we accumulate the reductions performed by the warps in each block via shared memory
-__global__ void matmul_backward_bias_kernel4(floatX* dbias, const floatX* dout, int B, int T, int OC) {
+__global__ void matmul_backward_bias_kernel2(floatX* dbias, const floatX* dout, int B, int T, int OC) {
     // this kernel is launched with 1D grid_dim of OC/32
     // for example let's say block_size is 128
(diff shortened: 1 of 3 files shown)


ngc92:
isn't this strictly losing information? The two different kernels in /dev/cuda compare whether it is more efficient to have shared_memory free warp-wide reductions, or a block-wide implementation, no?

train_gpt2.cu
```diff
@@ -634,7 +634,7 @@ __global__ void gelu_backward_kernel(floatX* dinp, const floatX* inp, const floa
 // the idea is to employ one block to reduce along several columns,
 // where each block has a width of 32 columns to ensure coalesced access.
 // at the end we accumulate the reductions performed by the warps in each block via shared memory
-__global__ void matmul_backward_bias_kernel4(floatX* dbias, const floatX* dout, int B, int T, int OC) {
+__global__ void matmul_backward_bias_kernel2(floatX* dbias, const floatX* dout, int B, int T, int OC) {
```
ngc92, line 637: should there be a number here?
not just `matmul_backward_bias_kernel`?

lancerts: the index of 2 is to be consistent of the index in the dev file (where the naive GPU kernel is indexed as 1).

lancerts:
> isn't this strictly losing information? The two different kernels in /dev/cuda compare whether it is more efficient to have shared_memory free warp-wide reductions, or a block-wide implementation, no?

Yeap, it is indeed removing the learnings. 
So maybe

ngc92:
or rewrite both versions  without cgs, if you're up for it?

PR #304, opened 2024-04-30
Stochastic rounding for all non-matmul activation gradients

ademeure:
I'd suggest trying this before the gazillion of inevitable f128-based performance optimisations in the next few days that will create a huge number of merge conflicts :) I wouldn't bother integrating this right now given that, but it'd be great to know if it helps the loss or not when pre-training from scratch.

---

Stochastic rounding is now implemented for the activation gradients in the backward pass for:
1. Adam (as before)
2. GELU
3. Layernorm
4. Softmax for attention
5. Fused classifier (activation gradients for logits only)
6. EDIT/NEW: The final encoder_backward kernel using atomicCAS and a retry loop (performance is better than what we currently have on main branch).

It is *NOT* implemented for:
- Any of the matmuls (impossible until we move away from cuBLAS unfortunately).
- Any of the weight/parameter gradients (useful with grandient accumulation?) and forward pass activations, etc...

The performance cost seems to be fairly negligible so we could integrate it elsewhere if it helps.

### DIFF
--- a/dev/cuda/encoder_backward.cu
+++ b/dev/cuda/encoder_backward.cu
@@ -84,6 +84,69 @@ __global__ void encoder_backward_kernel2(float* dwte, float* dwpe,
     }
 }
 
+// naive implementation with atomics
+__global__ void encoder_backward_kernel3(__nv_bfloat16* dwte, __nv_bfloat16* dwpe,
+                                        const __nv_bfloat16* dout, const int* inp,
+                                        int B, int T, int C) {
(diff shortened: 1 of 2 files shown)


ademeure:
Added encoder backward kernel with stochastic rounding (now it's "just" the matmuls missing stochastic for activation gradients).

Let me know if you need me to merge this with changes on main branch (but hopefully that can be avoided until we know whether it helps or not).

karpathy:
I tested this PR and did not see any difference, the results were in the noise. Closing this for now, may revisit in some form later.

PR #305, opened 2024-04-30
Make the fp32 function in pack128 const

lancerts:
The function should be marked as const. 
Otherwise, the .fp32 function is limited in its proper usage.
For example
```
float get_zero_index(const f128& x) {
       return x.fp32(0);  -------------> not allowed
}
```


karpathy:
deleted .fp32 closing

PR #307, opened 2024-04-30
Improve tanh derivative in backward gelu

akbariyeh:
It is cheaper to compute the derivative of tanh as 1 - tanh^2 than computing 1/(cosh^2). This will probably not make a measurable difference.


dagelf:
It makes a measurable difference, but it breaks the activation function for me. How are you compiling and running it? (Compiler/platform)

Issue #229, opened 2024-04-23
Windows Github actions / workflow is successfully building including Cuda 12.4 builds


rosslwheeler:
No issue - just a heads up. The Windows builds with GitHub actions is working. Adding the tests is next and then I'll be merging those changes into the ci.yml file (@Ricardicus). The only downside is that the current Cuda installation for Windows is slow but at least it's the latest (12.4) version. Be good if we can get containers with current 12.4 Cuda toolkits for all 3 OS's - that would speed things up. Next up is updating the Makefile to include the Windows build changes. These Makefile changes will be much smaller than in prior PRs. This will all be documented too...

Ricardicus:
Perhaps add a separate job for this, there are two jobs now:
- build-and-test-cpu, runs ubuntu that builds with and without OMP (build and test)
- build-with-cuda, runs in an nvidia container that has cuda-tools installed (just building)
Since Windows is such a world of its own I suggest it runs as an additional job, but that is perhaps already
what you had in mind. In my experience, MVCC is a lot stricter when compiling C. "build-and-test-cpu-mvcc" ?

Ricardicus:
Nice. Perhaps add a separate job for this, there are two jobs now:
- build-and-test-cpu, runs ubuntu that builds with and without OMP (build and test)
- build-with-cuda, runs in an nvidia container that has cuda-tools installed (just building)
Since Windows is such a world of its own I suggest it runs as an additional job, but that is perhaps already
what you had in mind. "build-and-test-cpu-mvcc" ?

rosslwheeler:
Yep, that's the plan. Be nice if we could find some self-hosted Cuda runners that someone could let us use???

Ricardicus:
Yes, it would be very nice to have a container with a virtual driver installed to run here on github. I can look into that.

Ricardicus:
Yeah so the virtual driver requires a purchase so I don't think we can use that for this. Otherwise there are other solutions for containers but they all rely on installations on the host which I don't think we can do here on Github. We can build and to static analysis, but the benchmarking etc. needs to be carried out elsewhere like it is today.

Ricardicus:
Yeah so the virtual driver requires a purchase so I don't think we can use that for this. Otherwise there are other solutions for containers but they all rely on installations on the host which I don't think we can do here on Github.

rosslwheeler:
[...]

PR #320, opened 2024-05-01
CI Windows changes

rosslwheeler:
@Ricardicus This is the first phase - just want to check on the structure. I'll put some more comments over here https://github.com/karpathy/llm.c/issues/229.


Ricardicus:
Nice! I was wondering when this would come since you added Windows support 😊 I have been using nmake from VS on windows. Nice to see there is another way. Is it possible to build the cuda also? Maybe it takes too long to install for the CI?

rosslwheeler:
Yes, the CI is taking too long when I install Cuda for WIndows dynamically but am trimming down the packages installed to see if I can get the time down.

Ricardicus:
The time it took is 5m 26s and Ubuntu variant is slower at 5m 33s. They run in parallel so this does not add more time to the CI I think. I think this looks solid!

Ricardicus:
Since the cuda build requires installing cuda which takes a lot of time, I think this is a start.. I can imagine that one could have a different trigger for that workflow action, like manually or periodically. I don't know if one can do it periodically here, haven't tested that. Would be neat. Perhaps just start here in this PR and add cuda windows build in another pr.

rosslwheeler:
Sound good! Makes sense to me. I'll do the follow-on PR once we get the Cuda timings down and/or have a solution periodically that makes sense.

Ricardicus:
Does it build with open MP support here in the windows CI?

rosslwheeler:
Yes, it's defaulting to OpenMP on. I didn't split out all the different versions (e.g. OpenMP off etc.) but can if we want to.

Ricardicus:
Ok! No I don't think it is needed. Just wondering. LGTM 👍

rosslwheeler:
Okay, this is ready to go. Reduced the Windows Cuda builds (Cuda installation mainly) time from 20+ minutes to 2 minutes. 

image

Ricardicus:
Very cool! Great work. I see that you download and install the cuda dependencies directly. I wonder why the official installation is so much slower.

rosslwheeler:
Yes, not sure why the executable installer is so much slower. 

Still looking into getting us some Cuda github runners too. Looks like we might need a team/enterprise GitHub account to get one (and that's not a guarantee either).

rosslwheeler:
Updated with this one https://github.com/karpathy/llm.c/pull/401 - it's ready to go.

PR #324, opened 2024-05-01
Refactoring parameter size filling

eymay:
Similar to the `.cu` version, in `.c` a new function is defined for parameter size table.
This will be useful for reuse with e.g. initializing weights from
scratch instead of a checkpoint for the issue #243.


karpathy:
good idea ty

PR #326, opened 2024-05-01
option to keep weights as fp32

ngc92:
adds an optional second copy of the weights in fp32 precision
~TODO missing free~


ngc92:
master weights are properly freed now

PR #328, opened 2024-05-01
feature/fp32 weight master copy

karpathy:
optionally keep a master copy of params in fp32
the added flag is -w 0/1, where 1 is default (i.e. by default we DO keep the fp32 copy)
increases memory for the additional copy of params in float, but the running time seems ~unaffected

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -1285,7 +1285,8 @@ __global__ void fused_classifier_kernel3(floatX* logits, floatX* losses, floatX*
     }
 }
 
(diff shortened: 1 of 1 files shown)


train_gpt2.cu
```diff
@@ -2429,6 +2454,7 @@ int main(int argc, char *argv[]) {
     int overfit_single_batch = 0; // useful for debugging, 1 = only load a single data batch once
     int max_steps = -1;
     int override_enable_tf32 = 1;
+    int use_master_weights = 1;
```
ngc92, line 2457: +1 for this naming change, the name in my original PR was just confusing :)

PR #329, opened 2024-05-02
Added example of removing cooperative groups in layernorm backwards

ChrisDryden:
Following the instructions in https://github.com/karpathy/llm.c/issues/292 I removed the cooperative groups from this kernel using warp sums. Did not have a detectable impact on performance for me.

### DIFF
--- a/dev/cuda/layernorm_backward.cu
+++ b/dev/cuda/layernorm_backward.cu
@@ -129,6 +129,13 @@ void layernorm_backward_cpu(float* dinp, float* dweight, float* dbias,
 // GPU kernels
 
 // GPU helper functions for atomicAdd on smaller than 32-bit types
+__device__ floatX warpReduceSum(floatX val) {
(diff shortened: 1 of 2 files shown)


dev/cuda/layernorm_backward.cu
```diff
@@ -129,6 +129,13 @@ void layernorm_backward_cpu(float* dinp, float* dweight, float* dbias,
 // GPU kernels
 
 // GPU helper functions for atomicAdd on smaller than 32-bit types
+__device__ floatX warpReduceSum(floatX val) {
```
ngc92, line 132: maybe we should move this to common.h; I guess we're going to use it for most of the kernels that currently have cgs

ngc92: maybe even a template version, then we have full feature parity with cg::reduce

karpathy: I agree we should move this to common.h

PR #339, opened 2024-05-03
Set GPU compute capability dynamically in Cuda makefile

rosslwheeler:
This is a draft proposal in case there is interest for kernel builds.

Usage:

Auto detect GPU capability:
>make
(e.g. if your GPU capability type is 80 then --generate-code=arch=compute_80,code=[compute_80,sm_80] is used with CFLAGS)

Do not specify capability:
>make GPU_COMPUTE_CAPABILITY=
(CFLAGS = -O3 --use_fast_math)

Override capability:
>make GPU_COMPUTE_CAPABILITY=86
(e.g. even if your GPU capability type is 80 then --generate-code=arch=compute_86,code=[compute_86,sm_86] is used with CFLAGS)

Tested on Linux Ubuntu 22.04 only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,11 +2,7 @@ name: Build and test
 
 on:
   push:
-    branches:
-      - master
-  pull_request:
-    branches:
-      - master
+  workflow_dispatch:
 
 jobs:
   build-and-test-cpu:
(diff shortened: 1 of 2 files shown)


rosslwheeler:
@Ricardicus - looks like the docker image doesn't have the extra executables. In my Ubuntu Cuda 12.4 default installation, it had __nvcc_device_query in the same path as nvcc. Let's see if there's interest first before doing anything.

rosslwheeler:
Fix the CI issue and resubmit.

PR #343, opened 2024-05-03
Performance: matmul_bias, cuda streams, fused_classifier (+remove cooperative groups)

ademeure:
I might need to split this into multiple PRs, let me know what you think (and I still need to add the new kernels to /dev/cuda/).

Major changes:

1. New super optimised matmul_backward_bias_kernel6
2. CUDA streams to parallelise a lot more things - ironically, the one big thing missing to prevent all remaining CPU-GPU synchromnisation is the "cudaEventSynchronize(end)" used for getting the performance...
3. Fix for cache hint in fused_classifier (significant perf gain, ~7 to ~5ms on my GPU)
4. Remove cooperative groups from remaining kernels + new blockReduce() helper

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -1626,20 +1626,11 @@ void matmul_backward(floatX* dinp, floatX* dweight, floatX* dbias,
 
         assert((OC % OC_per_warp) == 0); // there is no bounds checking in the kernel to maximise performance
 
-        // Run the memset on a parallel stream then immediately wait on it (runs in parallel with previous kernel)
-        cudaMemsetAsync(dbias_buffer, 0, OC * sizeof(float), parallel_streams[0]);
-        cudaEventRecord(parallel_events[0], parallel_streams[0]);
-        cudaStreamWaitEvent(main_stream, parallel_events[0], 0);
-
+        cudaMemsetAsync(dbias_buffer, 0, OC * sizeof(float), main_stream);
         matmul_backward_bias_kernel6<<<dim3(grid_size_x, grid_size_y),
                                        dim3(block_size_x, block_size_y),
                                        OC_per_warp * sizeof(float), main_stream>>>(dbias_buffer, dout, B, T, OC);
-        cudaEventRecord(main_event, main_stream);
-        cudaStreamWaitEvent(parallel_streams[0], main_event, 0); // cast_and_add_kernel is dependent on this
-
-        // run cast_and_add kernel on separate stream so it can run in parallel with the matmuls
-        cast_and_add_kernel<<<CEIL_DIV(OC, 256), 256, 0, parallel_streams[0]>>>(dbias, dbias_buffer, OC);
-        cudaEventRecord(parallel_events[0], parallel_streams[0]);
+        cast_and_add_kernel<<<CEIL_DIV(OC, 256), 256, 0, main_stream>>>(dbias, dbias_buffer, OC);
         cudaCheck(cudaGetLastError());
     }
 
@@ -1651,11 +1642,6 @@ void matmul_backward(floatX* dinp, floatX* dweight, floatX* dbias,
     cublasCheck(cublasGemmEx(cublas_handle, CUBLAS_OP_N, CUBLAS_OP_T, C, OC, B*T, &one,
                              inp, CUBLAS_LOWP, C, dout, CUBLAS_LOWP, OC, &one,
                              dweight, CUBLAS_LOWP, C, CUBLAS_LOWP_COMPUTE, CUBLAS_GEMM_DEFAULT_TENSOR_OP));
-
-    if (dbias != NULL) {
-        // make sure cast_and_add_kernel is done (hopefully finished in parallel with the matmuls)
-        cudaStreamWaitEvent(main_stream, parallel_events[0], 0);
-    }
 }
 
 void layernorm_backward(floatX* dinp, floatX* dweight, floatX* dbias, float* scratch,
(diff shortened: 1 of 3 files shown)


train_gpt2.cu
```diff
@@ -256,6 +260,29 @@ __device__ floatX warpReduceSum(floatX val) {
 }
 #endif
 
+using reduction_func_t = float (*) (float);
```
ngc92, line 263: not really a fan of the function pointer. Can we make this a template with a (compile-time constant) function pointer argument?

train_gpt2.cu
```diff
@@ -1572,6 +1607,42 @@ void matmul_backward(floatX* dinp, floatX* dweight, floatX* dbias,
     NVTX_RANGE_FN();
     float one = 1.0f;
     float zero = 0.0f;
+
+    // backward to bias, if given, does a +=
+    if (dbias != NULL) {
+        // Each warp is responsible for 32 * "x128::size" = 256 OCs at BF16 (OC must be a multiple of 256!)
```
karpathy, line 1613: ty for spelling things out 🙏 , like the example

train_gpt2.cu
karpathy: NASA approves of removing this i think, can be done later

train_gpt2.cu
```diff
@@ -256,6 +260,29 @@ __device__ floatX warpReduceSum(floatX val) {
...
+    __shared__ float shared_val[32];
+    int lane_id = threadIdx.x % 32;
+    int warp_id = threadIdx.x / 32;
+    int num_warps = blockDim.x / 32;
+
+    float warp_val = warp_reduction(val);
+    if (lane_id == 0) { shared_val[warp_id] = warp_val; }
+    __syncthreads();
+    // same strategy, now reduce across warps
+    warp_val = (lane_id < num_warps) ? shared_val[lane_id] : 0.0f;
```
karpathy, line 276: this assumes that the "base case" of the reduction is 0.0f right? Which may not be the case for arbitrary types of reduction.

train_gpt2.cu
ngc92:
```c++
template<reduction_func_t warp_reduction>
__device__ float blockReduce(float val, bool final_sync=false)
```
?
Haven't tried this, but it's how the kernel launcher works in one of the dev/cuda files

train_gpt2.cu
ngc92: in light of this comment, we may want to switch from function pointers to functors anyway, like the `cg::plus`  object. Something like
```c++
struct plus {
     static float operator()(float a, float b) { return  a + b; }
     static float neutral() { return 0.f;} 
}
```

train_gpt2.cu
```diff
@@ -1572,6 +1607,42 @@ void matmul_backward(floatX* dinp, floatX* dweight, floatX* dbias,
...
+    // backward to bias, if given, does a +=
+    if (dbias != NULL) {
+        // Each warp is responsible for 32 * "x128::size" = 256 OCs at BF16 (OC must be a multiple of 256!)
+        // Block size is 512 threads (16 warps) and we reduce those 16 values into 1 at the end
+        // blockDim.x is 32 --> single warp being responsible for those 256 OCs
+        // blockDim.y is 16 --> 16 parallel independent warps processing the same OCs for different BTs
+        // gridDim.x is OC / 256 --> each block processes 256 OCs
+        // grimDim.y is max(1, (cuda_num_SMs * cuda_threads_per_SM) / (512 * gridDim.x)); --> fill up the entire GPU!
+        const int warp_size = 32;
+        const int block_size = 512;
```
ChrisDryden, line 1620: Was testing out on the A100's making this 1024 would make a substantive difference, made a CR but didnt realize you had one out already

ChrisDryden:
https://github.com/karpathy/llm.c/issues/292 linking these together

train_gpt2.cu
ngc92: 1024 might be a terrible choice on non-A100/H100 cards (i.e., all those with 1.5k instead of 2k threads). Not sure if we should hard-code this. Haven't benchmarked this praticular kernel, but a slowdown of 33% wouldn't be unexpected.

ademeure:
Updated the PR and should be ready for merging!

I removed most of the CUDA stream parallelism for memset/memcpy, it was by far the riskiest/messiest part of the PR, the bug yesterday turned out to be I was implicitly relying on the "cudaEventSynchronize(end)" for timing in train_gpt2.cu, but this wasn't present in test_gpt2.cu, and my "fix" to set the cuBLAS calls to the new main CUDA stream instead of default stream made it possible for the inputs of the next step to be uploaded to GPU before the previous inputs were used!

We will probably want to reimplement this much more explicitly: e.g. have an event "used_scratch_buffer" set after the last read, and wait on it for before the next write. That would make it a lot clearer and less prone to bugs than waiting on a per-stream event like "parallel_stream[0]" in my previous version...

I kept the cuda stream for "loss_event" to avoid the CPU/GPU sync between forward and backward (mean_loss is now calculated at the end of gpt2_backward) which still provides a small gain. Compared to the previous version of this PR, we lost about ~0.4% performance on my RTX 4090. In the future it should be fairly easy to do even better than my previous PR by getting rid of the CPU/GPU synchronisation at the end of the training step and further improving memset() parallelism, so possibly up to ~1% from where we are here if we achieved ~perfect CPU/GPU parallelism.

train_gpt2.cu
ademeure: Yep that worked, I changed it to that instead of a function pointer for now (agreed with your other comment we could consider changing it a functor in the future).

train_gpt2.cu
ademeure: Agreed, for now I just added an "out_of_bounds" parameter that defaults to 0.0f so it's at least able to handle that case if we need it to.

train_gpt2.cu
ademeure: @ChrisDryden , you're right 1024 was a big win for the previous version of this kernel, but my new kernel6 doesn't seem to benefit, have you tried it on your end?

@ngc92 we could query "max_warps_per_multiprocessor" to determine that, for kernels that benefit from the largest possible block size at maximum occupancy, 768 would be optimal on non-A100/H100 cards (except Turing which is limited to 32 rather than 48!) - where it gets complicated is that depending on L2 cache size, I think some kernels like fused_classifier might actually benefit from *lower* occupancy... maybe something for me to investigate when I eventually start working on my "automatic sweeping of L2 cache residency per-kernel" idea.

PR #346, opened 2024-05-03
first attempt at moving cudnn out of the main file for faster compiles

ngc92:
I think this breaks nonCudnn build, and possibly also windows. I have little knowledge about makefiles, though, so if someone knows how to do things nicely, that would be great :)


ross-wheeler:
Sure - will take a look.

rosslwheeler:
@ngc92 - I can't seem to find cudnn_frontend.h?

rosslwheeler:
I have some Makefile changes but will submit those in a separate PR.

PR #348, opened 2024-05-04
Updated Matmul block size for .6ms speedup

ChrisDryden:
Went through all of the blocks on the A100 to see if any were not tuned correctly for an A100 and found that this block size had a stunning 2% ~(.6ms) improvement when changed and was fairly consistently that much faster even with cudnn.


ChrisDryden:
Was trying to go into the theory of why this is the case and saw the comment:
```
    // and atomically add everything together. atomics within one block are conflict-free!
```

Which can explain the dramatic speedup

ademeure:
Good catch, this would have been a good change for the previous matmul_bias_kernel5, but with the new version that's been integrated, I don't think it makes a difference anymore so we should probably close this?

PR #351, opened 2024-05-04
llm.c poster added. Productive pull soon :)

dhakalnirajan:
I added a llm.c poster on the README file. It is unconventional and counterproductive pull but I felt it was missing something (I was looking at llama.cpp repo) and I created this poster.


karpathy:
I can do better :)

PR #353, opened 2024-05-04
Fused layernorm residual

ngc92:
currently based on top of #352 
I'm not using  kernel 6, because
a) the performance seems to be really sensitive to parameters
b) ~i don't understand the performance measurements/are not 100% sure if it is actually correct despite passing our tests~

Update: The bug was I had `blockDim.z` where there  should be `threadIdx.z`, but since the first block_size=32 call writes correct results everywhere, all the untouched things were considered correct by the test.

Now, I get kernel 6 as a bit slower than kernel 5, but with a smaller shared memory footprint, so it  might be interesting for smaller cards or larger C.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -884,6 +884,87 @@ __global__ void layernorm_forward_kernel3(floatX* __restrict__ out, floatX* __re
     }
 }
 
+__global__ void fused_residual_forward_kernel5(floatX* residual, floatX* normed, floatX* mean, floatX* rstd,
+                                               const floatX* inp1, const floatX* inp2,
+                                               const floatX* weight, const floatX* bias,
+                                               int N, int C) {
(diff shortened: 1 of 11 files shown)


train_gpt2.cu
```diff
@@ -884,6 +884,87 @@ __global__ void layernorm_forward_kernel3(floatX* __restrict__ out, floatX* __re
     }
 }
 
+__global__ void fused_residual_forward_kernel5(floatX* residual, floatX* normed, floatX* mean, floatX* rstd,
+                                               const floatX* inp1, const floatX* inp2,
+                                               const floatX* weight, const floatX* bias,
+                                               int N, int C) {
+    constexpr const int WarpSize = 32;
```
karpathy, line 891: isn't there already a CUDA variable `warpSize`?

PR #355, opened 2024-05-05
 Refactoring & Improvements to reduce LOC

ademeure:
Refactoring and removing unused functions to reduce the number of lines of code and make everything slightly more consistent (while still having space for the code to breathe).

Also updates encoder_backward with my version from more_stochastic branch in order to delete atomicAddX() from the codebase, while improving accuracy very slightly by using stochastic rounding (it's literally the only place in the entire code where we are not accumulating in FP32!)

And changes the Makefile so that it compiles for the user's specific GPU when on Linux with nvidia-smi (generating both PTX and SASS for the binary so cuobjdump works to check the assembly).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ CUDA_OUTPUT_FILE = -o $@
 # -t=0 is short for --threads, 0 = number of CPUs on the machine
 # include PTX for both SM52 (Maxwell) and SM80 (Ampere, our main optimisation target at the moment) + native SASS for current GPU
 # this increases compile time by ~5%, but we need >=SM70 PTX for some optimisations, and it allows "cuobjdump --dump-sass" to work
-NVCC_FLAGS = -O3 -t=0 --use_fast_math -gencode=arch=compute_52,code=compute_52 -gencode=arch=compute_80,code=compute_80 -arch=native
+NVCC_FLAGS = -O3 -t=0 --use_fast_math -gencode=arch=compute_52,code=compute_52 -gencode=arch=compute_80,code=sm_80
 NVCC_LDFLAGS = -lcublas -lcublasLt
 NVCC_INCLUDES =
 NVCC_LDLIBS =
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ USE_CUDNN ?= 0
 ifeq ($(SHELL_UNAME), Linux)
     NVCC_ARCH := $(shell which nvidia-smi > /dev/null 2>&1 && nvidia-smi --query-gpu=compute_cap --format=csv,noheader,nounits | head -n 1 | sed 's/\.//g')
     ifdef NVCC_ARCH
-        NVCC_FLAGS += -gencode arch=compute_$(NVCC_ARCH),code=compute_$(NVCC_ARCH),code=sm_$(NVCC_ARCH)
+        NVCC_FLAGS += -gencode arch=compute_$(NVCC_ARCH),code=compute_$(NVCC_ARCH) -gencode arch=compute_$(NVCC_ARCH),code=sm_$(NVCC_ARCH)
     endif
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,7 @@ CUDA_OUTPUT_FILE = -o $@
 
 # NVCC flags
 # -t=0 is short for --threads, 0 = number of CPUs on the machine
-# include PTX for both SM52 (Maxwell) and SM80 (Ampere, our main optimisation target at the moment) + native SASS for current GPU
-# this increases compile time by ~5%, but we need >=SM70 PTX for some optimisations, and it allows "cuobjdump --dump-sass" to work
-NVCC_FLAGS = -O3 -t=0 --use_fast_math -gencode=arch=compute_52,code=compute_52 -gencode=arch=compute_80,code=sm_80
+NVCC_FLAGS = -O3 -t=0 --use_fast_math
 NVCC_LDFLAGS = -lcublas -lcublasLt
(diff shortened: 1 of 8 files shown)


Makefile
rosslwheeler, line 1: @ademeure - this is a different example of the same change from a few days ago in this PR: https://github.com/karpathy/llm.c/pull/339/files - in case that's useful for you.

train_gpt2.cu
```diff
@@ -1276,50 +1186,32 @@ void attention_forward(floatX* out, floatX* qkvr, floatX* att,
...
-                                     CUBLAS_LOWP_COMPUTE,
-                                     CUBLAS_GEMM_DEFAULT));
+                                     &beta, preatt, CUBLAS_LOWP, T, T * T,
+                                     B * NH, cublas_compute, CUBLAS_GEMM_DEFAULT));
 
     // multiply all elements of preatt elementwise by scale
     float scale = 1.0 / sqrtf(HS);
-    int grid_size = CEIL_DIV(B * NH * T * 32, softmax_block_size);
-    softmax_forward_kernel5<<<grid_size, softmax_block_size, 0, main_stream>>>(att, scale, preatt, B * NH, T);
-    cudaCheck(cudaGetLastError());
```
karpathy, line 1307: cudaCheck got lost here i think

train_gpt2.cu
```diff
@@ -1383,29 +1275,28 @@ void matmul_backward(floatX* dinp, floatX* dweight, floatX* dbias,
                                        dim3(block_size_x, block_size_y),
                                        OC_per_warp * sizeof(float), main_stream>>>(dbias_buffer, dout, B, T, OC);
         cast_and_add_kernel<<<CEIL_DIV(OC, 256), 256, 0, main_stream>>>(dbias, dbias_buffer, OC);
-        cudaCheck(cudaGetLastError());
```
karpathy, line 1386: hmm here too

train_gpt2.cu
```diff
@@ -2410,43 +2335,14 @@ int main(int argc, char *argv[]) {
...
-    if(cublas_compute_type); // unused in BF16 mode, avoid warning
-
-    // set up cuDNN (noop if not available)
-    create_cudnn();
+    const char* precision_str = (PRECISION_MODE == PRECISION_FP32)
+                              ? (cublas_compute == CUBLAS_COMPUTE_32F_FAST_TF32 ? "TF32" : "FP32")
+                              : (PRECISION_MODE == PRECISION_FP16 ? "FP16" : "BF16");
 
     printf0("| device                | %-50s |\n", deviceProp.name);
-    printf0("| TF32                  | %-50s |\n", enable_tf32 ? "enabled" : "disabled");
```
karpathy, line 2448: enable_tf32 was lost in the print, maybe ok...

karpathy:
All looks good happy to merge.

- A few stray cudaCheck(cudaGetLastError());
- A dropped print of enable_tf32 (?)
- CI failed for fp16 is this expected?

train_gpt2.cu
ademeure: That's actually intentional (but debatable whether it's a good idea) - I kept one cudaCheck per "wrapper function", e.g. one at the end of attention_forward rather than 1 after each of its 3 kernels. Feels like it's more than enough for initial debugging, and there were actually some cudaCheck() missing for one or two kernel launches, so if we decide we do want one for every kernel, we should add them as well.

train_gpt2.cu
ademeure: See definition of "precision_str" just above that: it now says "TF32" instead of "FP32" when enable_tf32 is on which feels like it should be enough?

Makefile
ademeure: Thanks! Didn't see that, it feels slightly cleaner (my changes came from Claude Opus...) so I removed the Makefile changes so we can integrate your #339 PR

cudnn_att.cu
```diff
@@ -30,28 +34,17 @@ static void cudaCheck(cudaError_t error, const char *file, int line) {
...
-
-        NvtxRange(const std::string& base_str, int number) {
-            std::string range_string = base_str + " " + std::to_string(number);
-            nvtxRangePush(range_string.c_str());
-        }
-
-        ~NvtxRange() { nvtxRangePop(); }
-    };
-}
+class NvtxRange {
```
ngc92, line 37: this seems like an  ODR violation waiting  to happen?

ademeure: Oops, fixed - I copy-pasted that as part of a chunk from train_gpt2.cu, didn't notice the namespace you added, thanks!

ademeure:
>     * CI failed for fp16 is this expected?

Nope that was a mistake, I didn't include a FP16 version of the new atomicStochasticAdd - I fixed it to use templates, which also let me get rid of "__bfloat1622float2" which afaik was the only missing functionality on older versions of CUDA, so train_gpt2.cu should (hopefully) now compile on CUDA 10/11 as well!

Makefile
rosslwheeler: Will do - @ademeure - will post a draft PR first for your review.

train_gpt2.cu
```diff
@@ -33,26 +34,15 @@ This reads & runs in fp32, B=4, T=64, LR=1e-4, val/sample never (200),
...
-#include <unistd.h>
-
+#include <string>
 // GPU / CUDA related
-#include <cublas_v2.h>
-#include <cuda_profiler_api.h>
 #include <cuda_runtime.h>
+#include <cublas_v2.h>
 #include <cublasLt.h>
-#include <cuda_bf16.h>
```
ngc92, line 53: where did this include go?

PR #357, opened 2024-05-05
cuDNN makefile changes

rosslwheeler:
Small cleanup changes on the build for cudnn. Added Windows support for cudnn. Tested on Ubuntu 22.04 and Windows 11.

- Command line and Windows changes - (top part of the Makefile) Change from := to ?= so we can override the CUDNN_FRONTEND_PATH from the command line. Fixed the locations for the CUDNN_FRONTEND_PATH for Windows and Linux to search for both $(HOME) or current directory)

- cuDNN Windows support (in the middle of the Makefile)

- Makefile cleanup - (lower part of Makefile) change "$<" to "$^" to pull in ALL of the dependencies. This eliminates the need to put (NVCC_CUDNN) on the recipe line. Removed linker files from object file compilation recipe. Added $(NVCC_CUDNN) to list of targets. That way it gets deleted on a make clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,16 +67,16 @@ else
 endif
 
 # Check and include cudnn if available
-# Currently hard-coding a bunch of stuff here for Linux, todo make this better/nicer
+# You can override the path to cudnn frontend by setting CUDNN_FRONTEND_PATH=your_path on the make command line
 # You need cuDNN from: https://developer.nvidia.com/cudnn
-# Follow the apt-get instructions
+# Follow the apt-get instructions or Windows instructions to install the cuDNN library
 # And the cuDNN front-end from: https://github.com/NVIDIA/cudnn-frontend/tree/main
-# For this there is no installation, just download the repo to your home directory
+# For this there is no installation, just download the repo to your home directory or directory of your choice
 # and then we include it below (see currently hard-coded path assumed in home directory)
 ifeq ($(USE_CUDNN), 1)
   ifeq ($(SHELL_UNAME), Linux)
     # hard-coded path for now
-    CUDNN_FRONTEND_PATH := $(HOME)/cudnn-frontend/include
+    CUDNN_FRONTEND_PATH ?= $(HOME)/cudnn-frontend/include
     ifeq ($(shell [ -d $(CUDNN_FRONTEND_PATH) ] && echo "exists"), exists)
       $(info ✓ cuDNN found, will run with flash-attention)
       NVCC_INCLUDES += -I$(CUDNN_FRONTEND_PATH)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -67,16 +67,16 @@ else
...
+# Follow the apt-get instructions or Windows instructions to install the cuDNN library
 # And the cuDNN front-end from: https://github.com/NVIDIA/cudnn-frontend/tree/main
-# For this there is no installation, just download the repo to your home directory
+# For this there is no installation, just download the repo to your home directory or directory of your choice
 # and then we include it below (see currently hard-coded path assumed in home directory)
 ifeq ($(USE_CUDNN), 1)
   ifeq ($(SHELL_UNAME), Linux)
     # hard-coded path for now
-    CUDNN_FRONTEND_PATH := $(HOME)/cudnn-frontend/include
+    CUDNN_FRONTEND_PATH ?= $(HOME)/cudnn-frontend/include
```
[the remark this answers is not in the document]
rosslwheeler: @ngc92 - I used the conditional assignment ?= which allows for make command line override.

so you could do: make USE_CUDNN=1 CUDNN_FRONTEND_PATH=dev 

Hoping that's good enough short-term until we figure out a better location?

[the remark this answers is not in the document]
ademeure: Maybe ideally we'd have the Makefile check a few locations automatically, including both $HOME and the current directory? This PR seems like an improvement without that so it doesn't need to be done at the same time necessarily.

Makefile
```diff
@@ -87,7 +87,19 @@ ifeq ($(USE_CUDNN), 1)
       $(error ✗ cuDNN not found. See the Makefile for our currently hard-coded paths / install instructions)
     endif
   else
-    $(info → cuDNN is not supported right now outside of Linux)
+    ifneq ($(OS), Windows_NT)
+      $(info → cuDNN is not supported on MAC OS right now)
+    else
+      $(info ✓ Windows cuDNN found, will run with flash-attention)
+      CUDNN_FRONTEND_PATH ?= ..\..\cudnn-frontend\include #override on command line if different location
```
karpathy, line 94: seems inconsistent? it's HOME above but it's ../../ here

karpathy:
I'm not amazing at Makefiles maybe there is a bit of a reasoning around what the changes here are and why they make sense?

Makefile
rosslwheeler: Yes, it's inconsistent - there's not a pre-defined HOME on windows - they have HOMEDRIVE and HOMEPATH. I can use the HOMEPATH since that's more like HOME on Linux if that's preferable? I'll make this change.

rosslwheeler:
@karpathy - notes on the above changes:

Windows changes - fixing the locations for the CUDNN_FRONTEND_PATH to (HOMEPATH and current directory) for Windows.

Makefile
karpathy: ok will wait for this before merging

Makefile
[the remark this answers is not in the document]
rosslwheeler: This has been fixed.

Makefile
rosslwheeler: This has been fixed.

PR #363, opened 2024-05-05
Modified version of ademeure's fused gelu_forward kernel

ChrisDryden:
Was experimenting with the fused gelu kernel to combine it to have the previous code when working with non-gelu matmuls that was built previously and when running it locally it appeared to have a performance benefit and was showing up with the correct losses for b16

Was hoping to get second opinion whether my observations were correct about this change

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -786,22 +786,6 @@ __global__ void residual_forward_kernel(floatX* out, floatX* inp1, floatX* inp2,
 }
 
 #define GELU_SCALING_FACTOR sqrtf(2.0f / M_PI)
-__global__ void gelu_forward_kernel2(floatX* out, const floatX* inp, int N) {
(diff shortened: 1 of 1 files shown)


ademeure:
You're right that I removed the code to set bias pointer by mistake, oops! But I still see the same lower performance on my end with an inefficient non-fused epilogue kernel.

Can you try running "python profile_gpt2cu.py" and see if you can find a line like this?

`08 fwd×12  cublasLt::epilogue::impl::globalKernel       5.74    871.3      0.0     1.81     3.19     1.70     3.38   798.18`

If you don't see one, then that's really interesting, because it would mean that this is GPU/CUDA/driver version dependent... if you don't see that kernel on your end, it'd be great if you could copy-paste the outputs of "nvcc --version" and "nvidia-smi" (and possibly of profile_gpt2cu.py as well).

PR #376, opened 2024-05-06
Replace non-standard C syntax with standard C syntax - fix win compilation

rosslwheeler:
- put unistd.h back for windows compilation - needs this to find M_PI
- #undef FLT_MAX - used to fix collision with redefinition below ~line 686
- fixed () with preprocessor conditional and removed nonstandard use of "not" and "and".
- uint -> unsigned int

Microsoft compilation is stricter than GCC at times. Tested on Win 11 and Ubuntu 22.04.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -54,7 +54,6 @@ This reads & runs in fp32, B=4, T=64, LR=1e-4, val/sample never (200),
 #include "utils.h"
 // defines: tokenizer_init, tokenizer_decode, tokenizer_free
 #include "tokenizer.h"
-#undef FLT_MAX
 
 // ----------------------------------------------------------------------------
 // CUDA precision settings
@@ -683,8 +682,8 @@ __global__ void softmax_forward_kernel5(floatX* out, float inv_temperature, cons
     const floatX* x = inp + idx * T;
 
     // not INF, so we don't get NaNs accidentally when subtracting two values.
-    const float FLT_MAX = 340282346638528859811704183484516925440.0f; // to avoid including float.h
-    float maxval = -FLT_MAX;
+    const float flt_max = 340282346638528859811704183484516925440.0f; // to avoid including float.h
+    float maxval = -flt_max;
     float sumval = 0.0f;
 
     const floatX* x_aligned = reinterpret_cast<const floatX*>(__builtin_assume_aligned(x, 16));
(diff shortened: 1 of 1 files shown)


train_gpt2.cu
```diff
@@ -53,6 +54,7 @@ This reads & runs in fp32, B=4, T=64, LR=1e-4, val/sample never (200),
 #include "utils.h"
 // defines: tokenizer_init, tokenizer_decode, tokenizer_free
 #include "tokenizer.h"
+#undef FLT_MAX
```
karpathy, line 57: I don't get this one. There's no re-definition I can see

rosslwheeler: in math.h
C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\ucrt\float.h

#define FLT_MANT_DIG     24                      // # of bits in mantissa
**#define FLT_MAX          3.402823466e+38F        // max value**
#define FLT_MAX_10_EXP   38                      // max decimal exponent
#define FLT_MAX_EXP      128                     // max binary exponent
#define FLT_MIN          1.175494351e-38F        // min normalized positive value


Around line 686 in train_gpt2.cu - those definitions aren't the same so the compiler errors it out.

```
    // not INF, so we don't get NaNs accidentally when subtracting two values.
    const float FLT_MAX = 340282346638528859811704183484516925440.0f; // to avoid including float.h
    float maxval = -FLT_MAX;
    float sumval = 0.0f;

```

karpathy: I'd rather just change the (local) varname here, e.g. make it flt_max or something. undef is not amazing because it is a global state kind of statement

rosslwheeler: Got it - updating...

karpathy: I think the NoneType CI error is from a PR I just merged from @Chillee , which I assumed was bug-free, should have paid more attention :) I could revert it, ... let me see if i can figure out what the source is here

rosslwheeler: Not a problem. Will check the CI to see when it's clear. The fix is ready.

karpathy: @rosslwheeler I pushed a fix to master, try now

rosslwheeler: Okay - let's see how it does now.

Chillee: haha I was just running the benchmark script - I assumed that CI would catch any other issues 😅

PR #381, opened 2024-05-08
Fix the float value to int in cudaMemset

lancerts:
cudaMemset takes the int (not float) value


zeux:
note that `cudaMemset` fills the buffer with the specified *byte* (if you pass a floating point value it will be converted to `int` during the call, and then `cudaMemset` implementation will truncate that to a byte). If initialization with -10000.f is important, it will need a kernel.

lancerts:
Thanks for the detailed explanation!

Issue #369, opened 2024-05-06
MultiGPU training hangs


chinthysl:
mpirun with multiple GPUs is hanging after `allocated 474 MiB for master copy of params`
Most probably due to the introduction of cudastreams. @karpathy @PeterZhizhin

karpathy:
Bleh I can reproduce this 🤦‍♂️ . Would most likely be helpful to harden our CI to include multi-GPU training, as it falls outside of it. iirc @Ricardicus helped set it up, not sure if he has knowledge of.

Ricardicus:
Hardening the CI would be great. But the CI here on github workflows only reaches as far as to do the CPU test only unfortunately..  I don't see a way to include real GPU tests automatically up here.. It is different if we had our own hardware we could connect to and run jobs on, but that would have to be a different solution.

ademeure:
Hmm, I can't immediately think of a reason why CUDA streams would break this, NCCL is still on the default stream which should keep everything synchronised both ways: https://docs.nvidia.com/cuda/cuda-runtime-api/stream-sync-behavior.html

One way to roughly restore the previous behaviour would be to replace this:
```cuda
    cudaCheck(cudaStreamCreate(&main_stream));
    cudaEventCreateWithFlags(&main_event, cudaEventDisableTiming);
    cudaEventCreateWithFlags(&loss_event, cudaEventDisableTiming);
    for (int i = 0; i < num_parallel_streams; i++) {
        cudaCheck(cudaStreamCreate(&parallel_streams[i]));
        cudaEventCreateWithFlags(&parallel_events[i], cudaEventDisableTiming);
    }
```
with this:
```cuda
    main_stream = 0;
    cudaEventCreateWithFlags(&main_event, cudaEventDisableTiming);
    cudaEventCreateWithFlags(&loss_event, cudaEventDisableTiming);
    for (int i = 0; i < num_parallel_streams; i++) {
        parallel_streams[i] = 0;
        cudaEventCreateWithFlags(&parallel_events[i], cudaEventDisableTiming);
    }
```
I don't have a multi-GPU setup to give this a try right now though...

rosslwheeler:
Hangs for me too - trying @ademeure changes above now.

chinthysl:
@Ricardicus @ademeure still hangs. changing nccl stream to main_stream doesn’t help. Will further look into multi stream usage and memcopyasyncs.

rosslwheeler:
[...]

PR #382, opened 2024-05-08
Fix incorrect GPU assignment in multi gpu setup

chinthysl:
Fix for #369

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -428,6 +428,7 @@ MultiGpuConfig multi_gpu_config_init(int *argc, char ***argv) {
     return result;
 #else
     printf("Multi-GPU support is disabled. Using a single GPU.\n");
+    cudaCheck(cudaSetDevice(0));
     MultiGpuConfig result;
     result.process_rank = 0;
     result.num_processes = 1;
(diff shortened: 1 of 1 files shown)


rosslwheeler:
Check out this PR - works fine on WSL2 Ubuntu 22.04. Thanks for the fix!

pjj:
This fix works on my end for the small/default gpt2 model (2.1m tok/s on 8 GPU).

The current code still breaks for the gpt2-xl model, albeit in a different way, unrelated to openMPI (with 1 GPU or 8GPU).  The error I'm getting is: 
```1373: void matmul_backward(floatX*, floatX*, floatX*, floatX*, floatX*, floatX*, float*, int, int, int, int): Assertion `(OC % OC_per_warp) == 0' failed.``` 

Commit 6c179fa works fine for gpt2-xl.  It is possible that a separate independent bug was introduced since that commit that only breaks the larger model.

karpathy:
Confirm this fixes the issue, stepping at ~460K tok/s on 4XA100 GPUs.

pjj:
I've verified that the fix by ademeure works on my end for GPT2-XL and opened a PR.

PR #395, opened 2024-05-09
Update README.md

yijunyu:
This project aims to achieve the same performance as llm.c and making the resulting code maintainable, e.g., turning the While loops generated from c2rust into For loops, which enables the parallelization through the `rayon` library.


karpathy:
cool! :)

PR #397, opened 2024-05-10
Update the -lcublas -lcublasLt flag in the comment

lancerts:
Due to the `void setup_main()` in the common.h,
```
    // setup cuBLAS and cuBLASLt
    cublasCheck(cublasCreate(&cublas_handle));
    cublasCheck(cublasLtCreate(&cublaslt_handle));
    cudaCheck(cudaMalloc(&cublaslt_workspace, cublaslt_workspace_size));
```
added ` -lcublas -lcublasLt` in the comment for compile for consistency.

Or, we can simply remove the comment for compile since everything is in the makefile already.

### DIFF
--- a/dev/cuda/adamw.cu
+++ b/dev/cuda/adamw.cu
@@ -6,8 +6,8 @@ References:
   * https://github.com/nvidia/apex/blob/master/csrc/multi_tensor_adam.cu
 
 Compile example:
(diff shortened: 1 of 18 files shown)


karpathy:
oh yeah an unfortunate side effect of code sharing here, pulling in dependencies that are not needed and bloating the compile command.

PR #399, opened 2024-05-10
Removed duplicated num_parameters calculation in train_gpt2.c

zuyu:
`model.num_parameters` has already computed before calling `malloc_and_point_parameters`, thus no needs to recompute it again.


karpathy:
I did it this way specifically so that the number of inputs passed around is minimized and that the functions are self-contained. I think ok as is for now ty

PR #402, opened 2024-05-12
Fix nul typo in Makefile

krrishnarraj:


krrishnarraj:
Realised its for windows.
But it creates a 'nul' file on linux

krrishnarraj:
@rosslwheeler

PR #403, opened 2024-05-12
Add Llm.cs by nietras to README

nietras:
@karpathy first of all thank you very much for creating this project and all the other content you produce for free. 🙏 I have previously ported micro-grad to C# too and here I link to my port of `llm.c` to C#. I started this before any CUDA code so mine is CPU only currently and I had no idea of the other port. Focus has been on adopting to .NET standards and making it easy to simply clone and run. Feel free to try it out, it's very easy ;)

I am very much looking forward to llm.c including all aspects of GPT2 training weight initialization and tokenization etc.

Regarding tokenization you might find inspiration in .NET https://github.com/dotnet/machinelearning/tree/main/src/Microsoft.ML.Tokenizers for that, and .NET has this feature that it can source generate code for regex expression, which might be handy if wanting to avoid regex. Let me know if that has any interest.
```csharp
        [GeneratedRegex(P50kBaseRegexPattern)]
        internal static partial Regex P50kBaseRegex();
```
https://github.com/dotnet/machinelearning/blob/4bf22dc5ecd078cf26521e1b4040724cfb3e52b1/src/Microsoft.ML.Tokenizers/Model/Tiktoken.cs#L1158-L1159

The generated code is supposedly readably and hence potentially possible to port to C.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,7 @@ Lastly, I will be a lot more sensitive to complexity in the root folder of the p
 
 - C#
   - [llm.cs](https://github.com/azret/llm.cs) by @[azret](https://github.com/azret): a C# port of this project
+  - [Llm.cs](https://github.com/nietras/Llm.cs) by @[nietras](https://github.com/nietras): a C# port of this project with focus on easy to get started on any platform. Clone and run ✅
 
 - CUDA C++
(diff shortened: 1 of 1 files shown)


karpathy:
nice! :)

PR #405, opened 2024-05-13
Backward bias

ngc92:
* fixed up `dev/cuda/matmul_backward_bias.cu` to support bf16.
* added kernel7 to dev/cude
* new  kernel 8 that reduces the number of channels handled by each block by a  factor of 4.  => reduction of global atomics
* Within-block reductions are handled deterministically  using warp shuffles. 
* For large enough OC, we can actually switch off the intermediate buffer and write the result directly. For the A4000, the OC dimension in the hidden layers would be large enough to disable the intermediate buffer for the 120M model with 1024 block size.
* On the A4000, I observe about 4% speed-up for this kernel.

While this change doesn't yet provide full determinism for bias gradient, it already drastically reduces the number of atomics. In a follow-up, we probably should just use a buffer of size  `OC * blockDim.y`, and do one more deterministic reduction in the second kernel call.

### DIFF
--- a/dev/cuda/matmul_backward_bias.cu
+++ b/dev/cuda/matmul_backward_bias.cu
@@ -243,6 +243,73 @@ __global__ void matmul_backward_bias_kernel7(float* dbias, const floatX* dout, i
     }
 }
 
+// We want to decrease the amount of channels handled by each block, so that we need fewer across-block reductions.
+// We do this by realizing the following: For scalar memory access, we need to read one element per thread in a warp
+// to read an entire cacheline, but for vectorized memory access, with 128 bit of data per thread, we only need eight
+// threads to fetch a cacheline, which means that we can already operate on a "depth" of four within a single warp.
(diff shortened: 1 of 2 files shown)


dev/cuda/matmul_backward_bias.cu
```diff
@@ -288,31 +479,30 @@ int main(int argc, char **argv) {
     // matmul_backward_bias(kernel_num, NULL, NULL, d_dbias, d_dout, NULL, NULL, NULL, B, T, C, OC, block_size_debug);
     // exit(EXIT_SUCCESS);
 
-    int block_sizes[] = {32, 64, 128, 256, 512, 1024};
+    int block_sizes[] = {32, 64, 128, 256, 512, 768, 1024};
```
karpathy, line 482: I realized we should have been adding this more places btw, when I saw that some GPUs often have thresholds set on 768, e.g. number of max threads per SM on some GPUs and such.

PR #410, opened 2024-05-14
Small update of matmul_backward_bias_kernel1

lancerts:
Results are not changed after the change.

Logically, it should be the same as
https://github.com/karpathy/llm.c/blob/master/dev/cuda/matmul_backward.cu#L102

Did I miss anything? @ngc92


ngc92:
the  bias kernel needs to add to the result, because it might be required for gradient accumulation

ngc92:
it's not merged yet, but hopefully it is on its way

PR #412, opened 2024-05-14
[wip] gradient accumulation, another attempt

karpathy:
Doesn't work. On master, we reproduce our Python script (almost) exactly by running:

```
make train_gpt2cu NO_MULTI_GPU=1 USE_CUDNN=1
./train_gpt2cu -b 4 -t 64 -l 1e-4 -v 200 -s 200 -a 1 -x 10
```

But doing this on this branch fails and the loss stagnates, something is wrong with the optimization.
Running an actual training loop, the loss does not decrease.

debugging...

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -1233,7 +1233,7 @@ __device__ SoftmaxParams prepare_softmax_blockwide3(int idx, const floatX* inp, 
 template <bool WriteLogits = true, bool WriteProbs = false>
 __global__ void __launch_bounds__(1024, MAX_1024_THREADS_BLOCKS)
                 fused_classifier_kernel5(floatX* logits, floatX* losses, floatX* probs,
-                                         const floatX* dlosses, const int* targets,
+                                         const float dloss, const int* targets,
                                          int B, int T, int V, int P) {
     int idx = gridDim.x - (blockIdx.x+1); // reverse order for cache hits on matmul data
     int ix = targets[idx];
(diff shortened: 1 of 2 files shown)


karpathy:
(The Python version works and behaves as expected. E.g. bumping the total batch size from 256 to 512 or 1024 produces exactly identical results, as it should be, because the batch is fixed so doing gradient accumulation over many fwd bwd passes over it create all the same gradients, so the mean of those doesn't change and everything remains the same)

karpathy:
Issue was with streams, which were complex and aren't adding much perf anyway atm.
Threw away all streams, back to default 1 stream for now. 
Will be more careful re-adding streams in future commits.
Gradient accumulation seems to work ok now.
I did some debugging a bit with commands such as:

```bash
# fp32 test
# make train_gpt2cu NO_MULTI_GPU=1 PRECISION=FP32
# ./train_gpt2cu -b 4 -t 64 -l 1e-4 -v 200 -s 200 -a 1 -x 10 -f 0 -e gpt2_124M.bin

# bf16 single gpu test
# make train_gpt2cu NO_MULTI_GPU=1 USE_CUDNN=1
# ./train_gpt2cu -b 4 -t 64 -l 1e-4 -v 200 -s 200 -a 1 -x 10

# make train_gpt2cu USE_CUDNN=1
make train_gpt2cu USE_CUDNN=1
mpirun -np 4 ./train_gpt2cu -b 4 -t 64 -l 1e-4 -v 200 -s 200 -a 1 -x 10 -d 4096
```

And the way I tested things was to use the new `-d` flag to force gradient accumulation. e.g. if `-d 512`, then the total batch size is 512, forcing gradient accumulation to be 2. And because the batches here are hacked to be all same, you expect the same results exactly, because all the accumulations are producing the same gradients, and then averaging out to the identical point. (Note that for the multi-gpu test I had to also hack this `loader->process_rank = 0` inside dataloader init, otherwise the dataloader tries to be different on each process).

PR #413, opened 2024-05-14
Add support for AMD devices

anthonix:
This unobtrusively adds support for AMD devices, in a way that minimizes changes or adding new code. 

Performance with bfloat16 on a 7900 XTX is ~50,000 toks/sec for a single GPU, and ~210,000 toks/sec for 4x GPUs (as a frame of reference, the latest pytorch 2.4.0.dev20240513 runs at ~42,000 toks/sec on a single device). 

Should this be merged here, or maintained as a separate fork?


anthonix:
FWIW, I've been playing around with more aggressive AMD specific optimizations that realize good gains over this, but I thought it was worth getting some feedback on how baseline AMD support could be integrated first.

karpathy:
This is very interesting to browse through and see! But yes, I think separate fork makes a lot more sense for AMD, and super happy to link to it in the notable forks section. It's cool that you seem to be getting some very nice throughputs!

anthonix:
Sounds good, will open another pull req with link to the fork in the README

PR #415, opened 2024-05-15
train_gpt.c: Removing the hardcoded GPT2_EOT

azret:
Removing the hardcoded GPT2_EOT and using the one from the tokenizer instead


karpathy:
Ah yes I changed the cuda code but not the .c code ty

PR #416, opened 2024-05-15
Setting up dev/cpu area with the first matmul_forward.c

azret:
@karpathy, Can we setup a dev/CPU area to play around?

@ngc92 I added your new matmul_forward as a first CPU kernel example.


karpathy:
hmm sure, i think so!

PR #417, opened 2024-05-15
Remove parallel CUDA streams while keeping main_stream and loss_event(?)

ademeure:
See discussion on Discord, I think whatever we eventually architect that's better than my naive folly will probably still need something similar to "main_stream" that's the default for everything, so I kept that (and added it for the one kernel where it was missing)

Also loss_event is a good example of something that kinda works as intended and it allows forward and backward to overlap, so kept it for now, to show this CPU/GPU parallelism can work even with just a single "main_stream" (we should be able to do the same for the main loop to avoid syncing for timing). But it'd be good to abstract it in a safer way somehow, so definitely wouldn't disagree if you wanted to remove it too!


karpathy:
I merged a version of this earlier, closing

PR #419, opened 2024-05-16
Include a matmul_backward_bias kernel based on PMPP CoarsenedSumReduction kernel in 10.15

lancerts:
The performance of the kernel depends on 
```
    const int coarse_factor 
    const int block_size_y 
```
With the setting in the PR,
On the 3070,

```
Using kernel 7


block_size 32 time 0.1357 ms
block_size 64 time 0.1362 ms
block_size 128 time 0.1352 ms
block_size 256 time 0.1359 ms
block_size 512 time 0.1366 ms
block_size 1024 time 0.1419 ms


Using kernel 8

block_size 32 time 0.1567 ms
block_size 64 time 0.1458 ms
block_size 128 time 0.1464 ms
block_size 256 time 0.1366 ms
block_size 512 time 0.1367 ms
block_size 1024 time 0.1387 ms


Using kernel 9

block_size 32 time 0.1376 ms
block_size 64 time 0.1368 ms
block_size 128 time 0.1365 ms
block_size 256 time 0.1388 ms
block_size 512 time 0.1381 ms
block_size 1024 time 0.1868 ms
```


karpathy:
Why delete 768 block size

lancerts:
> Why delete 768 block size

Opps, a bit messed up from a different local dev branch. Thanks for catching it,

ademeure:
Nice, I'm impressed perf is so close to kernel 7/8, probably the same approach but using f128 would be a tiny bit faster than 7/8 then! (I have only read an early edition of PMPP a long time ago, I should probably read through the new edition sometime!)

Maybe the most important thing to focus on in train_gpt2cu for that kernel is making it fully deterministic though, it felt like ngc92's kernel8 was an especially good starting point for that, but haven't really thought through what it would take in kernel9's case tbh.

lancerts:
> Nice, I'm impressed perf is so close to kernel 7/8, probably the same approach but using f128 would be a tiny bit faster than 7/8 then! (I have only read an early edition of PMPP a long time ago, I should probably read through the new edition sometime!)
> 
> Maybe the most important thing to focus on in train_gpt2cu for that kernel is making it fully deterministic though, it felt like ngc92's kernel8 was an especially good starting point for that, but haven't really thought through what it would take in kernel9's case tbh.

Yeah, totally, f128 should further speed it up. Just want to start simple and readable. Originally, just wanted to see how good the PMPP book kernel was compared to the highly optimized kernels in the repo.  The perf does vary wrt `coarse_factor` and `block_size_y` which requires GPU-specific tunning.

lancerts:
@karpathy rebased to main, lmk if there are additional comments, thanks.

PR #420, opened 2024-05-16
(optionally) recompute gelu activations to reduce activation memory

ngc92:
Activation  memory:
5706 MiB -> 5178 MiB  (-10%)
Runtime:
210.25 ms -> 211.69 ms (+0.7%)

we probably could get away with not using `act_sizes[12]` at all, but the recent cuda streams problems make a good case against premature optimization. Let's keep it like this for now, and figure out a nice unified solution for _all_ scratch buffers later on.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -1766,7 +1766,7 @@ typedef struct {
     floatX* output;
 } ActivationTensors;
 
-void fill_in_activation_sizes(size_t* act_sizes, size_t B, size_t T, GPT2Config config) {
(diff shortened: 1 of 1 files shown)


ademeure:
Very nice! I'd personally keep it as a configurable setting (especially if we end up adding more recomputable layers which are less of an obvious win) but use ternary operators rather than if/else, those are still very short lines of code, e.g.:
```
// if we recompute gelus, we just use the scratch buffer here
act_sizes[12] = recompute ? (B * T * 4*C) : (L * B * T * 4*C);
```

ngc92:
closing in favour of #422

PR #429, opened 2024-05-18
improved numerical error checking

ngc92:
*  tighter tolarances
*  relative tolerance based of bf16 epsilon
*  less verbose output if all is OK

I've checked these tolerances with both RTX 4060Ti and A4000 (which do give different errors, sometimes substantially). Values are chosen as tight as possible such that on neither GPU we exceed 90% of the error threshold, to allow for some more variety across GPUs.

### DIFF
--- a/test_gpt2.cu
+++ b/test_gpt2.cu
@@ -8,11 +8,15 @@ int check_tensor(float *a, float *b, int n, const char* label, float threshold=1
     int ok = 1;
     float max_diff = 0.0f;
     float max_rel_error = 0.0f;
+    float max_to_threshold = 0.f;
     float max_a = 0.0f;
     float max_b = 0.0f;
-    printf("%s\n", label);
+    float epsilon = 0.079;      // BF16 epsilon value
+    printf("%8s: ", label);
     for (int i = 0; i < n; i++) {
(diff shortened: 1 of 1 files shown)


test_gpt2.cu
```diff
@@ -8,33 +8,43 @@ int check_tensor(float *a, float *b, int n, const char* label, float threshold=1
...
-                max_diff, max_rel_error, max_a, max_b);
+        printf("TENSOR NOT OK, max diff: %.3e, with rel error: %.3e (calculated=%10f, ref=%10f), %.2f%% of maximum error\n",
+                max_diff, max_rel_error, max_a, max_b, max_to_threshold*100);
+    }
+
+    if(ok == 0) {
+        for (int i = 0; i < print_upto; i++) {
+            float t_eff = threshold + fabs(b[i]) * epsilon;
+            float diff = fabsf(a[i] - b[i]);
+            printf(diff <= threshold ? "OK " :  "NOT OK ");
```
karpathy, line 44: i think this condition here is a bug, it should match the one up above, i'll fix.

ngc92: yes

PR #432, opened 2024-05-19
only save missing bits to reconstruct fp32 master weights

ngc92:
I think I managed to get the bit-fiddling right, and this will effectively give us fp31 master parameters at the cost of only 16 additional bits (instead of the current 32).

Before merging, the code really needs to be critically reviewed, I'm especially unsure about the stochastic rounding code (why do we have two calls to the RNG for each rounding operation?).


gordicaleksa:
I assume two calls are due to the fact that we don't want each thread in the kernel to do stochastic rounding with the same seed.

At least that was the idea, but we'll end up having same seed eitherway because it's pseudorandom. :') 

Thus I'd say it's likely a bug.

llmc/bits.cuh
```diff
@@ -0,0 +1,97 @@
...
+__host__ __device__ SplitFloatResult split_float(float value, unsigned short threshold) {
+    unsigned int float_bits = bit_cast<unsigned int>(value);
+    // IEEE 754: float: S E(8) M (23)    bfloat: same, but significant 23-16 = 7 bits
+    // ideally, we'd just store the cut-off 16 bits, but that doesn't work if rounding
+    // is involved.
+    unsigned int rounded_bits = float_bits & 0x0000FFFFu;
+    if(rounded_bits > threshold) {
+        SplitFloatResult result;
+        result.b_float = __float2bfloat16_rn(bit_cast<float>(float_bits | 0xFFFFu));
+        result.bits = rounded_bits & (~1u) | 1u;
```
gordicaleksa, line 58: Why not just do "| 1u" why clear the bit first? The end result is equivalent?

llmc/bits.cuh
```diff
@@ -0,0 +1,97 @@
...
+// Reassembles a float from the bfloat16 part and the missing mantissa
+__host__ __device__ float assemble_float(nv_bfloat16 b_float, unsigned short bits) {
+    constexpr const unsigned short BF16_SIGN_MASK        = 0b1'00000000'0000000u;
+    constexpr const unsigned short BF16_EXPONENT_MASK    = 0b0'11111111'0000000u;
+    constexpr const unsigned short BF16_SIGNIFICANT_MASK = 0b0'00000000'1111111u;
+    unsigned short bf = bit_cast<unsigned short>(b_float);
+    if(bits & 1u) {
+        // if we rounded away from zero, we need to undo these changes.
+        // first, check if the significant (7 bits) of bf16 is zero
+        const unsigned short significant = bf & BF16_SIGNIFICANT_MASK;
```
gordicaleksa, line 78: nit: i think people usually refer to this one as `significand`

ngc92: right. Mantissa might actually be the correct term here. From wikipedia:
> To understand both terms, notice that in binary, 1 + mantissa ≈ significand, and the correspondence is exact when storing a power of two. This fact allows for a fast approximation of the base-2 logarithm, leading to algorithms e.g. for computing the [fast square-root](https://en.wikipedia.org/wiki/Methods_of_computing_square_roots#Approximations_that_depend_on_the_floating_point_representation) and [fast inverse-square-root](https://en.wikipedia.org/wiki/Fast_inverse_square_root). The implicit leading 1 is nothing but the hidden bit in IEEE 754 floating point, and the bitfield storing the remainder is thus the mantissa.

llmc/adamw.cuh
```diff
@@ -35,15 +34,36 @@ __global__ void adamw_kernel3(Tp* params_memory, float* master_params_memory, Tg
...
-    stochastic_rounding(param, &params_memory[idx], random);
-    // write the full, float version of the param into our master copy, if we maintain one
-    // this will be used in the next update
-    if (master_params_memory != NULL) { master_params_memory[idx] = param; }
+
+    // If we keep master parameter "copies", make sure to store the missing bits in the 'mantissas' array,
+    // otherwise we can directly go for stochastic rounding.
+    if (mantissas != NULL) {
+        if constexpr (std::is_same_v<Tp, __nv_bfloat16>) {
+            unsigned int random = Get2dNoiseUint(threadIdx.x, blockIdx.x, seed);
```
gordicaleksa, line 58: I think this is the only thing that'd have to get modified see the PR I created: https://github.com/karpathy/llm.c/pull/597

Other than this - lgtm!

llmc/bits.cuh
gordicaleksa: good point! mantissa is definitely the right term then

llmc/bits.cuh
ngc92: made this consistent, and added a comment explaining what the 1 and 0 mean for good measure

llmc/bits.cuh
ngc92: changed to mantissa

rosslwheeler:
@ngc92 - this is the other PR you were referring to for testing?

PR #437, opened 2024-05-19
AdamW constant pre-calculation outside of kernel

ChrisDryden:
This PR does the precalculations all of the constants multiplications with one another and reduces the amount of parameters required to be passed into the adamw kernel.

This PR also removes the bounds check inside the kernel and replaces it with an assert where the kernel is being called.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -1147,11 +1147,10 @@ __device__ float lerp(float start, float end, float weight) {
 }
 
 template <typename Tp, typename Tg>
-__global__ void adamw_kernel3(Tp* params_memory, float* master_params_memory, Tg* grads_memory, float* m_memory, float* v_memory, size_t num_parameters,
-                              float learning_rate, float beta1, float beta2, float beta1_correction, float beta2_correction, float eps, float weight_decay,
+__global__ void adamw_kernel3(Tp* params_memory, float* master_params_memory, Tg* grads_memory, float* m_memory, float* v_memory,
+                              float beta1, float beta2, float learning_rate_inv_beta1_correction, float inv_beta2_correction, float eps, float learning_rate_weight_decay,
                               float grad_scale, unsigned int seed) {
-    int idx = blockIdx.x * blockDim.x + threadIdx.x;
-    if (idx >= num_parameters) { return; }  // guard
+    size_t idx = blockIdx.x * blockDim.x + threadIdx.x;
 
     // get the gradient, m, and v for this parameter
     float grad = grad_scale * (float)grads_memory[idx];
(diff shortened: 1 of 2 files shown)


train_gpt2.cu
```diff
@@ -2412,13 +2411,14 @@ float gpt2_update(GPT2 *model, float learning_rate, float beta1, float beta2, fl
     // AdamW update
     int block_size = 512;
     int num_blocks = CEIL_DIV(num_parameters, block_size);
-    float beta1_correction = 1.0f - powf(beta1, t);
-    float beta2_correction = 1.0f - powf(beta2, t);
     unsigned int seed = random_u32(&model->rng_state);
+    float learning_rate_inv_beta1_correction = learning_rate*(1.0f/1.0f - powf(beta1, t));
+    float inv_beta2_correction = 1.0f/(1.0f - powf(beta2, t));
+    float learning_rate_weight_decay = learning_rate * weight_decay;
+    assert(num_parameters % block_size == 0); //bounds check
```
ngc92, line 2418: this check is cheap, essential for correctness, and whether it fails or not is dependent  on external program input, so maybe that should be a "real" check, and not an assert that may be turned off for a "release-mode" compile.

train_gpt2.cu
```diff
@@ -1147,11 +1147,10 @@ __device__ float lerp(float start, float end, float weight) {
 }
 
 template <typename Tp, typename Tg>
-__global__ void adamw_kernel3(Tp* params_memory, float* master_params_memory, Tg* grads_memory, float* m_memory, float* v_memory, size_t num_parameters,
-                              float learning_rate, float beta1, float beta2, float beta1_correction, float beta2_correction, float eps, float weight_decay,
+__global__ void adamw_kernel3(Tp* params_memory, float* master_params_memory, Tg* grads_memory, float* m_memory, float* v_memory,
+                              float beta1, float beta2, float learning_rate_inv_beta1_correction, float inv_beta2_correction, float eps, float learning_rate_weight_decay,
                               float grad_scale, unsigned int seed) {
-    int idx = blockIdx.x * blockDim.x + threadIdx.x;
-    if (idx >= num_parameters) { return; }  // guard
```
ngc92, line 1154: I'd be in favour of keeping the check. it doesn't cost us anything (measurable), and not having it there makes  the kernels brittle. If we change the kernel to vectorized loads, our requirement reg. the number of parameters in the model suddenly may depend on the precision we use for our weights.

karpathy:
@ChrisDryden  see the comments above about bounds checks, but also i'm curious does this have any benefit to merge, in terms of speed or something that you've measured? It's not computationally optimal but I do like that the AdamW algorithm is just very simply laid out in the kernel, and it's very clear what is happening there.

ChrisDryden:
Was testing this with the newest changes and I'm only getting a speedup of 3 microseconds on average in the kernel testing suite, doesn't seem worth the tradeoff. Will close it for now.

PR #438, opened 2024-05-20
Added warpsize as a constant expr for dev/cuda files

lancerts:


dev/cuda/attention_forward.cu
```diff
@@ -61,6 +61,9 @@ version 11 is kernel 10 skipping FP16/FP32 conversions (full FP16/BF16 network)
 // CUDA & cuDNN setup
 static bool first_run_validation = true; // always run e.g. permute on 1st run
 
+// WarpSize set as the compile time constant, this allows the compiler to optimize
+constexpr int WARP_SIZE = 32;
```
ngc92, line 65: this should probably go into common.h

PR #440, opened 2024-05-20
refactor datasets

karpathy:
Refactor how we treat datasets, because we're about to have more of them and we don't want them to clutter up root dir etc. this is only step 1, i'm about to refactor a bunch of the dataloading, how the .bin files work and are loaded, how the DataLoader works, etc. This is all needed to support good evals and training at scale.


rosslwheeler:
By no means exhaustive, but tried your feature branch with #440 and it seems fine in several of the tests I ran.

PR #442, opened 2024-05-21
Fully deterministic encoder backward kernels

ademeure:
This is a complete rewrite of the encoder backward pass, splitting it into two kernels (wte and wpe) which are both fully deterministic as they do not use atomics (assuming the seed for stochastic rounding is also deterministic, ofc).

The performance of the wte kernel and the associated CPU overhead are quite input-dependent, it might be quite slow for pathological cases (e.g. all tokens are the same). I tried updating /dev/cuda/ and there were a few problems. Most importantly, the performance on the random inputs was bad on the CPU side, and there was no way to hide the significant CPU overhead unlike in the real training loop.

Interestingly, profile_gpt2.cu is also broken when profiling the *current* encoder_backward: it is unrealistically fast (or more likely the train_gpt2.cu one is abnormally slow and needed fixing but not worth debugging it now...) - so in profile_gpt2.cu, this new kernel looks like it is much slower, but...

Performance actually improves by nearly ~1% on my RTX 4090 with a batch size of 24:

> Current: 106213 tokens/s
> New: 107216 tokens/s (+0.94%)

Unfortunately, while this made training *nearly* deterministic a week ago with only layernorm_backward still using atomics which didn't seem to noticeably impact training loss, the addition of gradient clipping seems to cause a lot of non-determinism, so we will also need to make that kernel deterministic to be identical bit-for-bit between runs (and make sure we save/restore the seed if we pause/restart a training run).

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -38,6 +38,9 @@ This reads & runs in fp32, B=4, T=64, LR=1e-4, val/sample never (200),
 #include <stdio.h>
 #include <stdarg.h>
 #include <string>
+#include <vector>
+#include <functional>
+#include <unordered_map>
 // GPU / CUDA related
 #include <cuda_runtime.h>
 #include <cublas_v2.h>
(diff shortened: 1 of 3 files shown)


train_gpt2.cu
```diff
@@ -1354,14 +1415,70 @@ void encoder_forward(floatX* out,
...
+            buckets[c_group + num_c_groups * inputs_cpu[bt]].push_back(data);
+            total_items++;
+        }
+    }
+
+    // Step 2: Sort buckets by size in descending order
+    // this is so the largest buckets are processed first by the GPU
+    // otherwise, if they started late, they would still be running with the rest of the GPU idle
+    std::vector<std::pair<uint64_t, std::vector<uint64_t>>> sortedBuckets(buckets.begin(), buckets.end());
+    std::sort(sortedBuckets.begin(), sortedBuckets.end(), // ugly because we don't have a typedef for the std::pair
```
ngc92, line 1452: why not just make a local typedef here?

PR #444, opened 2024-05-21
extend dataloader to be sharded

karpathy:


karpathy:
(WIP) needs more changes before it can be merged because the DataLoader now accepts filename glob patterns for train/val splits, so we re-define how `-i` flag works and what it means, and introduce a new `-j` flag for the val split.

PR #445, opened 2024-05-21
Fully deterministic layernorm backward

ademeure:
This causes a small ~0.3% performance loss on my RTX 4090, possibly worse on a A100 since that kernel might be a slightly larger % of total runtime.

It does the 2nd part of the reduction in a single block at the end of the same kernel, so it is extremely occupancy/latency limited, it might be possible to split the 2nd part of the kernel into a separate kernel and optimise it completely differently for better performance. But this feels good enough for now...

This also fixes a bug due to a missing CEIL_DIV that means the XL model with 1600 hidden dimension would not have worked correctly.

After this and encoder_backward, the only missing piece to get rid of all atomics (=> hopefully fully deterministic) is changing global_norm_squared_kernel to do the final reduction on the CPU.

### DIFF
--- a/dev/cuda/layernorm_backward.cu
+++ b/dev/cuda/layernorm_backward.cu
@@ -856,6 +856,185 @@ __global__ void __launch_bounds__(1024, MAX_1024_THREADS_BLOCKS)
     }
 }
 
+__global__ void layernorm_backward_kernel9(floatX* dinp, floatX* dweight, floatX* dbias, float* scratch,
+                                            const floatX* dout, const floatX* inp, const floatX* weight,
+                                            const floatX* mean, const floatX* rstd,
+                                            int B, int T, int C) {
+    constexpr int WARP_SIZE = 32;
+    int BLOCK_SIZE = blockDim.x;
+    int warpsInBlock = BLOCK_SIZE / WARP_SIZE; //number of warps in block
+    extern __shared__ float shared[]; // size = 2 * C + 1
+
+    int warpId = threadIdx.x / WARP_SIZE; // warp index within a block
+    int baseIdx = blockIdx.x * warpsInBlock + warpId;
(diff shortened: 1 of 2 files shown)


dev/cuda/layernorm_backward.cu
```diff
@@ -856,6 +856,185 @@ __global__ void __launch_bounds__(1024, MAX_1024_THREADS_BLOCKS)
...
+                dnorm_mean += dnorm_i;
+                dnorm_norm_mean += dnorm_i * norm_bti;
+            }
+        }
+        dnorm_mean = warpReduceSum(dnorm_mean) / C;
+        dnorm_norm_mean = warpReduceSum(dnorm_norm_mean) / C;
+
+        // now iterate again and accumulate all the gradients
+        // unfortunately we cannot use the same index for x128 arrays and shared memory
+        // as atomics can only be 32-bit rather than 128-bit (at least pre-SM90/Hopper)
```
ngc92, line 918: this comments seems out of data, there shouldn't be any (float) atomics here anymore for the deterministic  kernel

dev/cuda/layernorm_backward.cu
```diff
@@ -856,6 +856,185 @@ __global__ void __launch_bounds__(1024, MAX_1024_THREADS_BLOCKS)
...
+            store128cg(dinp_bt + global_index, dinp128);
+        }
+    }
+    __syncthreads();
+    // Each block writes its partial sum to global memory
+    // The last block to finish becomes responsible for summing up all the partial sums
+    // This is done by atomically incrementing a flag (cleared to 0 before launching the kernel)
+    unsigned int* scratchFlag = (unsigned int*)(scratch);
+    // Increment scratch pointer by a full cacheline so that everything remains cacheline aligned
+    scratch += 32;
```
ngc92, line 978: shouldn't this be an atomic add?

PR #447, opened 2024-05-22
HellaSwag eval in C

karpathy:
This was not super easy but ... first draft, apparently this works. needs cleanups, and also we are not yet utilizing the full batch dimension. we actually have to load in multiple examples and fully utilize batch.

TODOs still before merging:
- remove TODOs and prints and debugs
- make HellaSwag eval optional of course
- fully utilize the batch dimension by packing in more examples. Right now we only utilize the first 4 rows of the `BxT` batch.

But it does work, including multiple GPUs, and matches PyTorch reference.

### DIFF
--- a/dataloader.h
+++ b/dataloader.h
@@ -214,26 +214,33 @@ Copy pasting the section on the eval datafile format, from data_common.py:
     - <NUM><COMPLETION_TOKENS>, repeated NUM_COMPLETIONS times
 */
 
+// for now, could relax later
+#define ASSUMED_NUM_COMPLETIONS 4
+// helper macro for ceildiv
+#define CEIL_DIV(M, N) (((M) + (N)-1) / (N))
+
 typedef struct {
     // variables related to distributed training
     // each process/worker has to access different parts of the data
     int process_rank;
     int num_processes;
     // hyperparameters. use size_t to prevent overflow
-    size_t B;
-    size_t T;
+    size_t B; // (micro) batch size dimension of the tensor that feeds into the model
+    size_t T; // maximum context length of the model
     // input handling and its state
     FILE* eval_file;
     long file_size;
     uint16_t* buffer; // we fread data from file into this buffer
     // public variables that could be accessed from outside
     int num_examples; // in total across all processes
+    int num_batches; // to process the entire dataset across all processes
     int start_example_index; // the assignment of work for this process, start
     int end_example_index; // and end. start is inclusive, end is exclusive
+    int current_example_index; // the next example we would read
     int* inputs;  // input tokens into transformer
     int* targets; // target tokens for the transformer
     char* mask; // mask=1 at all completion token locations
-    int label; // the correct completion label
+    int* label; // the correct completion labels
     int num_completions; // number of completions for this example
 } EvalLoader;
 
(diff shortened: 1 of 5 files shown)


dataloader.h
```diff
@@ -195,3 +195,173 @@ void dataloader_free(DataLoader *loader) {
...
+void evalloader_reset(EvalLoader *loader) {
+    // we have to be careful that each process starts at the correct offset.
+    // For example if there are N examples in the file and 4 processes,
+    // then process 0 should start at 0, process 1 at N/4, process 2 at N/2, etc.
+    long header_bytes = HEADER_SIZE * sizeof(int);
+    // determine which example we want this process to start at
+    int process_stride = loader->num_examples / loader->num_processes;
+    loader->start_example_index = process_stride * loader->process_rank;
+    loader->end_example_index = process_stride * (loader->process_rank + 1);
+    if (loader->end_example_index > loader->num_examples) {
```
ngc92, line 249: can this actually happen  with the current code? `process_stride` should be DIV_CEIL'ed?, and then we get a slightly smaller last chunk, right?

dataloader.h
```diff
@@ -195,3 +195,173 @@ void dataloader_free(DataLoader *loader) {
...
+*/
+
+typedef struct {
+    // variables related to distributed training
+    // each process/worker has to access different parts of the data
+    int process_rank;
+    int num_processes;
+    // hyperparameters. use size_t to prevent overflow
+    size_t B;
+    size_t T;
```
ngc92, line 224: maybe add a comment regarding the exact meaning of T here. Is is max_seq_len of the model? Is it the longest example in the eval set? is it the longest example in  current `inputs`/`targets`?

dataloader.h
```diff
@@ -195,3 +195,173 @@ void dataloader_free(DataLoader *loader) {
...
+    freadCheck(header, sizeof(int), HEADER_SIZE, loader->eval_file);
+    if (header[0] != 20240522) { printf("Bad magic in eval file\n"); exit(EXIT_FAILURE); }
+    if (header[1] != 1) { printf("Bad version in data file\n"); exit(EXIT_FAILURE); }
+    loader->num_examples = header[2]; // number of tokens in the file
+    assert(loader->num_examples >= num_processes); // avoid headaches for now
+    size_t longest_example_bytes = header[3]; // longest example in the file
+    // basic sensibility check we could relax later. but roughly it's mostly
+    // the prompt/context and 4 completions, 2 bytes/token, so the longest example
+    // should be well below 5 times the context length or so (approx. napkin math)
+    assert(longest_example_bytes > 0 && longest_example_bytes < 5*T*2);
```
ngc92, line 295: I don't understand this condition

dev/data/data_common.py
```diff
@@ -45,3 +45,64 @@ def write_datafile(filename, toks):
...
+    # construct the header
+    header = np.zeros(256, dtype=np.int32)
+    header[0] = 20240522 # magic
+    header[1] = 1 # version
+    header[2] = len(datas) # number of examples
+    header[3] = 0 # reserved for longest_example_bytes, fill in later
+    # now write the individual examples
+    longest_example_bytes = 0 # in units of uint16s
+    full_stream = [] # the stream of uint16s, we'll write a single time at the end
+    assert len(datas) < 2**16, "too many examples?"
```
ngc92, line 72: where does this condition come from?

train_gpt2.cu
```diff
@@ -2729,11 +2740,51 @@ int main(int argc, char *argv[]) {
                 val_loss += model.mean_loss;
             }
             val_loss /= val_num_batches;
-            val_loss = multi_gpu_cpu_float_mean(val_loss, &multi_gpu_config);
+            val_loss = multi_gpu_cpu_float_sum(val_loss) / multi_gpu_config.num_processes;
             printf0("val loss %f\n", val_loss);
             logger_log_val(&logger, step, val_loss);
         }
 
+        // once in a while estimate HellaSwag accuracy
```
ngc92, line 2748: maybe move this into a dedicated function? to keep things a bit more manageable

dataloader.h
karpathy: Yes, refactoring this now.

dev/data/data_common.py
karpathy: It can't be more than 2**16 because we write len(datas) to the file, in a uint16 field. so it would overflow otherwise

train_gpt2.cu
karpathy: yeah... it's also about to get worse

dev/data/data_common.py
ngc92: why not just use an uint32 there, for forward compatibility? Its not like we need to save these two bytes

karpathy: len(datas) is only like 10K and probably won't grow too much, i think i'll keep as uint16 because everything else is

PR #451, opened 2024-05-23
init with random weights

karpathy:
first draft of random init, crashes with some cuBLAS error, debugging

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -2635,9 +2635,9 @@ void logger_init(Logger *logger, const char *filename) {
     if (filename != NULL) { logger->logfile = fopenCheck(filename, "w"); }
 }
 
-void logger_log_eval(Logger *logger, int step, float val_loss) {
+void logger_log_eval(Logger *logger, int step, float val) {
     if (logger->logfile != NULL) {
-        fprintf(logger->logfile, "s:%d eval:%.4f\n", step, val_loss);
+        fprintf(logger->logfile, "s:%d eval:%.4f\n", step, val);
     }
 }
 
@@ -2900,7 +2900,7 @@ int main(int argc, char *argv[]) {
             // careful because not all ranks may have the exact same allocation of number of examples
             eval_acc_norm = multi_gpu_cpu_float_sum(eval_acc_norm);
             printf0("HellaSwag: %d/%d = %f\n", (int)eval_acc_norm, eval_loader.num_examples, eval_acc_norm / eval_loader.num_examples);
-            logger_log_eval(&logger, step, eval_acc_norm);
+            logger_log_eval(&logger, step, eval_acc_norm / eval_loader.num_examples);
         }
 
         // once in a while do model inference to print generated text


karpathy:
This branch now initializes from scratch optionally, and the init exactly matches that of PyTorch.

Here is an example of how I am doing an actual production run: training from scratch, on FineWeb, evaluating on HellaSwag, on my 4XA100 40GB:

```bash
make train_gpt2cu USE_CUDNN=1
./train_gpt2cu -i "dev/data/fineweb10B/fineweb_train_*.bin" \
               -j "dev/data/fineweb10B/fineweb_val_*.bin" \
               -v 250 -s 250 -g 144 -o fineweb.log -b 32 \
               -e "d12"
```

the `-e "d12"` flag is new. `-e` was previously used to point to a model checkpoint to load weights from. But if this flag has the format `-d{number}`, we interpret that number as the depth of model, and initialize it from scratch.

For posterity potentially, this is some additional scratch code that I found helpful for making sure we match PyTorch. You could place this right around line 2120, right above `// copy them to GPU`: 

```c
// DEBUGGING
// because our implementation is identical to PyTorch, we should have identical weights here
// load the checkpoint that was saved in PyTorch and compare the weights
FILE *model_file = fopenCheck("gpt2_d12_bf16.bin", "rb");
int model_header[256];
freadCheck(model_header, sizeof(int), 256, model_file); // skip header
floatX* params_memory_cpu_pytorch = (floatX*)mallocCheck(model->num_parameters_bytes);
freadCheck(params_memory_cpu_pytorch, 1, model->num_parameters_bytes, model_file);
// so these two should be identical now
// we have to instead go tensor by tensor because only some of them seem to match
offset = 0;
int all_ok = 1;
for (int i = 0; i < NUM_PARAMETER_TENSORS; i++) {
    size_t n = model->param_elements[i];
    int ok = 1;
    printf("Checking tensor %d\n", i);
    float max_diff = 0.0f;
    for (size_t j = 0; j < n; j++) {
        floatX ours = params_memory_cpu[offset + j];
        floatX torch = params_memory_cpu_pytorch[offset + j];
        float diff = fabsf((float)ours - (float)torch);
        if (j < 10) {
            printf("%d:%lu ours: %f, torch: %f, diff: %f\n", i, j, (float)ours, (float)torch, diff);
        }
        if (diff > 0.000245) {
            ok = 0;
            printf("Tensor %d diff too large at %lu: %f\n", i, j, diff);
            exit(0);
        }
        if (diff > max_diff) {
            max_diff = diff;
        }
    }
    printf("Tensor %d max_diff: %f => %s\n", i, max_diff, ok ? "OK" : "FAIL");
    offset += n;
    all_ok &= ok;
}
printf("All tensors OK: %d\n", all_ok);
exit(EXIT_FAILURE);
```

HAPPY

karpathy:
Oh one more note (CC @azret ) I can't get bitwise match. Note in my code above I am using `0.000245` as the threshold for the diff. This was the max_diff , a bit mysteriously, for all the tensors. I think there's some possibly bfloat16-related (?) issue and/or some remaining difference here? The two implementations don't exactly and fully match.

karpathy:
Example command to do a real prod run on my 4XA100 40GB box.

Training:
- from scratch
- on FineWeb
- eval on HellaSwag
- ZeRO-1, recompute GeLU, grad clip 1.0
- total batch size here is ~300M. For this model size, GPT-3 was trained batch size 1M, so this can probably go up more.

```bash
# fineweb from scratch! (GPT-2 XL (1558M))
# -z 1 : ZeRO-1
# -r 1 : recompute GeLU
# total batch size 294912 => gradient accumulation steps of 6
# 294912 = 6 grad_accum * 12 batch_size * 1024 seq_length * 4 GPUs
mpirun -np 4 ./train_gpt2cu -i "dev/data/fineweb10B/fineweb_train_*.bin" \
            -j "dev/data/fineweb10B/fineweb_val_*.bin" \
            -v 200 -s 200 -g 144 -o fineweb.log -b 12 \
            -d 294912 \
            -r 1 \
            -z 1 \
            -l 2e-4 \
            -e "d48"
```

Remaining: need temperature schedule, weight decay weights only, and a bit of hyperparam tweaking potentially, but ok as example.

TimZaman:
/offtopic I just looked at your github activity - beast mode

karpathy:
( @TimZaman I was on vacation last two weeks in Europe, but otherwise yes that is llm.c development hehe)

azret:
@karpathy, 0.000245 is a bit too high. I need to dig deeper to see see if torch does something special for BF16.

Relevant ref code:

https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/core/DistributionsHelper.h
https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cpu/DistributionTemplates.h
https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/core/TransformationHelper.h

1. We do have slight differences. e.g. Torch uses double math funcs sqtr, log sin, cos where as we do sqrtf, logf sinf etc..
2. Torch does log1p in some cases where as I used an EPSILONE 1e-12 for the log

But even with this differences we should be much closer to torch than 0.000245. I'll take a closer look.

PR #452, opened 2024-05-23
Add cmake and fix OMP 

avflyer:
1. Add cmake build rules.
2. Fix OMP compilation error in VS2022.
3. OpenMP can be turn off on windows.


karpathy:
I don't think I'll add cmake to the project. What is cmake fixing here that can't be done in our Makefile?

rosslwheeler:
Also - we already have a CMake PR https://github.com/karpathy/llm.c/pull/59

avflyer:
> I don't think I'll add cmake to the project. What is cmake fixing here that can't be done in our Makefile?

Adding cmake is nothing to do with bug fix, it's a combined commit. Cmake could be really helpful for cross-platform development when project grows big and configurations become more and more complicated, and more friendly to those who favor IDEs, like visual studio.  But i also respected what you decide here.

By the way, the fix is on another PR (https://github.com/karpathy/llm.c/pull/460)

PR #460, opened 2024-05-25
Fix MSVC compilation error with openMP

avflyer:
Fix MSVC compilation error with openMP
https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c3015?view=msvc-170

Besides, with this PR, openMP can also be disabled on windows paltform with 'NO_OMP=1' in makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ else
   endif
   CC := cl
   CFLAGS = /Idev /Zi /nologo /Wall /WX- /diagnostics:column /sdl /O2 /Oi /Ot /GL /D _DEBUG /D _CONSOLE /D _UNICODE /D UNICODE /Gm- /EHsc /MD /GS /Gy /fp:fast /Zc:wchar_t /Zc:forScope /Zc:inline /permissive- \
-   /external:W3 /Gd /TP /wd4996 /Fd$@.pdb /FC /openmp:llvm
+   /external:W3 /Gd /TP /wd4996 /Fd$@.pdb /FC
   LDFLAGS :=
   LDLIBS :=
   INCLUDES :=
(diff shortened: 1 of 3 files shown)


rosslwheeler:
Not following exactly - are you saying the current code generates a C3015?

rosslwheeler:
Our Windows CI and all current Windows builds are passing. What version of MSVC are you using? Please run: Cl /Bv

avflyer:
> Our Windows CI and all current Windows builds are passing. What version of MSVC are you using? Please run: Cl /Bv. Also, what options are you using with Make?

Well, I configured this project with visual studio 2022, with MSVC 19.39.33523.0. I copyed all the c_flags and ld_flags from Makefile. Then I encounterd this error (C3015).
6ab8246f-3955-4e76-86ef-85bc19ae7850

As you suggested, this time I tried the CI's way of building, which employs an external make.exe program. Yes, it succeeded. I don'y know what causes this difference.

Besides, to my knowlege, putting declaration of varariables which have possible impacted on OMP before the '#pragma omp parallel' macro could be much more safe without other drawbacks.

rosslwheeler:
Glad it works for you now when using make and the Makefile! 

As far as the other changes, let's discuss those over in the Issues area. Microsoft does have a compiler bug with OpenMP 5.2 support which I've filed with them.

ngc92:
I'm not 100% sure, but I thought putting the declarations in front of the openmp statements would actually totally break this code.
Aren't variables from outside the parallel region automatically shared?

avflyer:
> I'm not 100% sure, but I thought putting the declarations in front of the openmp statements would actually totally break this code. Aren't variables from outside the parallel region automatically shared?

Thx for pointing out this. I checked the OMP usesage and found that I did have a misunderstanding of the memory model. declarations in front of the openmp statements will be implicitly handled as shared vars, not private. Adding the private(var) statement will make those vars private.

I have already fix the codes.

PR #466, opened 2024-05-26
.gitignore: ignore more for windows devs

nietras:
@karpathy suggested extra ignores for windows devs and e.g. if using local python "env" and CMake "out".


nietras:
no interest

PR #471, opened 2024-05-27
resume training v1

karpathy:
writes the files but doesn't load them yet, coming up in a bit


karpathy:
WINDOWS

rosslwheeler:
Looking at it.

PR #477, opened 2024-05-27
Added kernel development file for permute_backwards

ChrisDryden:
Created a development file for permute_backwards, will make one for permute backwards and permute forwards to create and test optimizations for these two kernels.


karpathy:
Not sure how much we care about this kernel

Issue #478, opened 2024-05-28
Recalculating the activations in the backwards pass to conserve memory


ChrisDryden:
@ngc92 Did an analysis of the areas that take up the most memory and its impact on the amount of batches that can be used and found that one of the largest contributors was the memory associated with the Layernorm recomputations:

```
Layernorm recomputation:
    2 layernorms in each of the 48 layers, 1600*2 bytes per token
    =>
    9600 MiB = 2*48*32*1024*1600*2

Missing-bits for master weights: (PR exists)
    optim-w from 32-bit to 16-bit
    372 MiB

FP16 optimizer states:  (PR exists, but very outdated)
    optim-m, optim-v from 32-bit to 16-bit
    743 MiB

ZeRO-2: (missing prerequisites: cuda streams, overlapping backward and comms)
    grads / nGPU
    2599 MiB
```

This Issue will track the implementation of adding the ability similar to how the GELU is recalculated in the backwards pass to recalculate the layernorm forwards activations so that we can reduce the memory.

ChrisDryden:
To start off, I will first implement the layernorm forward in the backwards pass implementation and use the ln1 and ln2 values directly from that layernorm forward to get an initial working version of recalculating the values in the backwards pass.

ChrisDryden:
In the above PR I was able to implement the reduced memory:
Went from:
allocating 1703 MiB for activations
val loss -inf
allocating 237 MiB for parameter gradients
allocating 30 MiB for activation gradients
allocating 474 MiB for AdamW optimizer state m
allocating 474 MiB for AdamW optimizer state v
allocating 474 MiB for master copy of params

To this:

allocating 1307 MiB for activations
val loss 4.504488
allocating 237 MiB for parameter gradients
allocating 30 MiB for activation gradients
allocating 474 MiB for AdamW optimizer state m
allocating 474 MiB for AdamW optimizer state v
allocating 474 MiB for master copy of params

ChrisDryden:
The PR was merged but still needs the second step of making a simplified kernel that doesnt recompute everything and reuses the values calculated in the forwards pass

PR #480, opened 2024-05-28
First step in creating option to recalculate layernorm activations in backwards pass

ChrisDryden:
This CR is the first step in implementing the goal described in https://github.com/karpathy/llm.c/issues/478 to be able to reduce the memory footprint by adding an option to recalculate the layernorm activations in the backwards pass. The next step would be to create another kernel which is a smaller version of the layernorm forward kernel. Once I get a review from @ngc92 I will implement the second part of this PR.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -2043,7 +2043,8 @@ void fill_in_activation_sizes(size_t* act_sizes, size_t B, size_t T, GPT2Config 
     size_t NH = config.num_heads;
     size_t C = config.channels;
     act_sizes[0] = B * T * C; // encoded
-    act_sizes[1] = L * B * T * C; // ln1
+    // if recompute >= 1 then we will recompute the layernorm forward activation during backward pass
+    act_sizes[1] = (recompute == 0) ? L * B * T * C : B * T * C; // ln1
     act_sizes[2] = L * B * T; // ln1_mean
     act_sizes[3] = L * B * T; // ln1_rstd
     act_sizes[4] = L * B * T * C; // atty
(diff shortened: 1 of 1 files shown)


ngc92:
lgtm; at some point we need to decide the exact meaning of the `recompute` flag, e.g., could be 
1 => recompute gelu,
2 => recompute gelu + ln
or more like a bitfield 1 => gelu, 2 => ln, 3 => gelu+ln

Note that we don't even need to have both ln1 and ln2; we could just reuse one of them

train_gpt2.cu
```diff
@@ -2682,6 +2684,9 @@ void gpt2_backward(GPT2 *model, int* inputs) {
         }
         matmul_backward(dl_bt4c, dl_fcprojw, dl_fcprojb, dresidual, l_fch_gelu, l_fcprojw, scratchF, B, T, 4*C, C);
         gelu_backward(dl_bt4c, l_fch, dl_bt4c, B*T*4*C);
+        if(model->recompute >= 1) {
+            layernorm_forward(acts.ln1, acts.ln1_mean, acts.ln1_rstd, l_residual2, params.ln1w, params.ln1b, B, T, C);
```
ngc92, line 2688: the ordering is reversed here. the next matmul requires ln2, not ln1

ngc92:
Turns out I was too optimistic when quickly skimming the code. There are two easily fixable mistakes:
1) order of the layernorms. In forward pass, it is ln1 -> ln2, but in backward, it is ln2 -> ln1
2) use paramters of the  current layer. params.ln1w gives the weights of the first layer's layernorm, not the current one.

ChrisDryden:
Updated it so that 1 only runs gelu and 2 runs gelu and the activation recalculation

karpathy:
Example of what I'm seeing

```
r0 630K tok/s, 15570GB
r1 620K tok/s 13445GB
r2 610 tok/s 12389GB
```

roughly, on 4X A100 40GB with -b 32

PR #486, opened 2024-05-28
MFU for other GPUs

ngc92:
extrapolating from spec sheets, yay :)


rosslwheeler (commented):
@ngc92 - the A5500 should be 1665 like you mentioned?

ngc92:
numbers for the workstation Ampere GPUs currently don't make sense, I get > 100%.

chinthysl:
Nice, I manually adjusted for H100 using the datasheet.

karpathy:
Merged to master, made a new file llmc/mfu.h, adjusted the code so that we don't use fancy features so also Windows doesn't break.

PR #487, opened 2024-05-29
`softmax_autoregressive_backward_kernel` does not use share memory in the kernel

huoyushequ:
`softmax_autoregressive_backward_kernel` does not use share memory in the kernel.

we do not need to launch the kernel with 256 bytes share memory, so remove it

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -1875,7 +1875,7 @@ void attention_backward(floatX* dinp, floatX* dqkvr, floatX* dpreatt, floatX* da
     // backward into preatt
     int hs = C / NH; // head size
     float scale = 1.0f / sqrtf(hs);
-    softmax_autoregressive_backward_kernel<<<dim3(T / 4, B * NH), 256, 256>>>(dpreatt, datt, att, B, T, C, scale);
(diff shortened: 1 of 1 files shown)


karpathy:
it's used inside `blockReduce` I believe?

huoyushequ:
oh, my mistake. the `blockReduce` use the static share memory, we do not need to specify it in the launch

huoyushequ:
'''__shared__ float shared_val[WARP_SIZE];''' float(4 bytes) x WARP_SIZE(32) = 128, I do not understand how the 256 in the launch comes

huoyushequ:
```shared float shared_val[WARP_SIZE];``` float(4 bytes) x WARP_SIZE(32) = 128, I do not understand how the 256 in the launch comes

karpathy:
A lot of this code changed relatively quickly recently, it may just be an oversight. I agree with you.

karpathy:
So I think this was done because at one point we were placing the kernels explicitly on streams, and the stream identity comes after the shared memory, so you have to put something there. But not sure why it's 256 instead of 128. Also of note this is the non-cudnn code path for explicit attention, so it is rarely exercised. Merging

huoyushequ:
thx for the merge

PR #488, opened 2024-05-29
cli: fix duplicate -c flag

rojas-diego:
The CLI flag `-c` was previously bound to both `grad_clip` and `weight_decay` the latter taking precedence over the other.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -3104,7 +3104,7 @@ void load_state(int* step, GPT2* model, DataLoader* loader, const char* filename
 
 // ----------------------------------------------------------------------------
 // CLI, poor man's argparse
-// unclaimed flags lol: k,p
+// unclaimed flags lol: k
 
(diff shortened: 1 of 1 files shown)


karpathy:
Gosh, wait this is bad right? This means any time we set -c to set the weight decay (to 0.1 usually), we are actually changing grad clip from its default. My preference mostly btw would be to just leave grad clip at 1.0 and not make it configurable at all.

karpathy:
Oh wait it's ok because they are else ifs, forgive my little heart attack. I'd suggest we leave grad clip at 1.0. Agree? Happy to do the fix, or feel free to adjust the PR.

rojas-diego:
Haha @karpathy I too had a rather short-lived mini-heart attack! Yeah I think keeping grad clip at 1.0 is reasonable. Especially since we're running out of CLI flags. I'll update the PR right now.

PR #489, opened 2024-05-29
Realtime training visualization using wandb

chinthysl:
Just an additional script to visualize and track metrics in realtime using wandb. This will be useful when have longer training runs and multinode training lasting many hours.

**Metric graphs from 2 training runs**
metric_graphs

**System overview**
system_graphs

### DIFF
--- a/dev/wandb_logger.py
+++ b/dev/wandb_logger.py
@@ -1,0 +1,48 @@
+import os
+import time
+import argparse
+import wandb
+
+# Please follow https://docs.wandb.ai/quickstart and setup wandb server and python client
+
+logging_metrics = {
+    'eval'      : 'HellSwag Evaluation',
+    'tel'       : 'Validation Loss',
+    'trl'       : 'Training Loss',
+    'norm'      : 'Normalized Gradient',
+    'lr'        : 'Learning Rate',
+    'latency'   : 'Latency Per Step (ms)',    
+    'mfu'       : 'A100 FP16 MFU %',
+    'tokens'    : 'Tokens Per Second'
+}
+
+def parse_line(line):
+    for metric, header in logging_metrics.items():
+        if metric in line:
+            parts = line.split()
+            step = int(parts[0].split(":")[1])
+            value = float(parts[1].split(":")[1])
+            wandb.log({header: value}, step=step)
+
+def wait_for_file(file_path):
+    while not os.path.exists(file_path):
+        print(f"Waiting for file {file_path} to be created...")
+        time.sleep(5)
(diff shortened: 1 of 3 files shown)


karpathy:
Ok yes this is probably a good idea 😅 . I'll leave some comments.

dev/wandb_logger.py
```diff
@@ -0,0 +1,48 @@
...
+    'tokens'    : 'Tokens Per Second'
+}
+
+def parse_line(line):
+    for metric, header in logging_metrics.items():
+        if metric in line:
+            parts = line.split()
+            step = int(parts[0].split(":")[1])
+            value = float(parts[1].split(":")[1])
+            wandb.log({header: value}, step=step)
```
karpathy, line 25: `parse_line` suggest we're parsing line, but then we do more than that and this is the actual logging line to wandb. either rename the function or parse_line returns a dict, and then call wandb on it outside or so.

train_gpt2.cu
```diff
@@ -3020,6 +3020,18 @@ void logger_log_train(Logger *logger, int step, float train_loss) {
     }
 }
 
+void logger_log_metrics(Logger *logger, int step, float norm, float lr, float latency, float mfu, float tokens) {
+    if (logger->active == 1) {
+        FILE *logfile = fopenCheck(logger->output_log_file, "a");
+        fprintf(logfile, "s:%d norm:%6.4f\n", step, norm);
```
karpathy, line 3026: At a previous unnamed company i got a bit burned with logging a zillion metrics to wandb, it bloated up files and made everything super slow, and people always just looked at the same 5 plots anyway. Do we want to record all of latency, mfu, and tok/s? Also, the variable here is `tokens` but actually it's `tokens_per_second`. These are all the same number, just converted one way or another. Hmm.

chinthysl: zillion 🤯. It’s true that the project becomes bloated. Let me refactor a bit.

chinthysl: @karpathy Maybe probably not a good idea to have this integration. Still somebody can have plotting mechanism which reads the output logs and produce really necessary graphs without bloating the system ☝️

karpathy:
One more thing to be careful with and think about 🤔 . If the process crashes or hangs and gets restarted, it starts to log again from the last checkpoint. i.e. some steps in the log will be repeated because we just append to the end. I don't know what wandb does in these cases. E.g. see run124M.sh for how I have a while loop over the job, and potentially CTRL+C it manually and see what happens

PR #490, opened 2024-05-29
Shared memory optimization in matmul_forward kernel1

vyom1611:
Optimizations:
1. Shared memory is used to load weights and inputs, which reduces global memory access latency. Each thread block loads a portion of the weight and input matrices into shared memory.

2. Both weights and inputs are loaded in a coalesced manner by threads within a block

- Also changed the launcher configuration corresponding with the changes

### DIFF
--- a/dev/cuda/matmul_forward.cu
+++ b/dev/cuda/matmul_forward.cu
@@ -52,25 +52,37 @@ void matmul_forward_cpu(float* out,
 // ----------------------------------------------------------------------------
 // GPU kernels
 
-// kernel 1: naive kernel, every thread handles one output element, direct global memory access
+// Optimized kernel 1: using shared memory for better performance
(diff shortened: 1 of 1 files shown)


chinthysl:
Is there a performance gain observed after changes? In both scenarios cuda threads have to access global memory and read the data. Without shared memory addition will happen on the fly using registers (a common practice in cuda programming). With shared memory, additional write and read has to be done to and from shared memory for the final result!

ngc92:
kernel 1 is supposed to be the absolute minimum, naive kernel, so it deliberately doesn't use shared memory.
if you want to see an example of how to use shared memory for a matrix multiplication, see e.g., here:
https://github.com/karpathy/llm.c/blob/master/dev/cuda/trimat_forward.cu

vyom1611:
Ah I see, thank you for clarifying.

PR #491, opened 2024-05-29
Merge the EPSILONE definitions in the normal_ and normal_fill_16 func…

yuanrongxi:
translates to English as "There are multiple definitions of EPSILONE in rand.h, and their values can be merged


karpathy:
it's ok

PR #494, opened 2024-05-29
pytorch parity v1

karpathy:
I believe we're getting the same results as `./run124M.sh` with this as of this PR:

```bash
# the same as run124M.sh but in PyTorch
# if you wish to train on just a single GPU, simply skip the torchrun part, i.e.
# python train_gpt2.py ... (all the other arguments the same)
torchrun --standalone --nproc_per_node=8 train_gpt2.py \
    --input_bin dev/data/fineweb10B/fineweb_train_000001.bin \
    --write_tensors 0 \
    --model d12 \
    --batch_size 32 \
    --sequence_length 1024 \
    --total_batch_size 524288 \
    --dtype bfloat16 \
    --compile 1 \
    --tensorcores 1 \
    --flash 1 \
    --num_iterations 18865 \
    --weight_decay 0.1 \
    --zero_stage 1 \
    --learning_rate 0.0006 \
    --warmup_iters 700 \
    --learning_rate_decay_frac 0.0 \
    --overfit_single_batch 0
```

but only for 1 data shard. the dataloader is still not set up to do multiple shards.

### DIFF
--- a/pyrun124M.sh
+++ b/pyrun124M.sh
@@ -1,9 +1,17 @@
 
-# the same as run124M.sh but in PyTorch
+# the same as run124M.sh but with PyTorch
+# current restrictions:
+# - does not write checkpoint, only logs of the train/val losses
+# - does not evaluate hellaswag accuracy
+# - cannot "resume training" (i.e. the `-y 1` flag)
+
 # if you wish to train on just a single GPU, simply skip the torchrun part, i.e.
 # python train_gpt2.py ... (all the other arguments the same)
 torchrun --standalone --nproc_per_node=4 train_gpt2.py \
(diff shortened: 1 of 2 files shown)


KellerJordan:
Hi Andrej thank you for your work

I just ran this code but hit a bug because of this line
https://github.com/karpathy/llm.c/blob/03c2e85601df77139857a16362472458ed45e6b7/train_gpt2.py#L369

I believe this update to the current position should come before the if statement at line 366 instead.

Being in the else block means that line 366 only checks that the new start position will be within range. But the new end position can be out of range, leading to a reshape error being thrown at line 363.

After fixing this, the code works fine. Thanks for your time

karpathy:
ty @KellerJordan  you're right of course and this is also how the dataloader.h code does it. ty for the flag

PR #498, opened 2024-05-30
Adding example of templatized blockdim for faster iteration speeds of .2ms

ChrisDryden:
Was doing research into compilation optimizations and discovered that the blockDim.x is not set at compile time even though all of our blockDim's are set as consts. When experimenting in the cuda dev kernels I found some kernels had signifigant speed improvements. 

Adding these as templated constants, I found that the speed increased by approximately .2ms/iter


karpathy:
I don't like templates, would have to be a larger gain.

PR #500, opened 2024-05-30
[Python code]: fix leak of validation examples into training loop

KellerJordan:
I was running this code with validation enabled every 128 steps when I noticed a strange pattern in the loss:

Screenshot 2024-05-30 at 10 41 34 AM

This is caused by the fact that the validation batches share a scope and name with the training batches. So, the network ends up training on the last batch of validation examples once after every run of validation.

This can be fixed by giving the valid batches a different name. I also gave the batches used for sampling a different name.

Thanks

### DIFF
--- a/train_gpt2.py
+++ b/train_gpt2.py
@@ -746,9 +746,9 @@ if __name__ == "__main__":
             with torch.no_grad():
                 val_loss = 0.0
                 for _ in range(args.val_max_steps):
-                    x, y = val_loader.next_batch()
-                    x, y = x.to(device), y.to(device)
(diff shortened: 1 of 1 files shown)


karpathy:
I diverged from the way train_gpt2.cu iterates because I had to save the first iteration's data to .bin files, but now we get this bug because of it. I don't actually love this solution because there is a deeper issue here. Basically I want to mirror the logic as it is in the train_gpt2.cu file. Let me see

karpathy:
I pushed a fix to master, have a look. closing.

PR #507, opened 2024-05-31
Added script to download all of the tokenized fineweb100B data from huggingface

ChrisDryden:
I created this to show an example of what it would look like to download all of the files from Huggingface, this is the first step and after all of the files are uploaded to Huggingface I will upload all of the compressed versions to the same location and then add the decompression steps.

### DIFF
--- a/dev/data/fineweb100B.sh
+++ b/dev/data/fineweb100B.sh
@@ -10,20 +10,45 @@ SAVE_DIR="fineweb100B"
 # Create the directory if it doesn't exist
 mkdir -p "$SAVE_DIR"
 
-# Download the single validation file
-VAL_FILE_NAME="fineweb_val_000000.bin"
-curl -L -o "${SAVE_DIR}/${VAL_FILE_NAME}" "$VAL_URL"
-echo "Downloaded ${VAL_FILE_NAME} to ${SAVE_DIR}"
+# Function to download, decompress, and delete files
+download_and_decompress() {
+    local FILE_URL=$1
+    local FILE_NAME=$(basename $FILE_URL | cut -d'?' -f1)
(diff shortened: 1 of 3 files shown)


ChrisDryden:
Also added the compressed version too, it will take a while to finish uploading but its around halfway done

rosslwheeler:
@ChrisDryden - for the compressed version: What about having all the shards compressed to a single file and downloading them as one file and then decompressing them? Not sure about the disk size impact.

ChrisDryden:
I was thinking about having both options, but then I discovered that the biggest bottleneck with xz compression is that it only works on one thread. Once I figured out how to split each file request and decompression into multiple threads then I discovered that the download speeds are limited per file and by having many requests at the same time its a huge increase in download speeds.

AFAIK from my AWS knowledge huggingface is fronted by cloudfront cdn and they have caching limitations past 5gb for speed to it would be signifigantly faster to split into many files for both decompression and network speeds

lrvl:
Any long-term solutions like using bittorrent ? @karpathy 

- **No Scripting**: Everything is done within your BitTorrent client
- Resilience: No single point of failure (like Hugging Face)
- Scalability: More downloads = faster speeds for everyone
- Cost-Effective: Reduces bandwidth costs for you as the model provider
- Community-Driven: Embraces the decentralized ethos
- DHT (Decentralized Goodness): BitTorrent clients can use DHT to find other peers even without a central tracker

karpathy:
@ChrisDryden  I noticed that fineweb100B here does not support parallel download. I can adjust the code ok I'm just curious why this was done? Is there any throttling or so?

ChrisDryden:
I just hadn't gotten around to updating it, I haven't come across any throttling so far

karpathy:
pushed a version of this code to master. closing.

Issue #504, opened 2024-05-31
Modal script - benchmarking, profiling and libraries


vyom1611:
Running the cuda code on modal using the benchmark_on_modal.py script is very useful, but I was wondering if there is a way to install cuDNN onto that app because we cannot run a lot of the faster kernels without libraries like such. Is it also possible to profile using nsight systems on modal?

Using modal is very convenient for people like me who do not have access to gpus, so if we can update the script and add more documentation on how to run profiling and all kinds of kernels on modal, I bet more folks would be able to access and run the codebase.

gordicaleksa:
I won't be able to help you with Modal, but I'll just say that our goal is ultimately to have custom CUDA kernels that outperform cuDNN. So you should, ideally, be able to just run our code without external dependencies.

One can dream. :)

awayzjj:
@vyom1611 Hi, I try to run the demo 
```
GPU_MEM=80 modal run benchmark_on_modal.py \
    --compile-command "nvcc -O3 --use_fast_math attention_forward.cu -o attention_forward -lcublas" \
    --run-command "./attention_forward 1
```
, but failed. Could you please give me some suggestions?  Did I miss some steps? Thank you very much!
<img width="1285" alt="截屏2024-06-16 13 31 55" src="https://github.com/karpathy/llm.c/assets/38181615/fb1616f9-550e-4e2b-95f0-ed2956beb00b">
<img width="1140" alt="截屏2024-06-16 13 32 11" src="https://github.com/karpathy/llm.c/assets/38181615/a8faa957-f201-4840-b2f2-34834c189f92">

vyom1611:
Hi, try running the compile command with the -lcublast option.

awayzjj:
Hi, I have already use the -lcublast option, you can see my logs above.

awayzjj:
@vyom1611 Thank you very much! It worked after adding the -lcublasLt option.
```
nvcc -O3 --use_fast_math attention_forward.cu -o attention_forward -lcublas -lcublasLt
```
Have you considered adding support for "ncu"? I tried, but encountered an error and no profile file was generated.
<img width="1207" alt="截屏2024-06-16 16 12 13" src="https://github.com/karpathy/llm.c/assets/38181615/89f10934-37ad-4a4b-8443-91011e02336c">

vyom1611:
[...]

PR #510, opened 2024-06-01
Benchmark modal script revamped - profiling and external libraries (Issue #504)

vyom1611:
Basically added support for libraries like cuDNN on modal so that people can run using libraries on cloud gpus. Basically added my own custom docker image for cuda environment setup and also added support for profiling using nsight systems which comes pre-installed. Based on issue #504

Features:
- cuDNN supported
- openMPI supported
- Nsight Systems supported and now by attaching a volume, you can download the report to view it on your local machine in GUI form
- instructions in the benchmark_on_modal.py script on how to use the volumes/download reports and existing errors.
-(Please not that profile_gpt2cu.py does not work with this and you have to manually change the command to run nsys profile)

Examples:
training the gpt2 model with cuDNN use:
GPU_MEM=80 modal run dev/cuda/benchmark_on_modal.py \
    --compile-command "make train_gpt2cu USE_CUDNN=1" \
    --run-command "./train_gpt2cu -i dev/data/tinyshakespeare/tiny_shakespeare_train.bin -j dev/data/tinyshakespeare/tiny_shakespeare_val.bin -v 250 -s 250 -g 144 -f shakespeare.log -b 4"

For profiling using nsight system:
GPU_MEM=80 modal run dev/cuda/benchmark_on_modal.py \
    --compile-command "make train_gpt2cu USE_CUDNN=1" \
    --run-command "nsys profile ./train_gpt2cu -i dev/data/tinyshakespeare/tiny_shakespeare_train.bin -j dev/data/tinyshakespeare/tiny_shakespeare_val.bin -v 250 -s 250 -g 144 -f shakespeare.log -b 4"

NOTE: Currently there is a bug in the profiling using nsight system which produces a lot of errors on the command line but it
does not actually interfere with the model training and validation. The report (that you download) is still generated and can be viewed from Nsight Systems. Additionally, 'profile_gpt2cu.py' does not work with this too.

Basically enables more features to be used on the cloud gpus. Please feel free to add-on/optimize this script so that more folks can explore the codebase and test out all kinds of features and profiling the codebase has to offer currently.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ CUDA_OUTPUT_FILE = -o $@
 
 # NVCC flags
 # -t=0 is short for --threads, 0 = number of CPUs on the machine
-NVCC_FLAGS = -O3 -t=0 --use_fast_math
+NVCC_FLAGS = -O3 -t=0 --use_fast_math -std=c++17
 NVCC_LDFLAGS = -lcublas -lcublasLt
 NVCC_INCLUDES =
 NVCC_LDLIBS =
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ CUDA_OUTPUT_FILE = -o $@
 # NVCC flags
 # -t=0 is short for --threads, 0 = number of CPUs on the machine
 NVCC_FLAGS = -O3 -t=0 --use_fast_math
+NVCC_FLAGS += --std=c++17
 NVCC_LDFLAGS = -lcublas -lcublasLt
 NVCC_INCLUDES =
 NVCC_LDLIBS =
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,6 @@ CUDA_OUTPUT_FILE = -o $@
 # NVCC flags
 # -t=0 is short for --threads, 0 = number of CPUs on the machine
 NVCC_FLAGS = -O3 -t=0 --use_fast_math
-NVCC_FLAGS += --std=c++17
 NVCC_LDFLAGS = -lcublas -lcublasLt
 NVCC_INCLUDES =
 NVCC_LDLIBS =
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,16 @@ NVCC_CUDNN =
 # By default we don't build with cudnn because it blows up compile time from a few seconds to ~minute
 USE_CUDNN ?= 0
 
+# We will place .o files in the `build` directory (create it if it doesn't exist)
+BUILD_DIR = build
+ifeq ($(OS), Windows_NT)
+  $(shell if not exist $(BUILD_DIR) mkdir $(BUILD_DIR))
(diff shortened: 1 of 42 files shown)


vyom1611:
I am still working on a newer base image for this so that the nsight errors disappear but meanwhile it should be working for most of it

Makefile
```diff
@@ -14,6 +14,7 @@ CUDA_OUTPUT_FILE = -o $@
 # NVCC flags
 # -t=0 is short for --threads, 0 = number of CPUs on the machine
 NVCC_FLAGS = -O3 -t=0 --use_fast_math
+NVCC_FLAGS += --std=c++17
```
karpathy, line 17: I'm not 100% sure what the repercussions would be here right now, is it needed for the modal script?

vyom1611: Currently yes, it prints out 100s of lines of errors in the terminal without that (I believe its because of the docker image I created having the newest version of cuda), but the current script will not work without that.

Screenshot 2024-06-04 at 9 30 24 AM

rosslwheeler: What version of GCC are you using?

vyom1611: As in on my container? I am unsure because I used a cuda image from docker hub as a base, but the current image I am using is this: 

[https://hub.docker.com/layers/totallyvyom/cuda-env/latest/images/sha256-a273f155a5ad9dcd4dd366a6da2f48f80f10b675bd136ab350f73b695d23b23b?tab=layers](url)

and this is the cuda image I am layering on:

[https://hub.docker.com/layers/nvidia/cuda/12.4.1-cudnn-devel-ubuntu20.04/images/sha256-f18cf1a9ac2842e59f13b0d0729594da8cbd68cadd2379308cdd98c0374dbd80?context=explore](url)

rosslwheeler: That helps - it's probably (I'm guessing) the ubuntu 20.04 compiler is too old. Try finding a container which has a newer version of ubuntu like 22.04. That should fix your compilation issues.

vyom1611: I am messing around with images for newer versions to also fix the errors I was getting with nsight, but it takes while because reloading all the containers after every little detail takes so much time, it is such a hassle to build a cuda dev env on a docker 😅

rosslwheeler: Yep, your right on that. Good luck!

vyom1611:
I just fixed the gcc versioning mistake, and cuDNN should also work with the modal script now.

vyom1611:
Since the -std=c++17 flag was added into Makefile in a previous commit, this should work perfectly now. 

- cudNN works ✅ 
- Nsys profile using "nsys profile --cuda-graph-trace=graph --python-backtrace=cuda --cuda-memory-usage=true  {executable} also works (although a GPU uuid error is generated, almost everything is profiled) ✅ 
- OpenMPI also installed ✅

PR #513, opened 2024-06-02
Added packed layernorm_forward

ChrisDryden:
This is the implementation of using packed data types for layernorm and has an associated speedup of around 50% when testing with the kernel, waiting for the PR for making the data types in that kernel into floatX to merge before merging the test kernels for this.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -663,9 +663,13 @@ __global__ void layernorm_forward_kernel3(floatX* __restrict__ out, floatX* __re
 
     // mean
     float sum = 0.0f;
-    for (int i = lane_id; i < C; i += WARP_SIZE) {
-        sum += (float)x[i];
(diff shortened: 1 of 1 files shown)


ChrisDryden:
https://github.com/karpathy/llm.c/pull/319 this one adds the floatX to the dev cuda kernel for this

ChrisDryden:
I am very embarrassed I swear running this yesterday I was getting numbers that were closer to 600GB/s for both kernel 6 and kernel 9 throughput and around 900GB/s for kernel 8 throughput and now when I run those kernels its  around 950GB/s and 1050GB/s for the 8. If that really is the case and you'd rather keep the simplicity and not have the packing then there is also the option of going for kernel 9 that is now the same speed as kernel 6 that will give the same performance as before but split it into two simpler kernels?

PR #517, opened 2024-06-02
add edu fineweb support, with 10B and 100B version

eliebak:
Adding edu-fineweb support https://huggingface.co/datasets/HuggingFaceFW/fineweb-edu

How to use it : 
for 10B : `python fineweb.py -t "edu" -v "10B"`
for 100B : `python fineweb.py -t "edu" -v "100B"`

And in the launching script: 
```
-i "dev/data/fineweb10B/edu_fineweb_train_*.bin" \
-j "dev/data/fineweb10B/edu_fineweb_val_*.bin" \
```

### DIFF
--- a/dev/data/fineweb.py
+++ b/dev/data/fineweb.py
@@ -28,27 +28,42 @@ import argparse
 from data_common import write_datafile
 # ------------------------------------------
 
-parser = argparse.ArgumentParser(description="FineWeb dataset preprocessing")
+parser = argparse.ArgumentParser(description="FineWeb and Edu-FineWeb dataset preprocessing")
+parser.add_argument("-t", "--type", type=str, default="classic", help="Edu or classic fineweb")
 parser.add_argument("-v", "--version", type=str, default="10B", help="Which version of fineweb to use 10B|100B")
 parser.add_argument("-s", "--shard_size", type=int, default=10**8, help="Size of each shard in tokens")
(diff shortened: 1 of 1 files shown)


karpathy:
Wow, this is amazing! I'll take a look today. So we probably want to hard switch to this, at our scale. If you're not training more than a few T epochs you'd always want to use this it looks like?

karpathy:
If the results in the huggingface page are true we might as well hard switch to that, forget if statements

eliebak:
Yes, i think the best thing is switching to fineweb-edu and also doing arc evals (or other reasoning task) instead of hellaswag. 

Screenshot 2024-06-02 at 19 18 50

karpathy:
Ah yes ok so we'll actually see a regression but it's because we don't measure properly :D
At some point you do have to decide on what eval domains you care about, and cater the pretraining mix to that.

eliebak:
Yes, that's why I think it's worth keeping the if statement even if it's a bit messy for now. When we have a proper way to evaluate the model, we can suppress them, right?

cometyang:
@eliebak, I tried to run train both 124M model on Fineweb and Finewebedu, the HellaSwag result does not improved while validation error get better.  Have you observed the same thing?
https://github.com/karpathy/llm.c/discussions/580#discussioncomment-9789323

PR #518, opened 2024-06-02
Generalize MFU logic to common non-A100 GPUs

gordicaleksa:
I've added support for some common GPUs.

I can refactor it a bit if you think that repeating this computation after every training step is unnecessary (which it is it's a tradeoff but it's a few cycles on the CPU).

We could instead call it once somewhere in the beginning of the main func and later just pass the promised flops into mfu func. I wanted to keep everything in the same place so made a tradeoff for readability.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -2857,10 +2857,14 @@ void gpt2_multi_gpu_gather(GPT2 *model, MultiGpuConfig* multi_gpu_config)
 #endif
 }
 
-float gpt2_estimate_mfu(GPT2 *model, int num_tokens, float dt) {
-    // estimate model flops utilization (MFU) in units of A100 bfloat16 peak FLOPS
+typedef struct {
+    std::string device_name;
+    float mfu;
+} GpuMFUInfo;
+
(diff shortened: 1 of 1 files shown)


ngc92:
have you  seen this?
https://github.com/karpathy/llm.c/pull/486
currently stalled, because getting the actual expected MFU is apparently quite nontrivial. E.g., are we looking at bf16 with fp32 accumulate or with bf16 accumulate (makes a difference for some cards).

gordicaleksa:
Oops, no i haven't! Thanks for flagging!

TBH given that our goal is to have everything in a single file i wouldn't try to be exhaustive here, let's just list the most common GPUs and let people tweak if they need it (similar to how GPT-2 hyperparams are hardcoded).

wdyt?

ngc92:
I don't think we'd actually put this into the main file in the end. The main intention of the other PR was to have some very basic performance data coded into the file, and then be able to extrapolate across a wide range of GPUs.

karpathy:
Yep we'll hide this somewhere, imminently I think. Sounds like we should close this one and merge @ngc92 PR https://github.com/karpathy/llm.c/pull/486

karpathy:
Merged an adjusted https://github.com/karpathy/llm.c/pull/486 just now, closing.

PR #519, opened 2024-06-02
Fix mem leak

gordicaleksa:
We didn't free up CPU buffer memory.

Also no need to use `NUM_ACTIVATION_TENSORS` size_t slots for bw_act_sizes, we just need `NUM_BACKWARD_TENSORS`.


karpathy:
🤦‍♂️  good catch(es)

PR #521, opened 2024-06-02
Replaced parameter B with the total parameters calculated outside of the kernel

ChrisDryden:
Similar to this CR but with some more opinionated variable name changes, since N is already taken, I renamed it to total_parameters, but open to any other options


karpathy:
Does this actually make difference? Maybe I'd prefer the more consistent naming scheme.

PR #522, opened 2024-06-02
Add master weights to resume state

gordicaleksa:
We're currently not saving master weights as part of the state -> we lose some precision because otherwise when we resume we'll have to reconstruct the master weights by upcasting from lower precision params.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -1188,6 +1188,8 @@ void save_state(const char* filename, int step, GPT2* model, DataLoader* loader)
     fwrite(cpu_buffer, sizeof(float), shard_num_parameters, state_file);
     cudaCheck(cudaMemcpy(cpu_buffer, model->v_memory, shard_num_parameters * sizeof(float), cudaMemcpyDeviceToHost));
     fwrite(cpu_buffer, sizeof(float), shard_num_parameters, state_file);
+    cudaCheck(cudaMemcpy(cpu_buffer, model->master_weights, shard_num_parameters * sizeof(float), cudaMemcpyDeviceToHost));
+    fwrite(cpu_buffer, sizeof(float), shard_num_parameters, state_file);
     free(cpu_buffer);
     fclose(state_file);
 }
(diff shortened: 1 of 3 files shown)


train_gpt2.cu
```diff
@@ -2954,6 +2954,8 @@ void save_state(const char* filename, int step, GPT2* model, DataLoader* loader)
     fwrite(cpu_buffer, sizeof(float), shard_num_parameters, state_file);
     cudaCheck(cudaMemcpy(cpu_buffer, model->v_memory, shard_num_parameters * sizeof(float), cudaMemcpyDeviceToHost));
     fwrite(cpu_buffer, sizeof(float), shard_num_parameters, state_file);
+    cudaCheck(cudaMemcpy(cpu_buffer, model->master_weights, shard_num_parameters * sizeof(float), cudaMemcpyDeviceToHost));
```
ngc92, line 2957: this probably needs a check to verify whether master weights are enabled

gordicaleksa: good catch (it was hardcoded everywhere to 1 but still :D)

ngc92: should be settable from command line, I think

train_gpt2.cu
```diff
@@ -2985,6 +2990,10 @@ void load_state(int* step, GPT2* model, DataLoader* loader, const char* filename
     cudaCheck(cudaMemcpy(model->m_memory, cpu_buffer, shard_num_parameters * sizeof(float), cudaMemcpyHostToDevice));
     freadCheck(cpu_buffer, sizeof(float), shard_num_parameters, state_file);
     cudaCheck(cudaMemcpy(model->v_memory, cpu_buffer, shard_num_parameters * sizeof(float), cudaMemcpyHostToDevice));
+    if (state_header[4] == 1) {  // if we used master weights during the state save
```
ngc92, line 2993: I think this might crash, because we might not have allocated the corresponding memory buffer yet

gordicaleksa: nice catch, fixed!

gordicaleksa:
* Tested for full determinism -> achieved!
* Profiled the run and it looks like not using atomic adds sped up things by a tiny bit! Main thing: no regression in speed here, we can later optimize further if needed.

2 steps were needed to achieve full determinism in both norm & loss:
* Fixed a race condition bug in the fused classifier kernel that was causing us to have infinities in our loss!
* Made sure we don't use atomic adds (due to non-associativity of floating point numbers).

How I tested it:
* Run 10 steps and save a checkpoint & state after 5 steps.
* Restart from the step 5 state and make sure loss & norm are precisely the same in both runs.
(Note I've also run for 20 steps to confirm longer horizons are also fully equivalent)

ngc92:
> How I tested it:
> 
>     * Run 10 steps and save a checkpoint & state after 5 steps.
> 
>     * Restart from the step 5 state and make sure loss & norm are precisely the same in both runs.
>       (Note I've also run for 20 steps to confirm longer horizons are also fully equivalent)

This sounds like it would be a good addition to our general CI testing

gordicaleksa:
Determinism working again even with the latest changes (strided global norm, overlap of comms & computation).

rosslwheeler:
> > How I tested it:
> > ```
> > * Run 10 steps and save a checkpoint & state after 5 steps.
> > 
> > * Restart from the step 5 state and make sure loss & norm are precisely the same in both runs.
> >   (Note I've also run for 20 steps to confirm longer horizons are also fully equivalent)
> > ```
> 
> This sounds like it would be a good addition to our general CI testing

@gordicaleksa - for CI can I run this now or do I need your changes in this PR to do the above?

gordicaleksa:
@rosslwheeler we need the PR merged first.

PR #523, opened 2024-06-02
Testability

ngc92:
Towards having our CI actually _test_ our code.
@rosslwheeler I just added the commands to run, I guess there is some more config needed to ensure this runs on GPU runners

### DIFF
--- a/test_gpt2.cu
+++ b/test_gpt2.cu
@@ -320,5 +320,5 @@ int main(int argc, char *argv[]) {
     free(expected_grads_memory);
     free(grads_memory_cpu);
     free(grads_memory_cpu_float);
-    return 0;
+    return allok ? EXIT_SUCCESS : EXIT_FAILURE;
 }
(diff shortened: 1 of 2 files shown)


rosslwheeler:
@ngc92 Will fix the above and try the run and let you know of any changes.

rosslwheeler:
@ngc92 fixed the tests and it all runs/passes with your changes.

karpathy:
This is the "actually run code" CI right? which is great. But it seems to still be draft and is failing, is there more intended functionality?

rosslwheeler:
@karpathy - yes, I have a separate PR for this which includes your suggestions as well. This will include the GPU so we'll need to set the runner up on your GitHub account. I'm just waiting on ubicloud to get your invite/account set up.

ngc92:
This PR actually enables testing the recomputation modes, and fixes an error with them.

PR #524, opened 2024-06-02
Fix zero stage 1 norm computation

gordicaleksa:
As per our Discord discussion:
Right now, in ZeRO stage 1, each GPU will compute a grad norm across all of its shards - but only a single shard has the correct numbers (after NCCL's reduce-scatter).

So what we should do instead is compute norm squared per shard that the device owns and then reduce (sum) them up across devices, take a sqrt and only then call AdamW kernel to update the params.


gordicaleksa:
WIP - need to convince myself this is correct and run tests :)

PR #526, opened 2024-06-03
attempt to move cudnn attention to the llmc directory

karpathy:


karpathy:
trying to hide the cudnn attention into llmc.
open for comments to anyone who maybe knows more Makefiles
cc @rosslwheeler i'm not 100% sure this will be ok Windows, waiting for CI

rosslwheeler:
@karpathy it passes. Will take a look tonight - verifying some inbound CI changes.

Makefile
```diff
@@ -22,6 +22,21 @@ NVCC_CUDNN =
 # By default we don't build with cudnn because it blows up compile time from a few seconds to ~minute
 USE_CUDNN ?= 0
 
+# We will place .o files in the `build` directory (create it if it doesn't exist)
```
rosslwheeler, line 25: Based on the below - please change this comment to remove (create it if it doesn't exist) - I don't think we need an explicit mkdir here.

Makefile
```diff
@@ -22,6 +22,21 @@ NVCC_CUDNN =
...
+BUILD_DIR = build
+ifeq ($(OS), Windows_NT)
```
rosslwheeler, line 27: Delete line 27 - line 38. I don't think we need to create the directory. I tested on Windows and Ubuntu 24. And the path separator for Windows can be / here in the Makefile. Tested on both OS'.

rosslwheeler: Optionally - you can remove line 26 as well and just replace BUILD_DIR with build in the two instances below.

rosslwheeler:
@karpathy - please let me know if you want me to PR this if that's easier. I know there's a few changes here.

rosslwheeler:
Likely a permissions difference. Thought that could be an issue.

PR #528, opened 2024-06-03
Refactor eval loader - remove 2 unnecessary malloc calls per iteration

gordicaleksa:
Stepped through the code and convinced myself we don't need to clear inputs/targets in every single iteration, BUT:

We do need to initialize inputs/targets with all zeros (using calloc) as otherwise we might end up with ints that are outside of the vocab and that would mess up the forward pass (initial embedding logic, i.e. `encoder_forward`).


gordicaleksa:
We can do some additional checks in `evalloader_init` similar to what we do in the training data loader (i.e. check that expected file size matches actual file size).

For that it might be easier to bump up version to 2 and add that info into the header during the file construction.

Alternatively we can do a full pass through the file once during the `evalloader_init` and fail fast if any issues. Right now if the file is corrupt we'll catch it worst case after having computed first 10k examples...

PR #531, opened 2024-06-03
Attention bug - WIP

gordicaleksa:
We might have a bug in the attention due to how we call `cublasGemmStridedBatchedEx`.

We compute K^T @ Q instead of Q @ K^T and that combined with causal masking might have introduced a bug.

The below implementation seems to be working, i kicked off a bigger run on my 8xH100 node, 124M params.

WIP. See [this discord thread](https://discord.com/channels/1189498204333543425/1227345713348870156/1247168948290457713).


gordicaleksa:
Closing this down as it turned out this has to do with the fact that Cublas uses column-major representation, there is no bug.

PR #535, opened 2024-06-03
moved device utilities into a separate file, and follow-up cleanup

ngc92:


rosslwheeler:
It appears that this commit may have broken the USE_CUDNN Windows build. Am fixing it. Turns out we don't test for this on Windows. I'm going to add this to CI as well as @ngc92 and @karpathy command line GPU suggestions.

Makefile
```diff
@@ -245,7 +245,7 @@ train_gpt2: train_gpt2.c
 test_gpt2: test_gpt2.c
 	$(CC) $(CFLAGS) $(INCLUDES) $(LDFLAGS) $^ $(LDLIBS) $(OUTPUT_FILE)
 
-$(NVCC_CUDNN): llmc/cudnn_att.cu
+$(NVCC_CUDNN): llmc/cudnn_att.cpp
```
rosslwheeler, line 248: @ngc92 - this broke the windows CUDNN build - not sure why - changing it back fixes it.

rosslwheeler: Fixed it - need this at the top of the source file: #define NOMINMAX

PR #538, opened 2024-06-03
Removing dependencies that are seemingly not needed to compile

ChrisDryden:
Mainly making this CR to test the integration suite to see if this compiles without all of the headers listed. Not sure why its still able to compile for me.


ChrisDryden:
Interesting builds failed on windows but passed on the others

PR #544, opened 2024-06-04
[layernorm.cuh] Minor fix with 32 replaced by WARP_SIZE

lancerts:
[layernorm.cuh] Minor fix with 32 replaced by WARP_SIZE


llmc/layernorm.cuh
```diff
@@ -172,7 +172,7 @@ __global__ void __launch_bounds__(512, 2) // todo - any warnings on Turing with
     int iterations_C = CEIL_DIV(C, C_per_iteration); // + 2;
 
     // the first half of shared memory is bias, second is weight
-    size_t rounded_C = CEIL_DIV(C, (32 * x128::size)) * (32 * x128::size);
+    size_t rounded_C = iterations_C * C_per_iteration;
```
ngc92, line 175: are you sure this optimization is safe?

PR #546, opened 2024-06-04
Fix periodic inference during training

kmyusk:
Problem: Enabling the generation block crashes the program 
Solution: Change the variable names for x and y so that it does not affect downstream code
Testing: Runs as expected with tinyshakespeare


kmyusk:
I think [this](https://github.com/karpathy/llm.c/blob/5379b1d9e9f6e9aae7369b73ec8e72ed477bda27/train_gpt2.py#L803-L804) is the problematic part. If the condition is not met, it is not overwritten.

PR #550, opened 2024-06-04
Re-introduce cuda streams

ngc92:
For now, this "just" adds a single stream, and moves as many operations there as possible
(everything except device->host memcpy, if I haven't missed anything)
to ensure that "local reasoning" is enough to understand any stream parallism and host-device asynchronicity, for now, every "main" function ends with a full device synchronization.
We may want to lift that restriction in the future, but we should be *very* careful about that.


ngc92:
closing this in favour of #552

PR #551, opened 2024-06-04
Fix PyTorch DDP loss bug

gordicaleksa:
In our C implementation with reduce the loss properly before displaying it using print0.

In our PyTorch losses were not reduced that caused us to think we have a divergence between implementations.

I've put it outside of the grad accumulation loop for efficiency reasons.


karpathy:
great find @gordicaleksa !

PR #556, opened 2024-06-05
Utilities for cuda streams + disk IO

ngc92:
handling disk io for checkpointing with cuda streams is a nontrivial task. If you're  not careful, you can easily get broken code (need to wait for data to be on the CPU before you can start writing the buffer to the disk), or synchronous behaviour (because memory is not page-locked, so async copies are not possible)
Therefore, this PR introduces two new utility functions to do the disk <-> device data transfer.

In addition to being less error prone and reducing code duplication, this also gives us a single point at which we can implement double buffering to actually get an overlap between device transfer and disk writes; with an added bonus that we no longer need to allocate giant CPU-side arrays (think: 8 A100s on the same node each wanting to write 40 GB model state; we'd attempt to allocate 320 GB of host memory there; maybe the boxes are big enough even under that scenario, but do we really want to do that?)

I've also added a unit test for these functions; its very rough, I am not going to touch the makefile so for now you need to compile yourself, and it leaves behind its temp file. But it already payed off, because my first implementation had a wrong offset somewhere, and the test caught this :)

There is also the problem notices first with #522 that we currently miss master weights in the saved state. This is hacked in here very quickly, but its not really a good solution; should be combined at least with #522's addition of a flag in the file that indicates whether to expect master weights or not.

Because we now do file IO with cuda, we get that `cuda_common.h` includes `utils.h`, and requires us to mark all functions there inline. I also had to add a checked write function, for which I've just copied the error handling; not 100% sure if that makes sense.

For the double-buffered transfers, I've put a size of 32MiB, but that is not based on any actual data. I just wanted to put something there that is neither super tiny nor super big :)

### DIFF
--- a/dev/test/device_file_io.cu
+++ b/dev/test/device_file_io.cu
@@ -9,9 +9,10 @@ nvcc -o device_file_io device_file_io.cu && ./device_file_io
 #include "../../llmc/cuda_common.h"
 #include <vector>
 #include <random>
+#include <cstdio>
 #include <algorithm>
 
-void test(size_t nelem, size_t wt_buf, size_t rd_buf) {
+void test(size_t nelem, size_t wt_buf_size, size_t rd_buf_size) {
 
     float* data;
     cudaCheck(cudaMalloc(&data, nelem*sizeof(float)));
@@ -28,7 +29,7 @@ void test(size_t nelem, size_t wt_buf, size_t rd_buf) {
     cudaStreamCreate(&stream);
 
     FILE* tmp = fopenCheck("tmp.bin", "w");
-    device_to_file(tmp, data, nelem * sizeof(float), wt_buf, stream);
+    device_to_file(tmp, data, nelem * sizeof(float), wt_buf_size, stream);
     fcloseCheck(tmp);
 
 
@@ -36,20 +37,22 @@ void test(size_t nelem, size_t wt_buf, size_t rd_buf) {
     cudaCheck(cudaMalloc(&reload, nelem*sizeof(float)));
 
     tmp  = fopenCheck("tmp.bin", "r");
-    file_to_device(reload, tmp, nelem * sizeof(float), rd_buf, stream);
+    file_to_device(reload, tmp, nelem * sizeof(float), rd_buf_size, stream);
     fcloseCheck(tmp);
 
     std::vector<float> cmp(nelem);
     cudaCheck(cudaMemcpy(cmp.data(), reload, nelem * sizeof(float), cudaMemcpyDeviceToHost));
     for(int i = 0; i < nelem; ++i) {
-        if(random_data[i]  != cmp[i])  {
+        if(random_data[i] != cmp[i])  {
             fprintf(stderr, "FAIL: Mismatch at position %d: %f vs %f\n", i, random_data[i], cmp[i]);
+            remove("tmp.bin");
             exit(EXIT_FAILURE);
         }
     }
 
     cudaCheck(cudaFree(reload));
     cudaCheck(cudaFree(data));
+    remove("tmp.bin");
 }
 
 int main() {
--- a/dev/test/device_file_io.cu
+++ b/dev/test/device_file_io.cu
@@ -1,4 +1,12 @@
-#include "llmc/cuda_common.h"
+/*
+Tests device <-> file IO functions
+
+compile and run as (from dev/test directory)
+nvcc -o device_file_io device_file_io.cu && ./device_file_io
+*/
+
+
+#include "../../llmc/cuda_common.h"
 #include <vector>
 #include <random>
 #include <algorithm>
--- a/llmc/cuda_common.h
+++ b/llmc/cuda_common.h
@@ -194,6 +194,7 @@ inline void file_to_device(void* dest, FILE* src, size_t num_bytes, size_t buffe
 
     // copy the last remaining write buffer to gpu
     cudaCheck(cudaMemcpyAsync(gpu_write_ptr, write_buffer, write_buffer_size, cudaMemcpyHostToDevice, stream));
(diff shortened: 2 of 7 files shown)


dev/test/device_file_io.cu
```diff
@@ -0,0 +1,61 @@
...
+    std::mt19937 rng(42);
+    std::uniform_real_distribution<float> dist(-100.f, 100.f);
+    std::generate(random_data.begin(), random_data.end(), [&](){ return dist(rng); });
+
+    cudaCheck(cudaMemcpy(data, random_data.data(), random_data.size()*sizeof(float), cudaMemcpyHostToDevice));
+
+    cudaStream_t stream;
+    cudaStreamCreate(&stream);
+
+    FILE* tmp = fopenCheck("tmp.bin", "w");
```
gordicaleksa, line 30: let's remove the file when the test ends in case we want to integrate this into a CI?

not sure what runner machines do with storage (i.e. is it persistent or volatile) - but still a good practice

dev/test/device_file_io.cu
```diff
@@ -0,0 +1,61 @@
...
+nvcc -o device_file_io device_file_io.cu && ./device_file_io
+*/
+
+
+#include "../../llmc/cuda_common.h"
+#include <vector>
+#include <random>
+#include <algorithm>
+
+void test(size_t nelem, size_t wt_buf, size_t rd_buf) {
```
gordicaleksa, line 14: nit: got confused with what these are until i analyzed `device_to_file` function and realized these are sizes
maybe rename to `wt_buf_size` and `rd_buf_size`?

llmc/cuda_common.h
```diff
@@ -116,4 +118,83 @@ class NvtxRange {
...
+    char* buffer_space;
+    cudaCheck(cudaMallocHost(&buffer_space, 2*buffer_size));
+    // split allocation in two
+    void* read_buffer = buffer_space;
+    void* write_buffer = buffer_space + buffer_size;
+
+    // prime the read buffer; first copy means we have to wait
+    char* gpu_read_ptr = (char*)src;
+    size_t copy_amount = std::min(buffer_size, num_bytes);
+    cudaCheck(cudaMemcpyAsync(read_buffer, gpu_read_ptr, copy_amount, cudaMemcpyDeviceToHost, stream));
```
gordicaleksa, line 133: curious: does async matter here given that we call synchronize immediately on the next line?

llmc/cuda_common.h
```diff
@@ -116,4 +118,83 @@ class NvtxRange {
...
+        cudaCheck(cudaStreamSynchronize(stream));     // wait for both buffers to be ready.
+
+        std::swap(read_buffer, write_buffer);
+        rest_bytes -= copy_amount;
+        write_buffer_size = copy_amount;
+        gpu_read_ptr += copy_amount;
+    }
+
+    // make sure to write the last remaining write buffer
+    fwriteCheck(write_buffer, 1, write_buffer_size, dest);
```
gordicaleksa, line 156: is this (implicitly) synchronized?

llmc/cuda_common.h
```diff
@@ -116,4 +118,83 @@ class NvtxRange {
...
+        cudaCheck(cudaStreamSynchronize(stream));     // wait for both buffers to be ready.
+
+        std::swap(read_buffer, write_buffer);
+        rest_bytes -= copy_amount;
+        write_buffer_size = copy_amount;
+    }
+
+    // copy the last remaining write buffer to gpu
+    cudaCheck(cudaMemcpyAsync(gpu_write_ptr, write_buffer, write_buffer_size, cudaMemcpyHostToDevice, stream));
+    cudaCheck(cudaFreeHost(buffer_space));
```
gordicaleksa, line 197: cudaFreeHost is blocked until the line above finishes?

train_gpt2.cu
```diff
@@ -1231,20 +1222,24 @@ void load_state(int* step, GPT2* model, DataLoader* loader, const char* filename
         printf0("allocating %zu MiB for AdamW optimizer state v\n", (shard_num_parameters * sizeof(float)) >> 20);
         cudaCheck(cudaMalloc((void**)&model->v_memory, shard_num_parameters * sizeof(float)));
     }
+
+    if(state_header[4] == 1 && !model->use_master_weights) {
```
gordicaleksa, line 1226: this will have to be refactored a bit due to recent changes

gordicaleksa:
Left some comments - lgtm!

llmc/cuda_common.h
ngc92: good question. Allocation is listed as an implicit sync, I would  assume deallocation also needs to be, but I'm not 100% sure, so maybe we should be explicit here.
https://docs.nvidia.com/cuda/cuda-c-programming-guide/#implicit-synchronization

llmc/cuda_common.h
ngc92: no,  its just for consistency; since non-async versions of this function don't take a stream argument, if you want this transfer to show up in the "right" stream in nsight, you need to use the async version.

dev/test/device_file_io.cu
rosslwheeler: You can assume on the runners it's not persistent.

ngc92: we might still run multiple test cases in one container, so I think cleaning up is still the right choice

dev/test/device_file_io.cu
ngc92: yes, good idea

dev/test/device_file_io.cu
rosslwheeler: Yes, good point - was just commenting on the runners. Always cleaning up is a good strategy regardless of persistence. You never know where the code will be used next. Especially, if you run it locally...

PR #558, opened 2024-06-06
Fix issue 555, support c++11 and fix compiler warnings

### LINKED ISSUE #555: apparent compatibility issues with earlier c++ versions after recent pushes
So I had an earlier version of the code which would compile 'train_gpt2cu' just fine on my system, with
```
nvidia driver v 545.23.08
CUDA version 12.3
gcc 9.4.0
Ubuntu 20.04
```
But the recent updates of the code, specifically the one that added the .cuh files (matmul.cuh) made the code incompatible with my system and compiling it results in errors:
```
llmc/matmul.cuh(16): error: namespace "std" has no member "bool_constant"
                                               std::bool_constant<UseAuxBuffer>) {
```
and other places.
I did some search, apparently this is a c++ version compatibility issue and setting the flag -std=c++17 would resolve the issue as this `bool_constant` was introduced in c++ 17. I was wondering if it is possible to avoid this? Many of us don't have the exact same day updates of these libraries and it would cause a lot of other hardware compatibility issues for everyone to update their drivers and compiler versions. So I was wondering if it is possible to avoid them if they are not necessary or add macros to still support the earlier compiler and cuda versions.

After all, as @karpathy mentioned, this project also has educational purposes and for someone like me it is getting a little hard to follow as well. But anyways, thank you everyone for your great work and teachings.

lancerts:
Fix #555. 

- fix compiler error (compiler dependent) of following type: 
```
matmul_backward_bias.cu(151): error: no operator "+=" matches these operands
            operand types are: floatX += float
          dbias[idx] += block_sum;
```

- support c++11, replace the c++17 only features with c++11 equivalence 
          - fix `error: namespace "std" has no member "bool_constant"`, 
          - replace `std::is_same_v`
          - handle  `constexpr if statements` with function overload since `constexpr if statements are a C++17 feature`. No runtime speed impact since the branch is still determined at the compile time. But it is a bit verbose without  `if constexpr`.

Tested with `nvcc -O3 -lcublas -lcublasLt -std=c++11 matmul_backward_bias.cu -lineinfo -o matmul_backward_bias`
and executes kernels with changes (all passed).

### DIFF
--- a/dev/cuda/matmul_backward_bias.cu
+++ b/dev/cuda/matmul_backward_bias.cu
@@ -116,7 +116,7 @@ __global__ void matmul_backward_bias_kernel2(floatX* dbias, const floatX* dout, 
     sum = cg::reduce(warp, sum, cg::plus<float>{});
     // write the result to output (global memory)
     if(warp.thread_rank() == 0) {
-        dbias[idx] += sum;
+        dbias[idx] = (float)dbias[idx] + sum;
     }
 }
 
@@ -148,7 +148,7 @@ __global__ void matmul_backward_bias_kernel3(floatX* dbias, const floatX* dout, 
     float block_sum = cg::reduce(warp, warp_sum, cg::plus<float>{}); // sum(x)
     // write the result to output (global memory)
     if(threadIdx.x == 0) {
-        dbias[idx] += block_sum;
+        dbias[idx] = (float)dbias[idx] + block_sum;
     }
 }
 
@@ -188,7 +188,7 @@ __global__ void matmul_backward_bias_kernel4(floatX* dbias, const floatX* dout, 
         for (int j = 0; j < vstep; j++) {
             dout_sum += smem[lane_id + j * warpSize];
         }
-        dbias[tl + lane_id] += dout_sum;
+        dbias[tl + lane_id] = (float)dbias[tl + lane_id] + dout_sum;
     }
 }
 
--- a/dev/cuda/matmul_backward_bias.cu
+++ b/dev/cuda/matmul_backward_bias.cu
@@ -272,9 +272,19 @@ __global__ void matmul_backward_bias_kernel7(float* dbias, const floatX* dout, i
 // threads to fetch a cacheline, which means that we can already operate on a "depth" of four within a single warp.
 // => blockDim.x == 4, blockDim.y == 32/4 = 8
 //
+template<typename OutFloat>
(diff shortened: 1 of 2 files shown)


abhilash1910 (approved):
@lancerts  These changes look great for C++11 back portabilitty .
But is there a genuine reason why C++17 standards should not be used ?

lancerts:
@abhilash1910  C+17 standard itself is cool. This is mainly for wider portability. (If we can achieve the same goal with less fancier stuff, maybe we just do it this way.)

ngc92:
that is a great argument why we shouldn't use C++ 20 (yet); but restricting to C++11 doesn't really make sense;
to put things into perspective, python3 was released in  2008, only three years earlier; the GPUs that came out in 2011 where the GTX 500 series; the last version of cuda to support this hardware was CUDA 8.

what we've been really lacking so far is verification that our code work on _reasonably_ old systems. Hopefully, with #559  we'll have less of these surprises, as that compiles the code on ubuntu 20.04, which is currently the oldest non EOL ubuntu LTS version.

That being said, I believe the += problem fixed in this PR happens still with cuda 11.8, which is not that old, and we should support it.

lancerts:
> that is a great argument why we shouldn't use C++ 20 (yet); but restricting to C++11 doesn't really make sense; to put things into perspective, python3 was released in 2008, only three years earlier; the GPUs that came out in 2011 where the GTX 500 series; the last version of cuda to support this hardware was CUDA 8.
> 
> what we've been really lacking so far is verification that our code work on _reasonably_ old systems. Hopefully, with #559 we'll have less of these surprises, as that compiles the code on ubuntu 20.04, which is currently the oldest non EOL ubuntu LTS version.
> 
> That being said, I believe the += problem fixed in this PR happens still with cuda 11.8, which is not that old, and we should support it.

+1 on c+17 is a nice standard to adopt here. 
Wrt c+17:
- Option 1, no change, just use c++17 for the whole repo since the compiler flag is now specified in training.
- Option 2, no change on the llmc/matmul.cuh, the training adopts c++17. Make dev/cuda/matmul_backward_bias.cu with the change here to be compatible with c+11. [In case someone plays with the dev but has c+11 only?]

Wrt `+= problem`
Yes, the += problem happens to me on cuda 12.1 but not on cuda 12.4 so probably converted a PR to just fix those (pending option 1vs2 above).

PR #559, opened 2024-06-06
set compile flag and add ci check

ngc92:


rosslwheeler:
@ngc92 - do you want to make the same change for the cuda Makefile?

PR #561, opened 2024-06-06
Fix the compiler warnings and errors

lancerts:
Fix such error that happens for CUDA 11.8 per discussion https://github.com/karpathy/llm.c/pull/558#issuecomment-2153054126
```
matmul_backward_bias.cu(151): error: no operator "+=" matches these operands
            operand types are: floatX += float
          dbias[idx] += block_sum;
```


dagelf:
```
dev/cuda/common.h:164:typedef float floatX;
```
LGTM

PR #564, opened 2024-06-07
layernorm_forward kernel#1 with copy compute overlap is 3x faster than kernel#5

josh-ramer:
I added a layernorm_forward kernel, #6, which just applies the copy compute overlap to kernel#1. It turns out that these kernels were heavily memory bound & dominated by the host to device & device to host memory copies. I modified the benchmarking code to include memory copy operations. Applying the copy compute overlap to any of the other kernels is as easy as modifying kernel#6, just change the kernel number in the kernel launch line. Here is the script that generated the data used in the [writeup](https://external-brain.github.io/palace-window/20240531-111932-llmc#copy-compute-overlap).
```bash
for b in {8,32,128,512}
do
     ./layernorm_forward 5 $b
     ./layernorm_forward 6 $b
done
```
I've prepared a [little overview of the work with profiles, graphs, & the raw data](https://external-brain.github.io/palace-window/20240531-111932-llmc#copy-compute-overlap).

### DIFF
--- a/dev/cuda/common.h
+++ b/dev/cuda/common.h
@@ -1,3 +1,4 @@
+#include <driver_types.h>
 #include <stdlib.h>
 #include <stdio.h>
 #include <cuda_runtime.h>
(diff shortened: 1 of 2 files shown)


josh-ramer:
Working on making a clang format file that makes the diff less dramatic. I couldn't find any standards when I search the repo, but I'll at least match the indents & use snake case (if that's what's generally used).

rosslwheeler:
There was this PR: https://github.com/karpathy/llm.c/pull/270

josh-ramer:
Thanks Ross!

ngc92:
unfortunately, the diff is currently not really readably, so I might have just misunderstood something, but this seems to optimize for a non-existent use-case?

For any actual computations, all the kernels inputs and outputs will be buffers in GPU memory; no memcpy operation will be involved. As such, it is deliberate that the benchmarking code excludes any memcopies.

Btw, even with all the data in GPu memory, the kernels are still heavily memory bound regarding GPU-RAM to SM memory transfers ;)

josh-ramer:
Hah, that makes sense, I wondered if that might be the case. Do you have anything similar to point me at? If not similar anything in general you want help with?

PR #565, opened 2024-06-07
Consolidate memory

ngc92:
Two steps:
First, we make the backward kernel for the causal softmax inplace. This comes at a tiny cost, because we now need to write out explicitly the zeros to the masked locations, but allows non-cudnn builds to avoid the T² tensor here. Since we'd expect non-cudnn builds more for the home-style application (if you go to the trouble of securing an A100 or H100, you should really use CUDNN to actually make use of that compute power), where memory is very scarce, I believe we should just have this as the general solution, not offering any flags or ifdefs.

With this change done, we can do further simplifications to our memory management. Backward activations shrink to (unconditionally) two small tensors, so we can just add them to the general Activation struct  and get rid of all the additional management code needed for backward activations. Also, we get one less cudaMalloc call.

### DIFF
--- a/llmc/attention.cuh
+++ b/llmc/attention.cuh
@@ -149,8 +149,8 @@ __global__ void softmax_forward_kernel5(floatX* out, float inv_temperature, cons
     }
 }
 
-__global__ void softmax_autoregressive_backward_kernel(floatX* dpreatt, const floatX* datt, const floatX* att,
-                                                       int B, int T, int C, float scale) {
+__global__ void softmax_autoregressive_backward_inplace_kernel(floatX* datt, const floatX* att,
+                                                               int B, int T, int C, float scale) {
     constexpr const int BlockSize = 256;
     constexpr int T_per_block = 4;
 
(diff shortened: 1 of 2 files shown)


karpathy:
Nice! 

I guess one downside here is that we'll pay the memory cost of backward (4BTC + BTC) even if we never call backward, which we didn't have previously? But I guess it's nice to consolidate here and delete a bunch of LOC.

This PR will probably need some mild updates also once we get the `main_stream` PR into master.

ngc92:
> I guess one downside here is that we'll pay the memory cost of backward (4BTC + BTC) even if we never call backward, which we didn't have previously?

Not really. The memory cost we pay for backward is that we need to keep around 10s of GBs of activations in the forward pass, which wouldn't be needed for inference. Compared to that, the additional buffers here are maybe a 100MiB, basically negligible.

karpathy:
Yep totally, I meant in a strictly pedantic way. Ok.

PR #568, opened 2024-06-07
Improve logger, add grad norm + learning rate

gordicaleksa:
Other than loss, gradient norm and learning rate tell a lot of information about the training dynamics:
1) modified the logger to support logging these 2 quantities
2) updated jupyter notebook parser


karpathy:
agree!

PR #573, opened 2024-06-07
Dataloader - introducing randomness

gordicaleksa:
On the way to fully random train data shuffling...

This PR does the following:
1) Each process has a different unique random seed
2) Each process train data loader independently chooses its starting shard and starting example within that shard
3) This should not impact save/load state since we store the `rng_state` inside of the header and later load it

For handling multiple epochs:
1) We can explicitly call `dataloader_reset` on each epoch boundary inside the training loop -> that would enable us new random shard/position for each process.
2) We could alternatively delegate this responsibility to the data loader, which might be a bit nicer - we can keep the start shard index (and start position) and when we hit the starting position again we call `dataloader_reset` automatically.

Regarding "page" shuffling:
1) We could create a permutation of shard indices and iterative over it. That would be a partial solution.
2) In order to do intra-shard shuffling we'd need to extract the `eos` guarded tokens (i.e. unique docs) and randomly permute them and store them to the file system (and then override on each epoch so we upper bound mem footprint with 2x the original size) or for smaller datasets maybe keep the permutation in RAM.

Just some quick thoughts.

Looking into how we can integrate this over multiple PRs as opposed to all in a single go.

Let me know what you think!

### DIFF
--- a/llmc/dataloader.h
+++ b/llmc/dataloader.h
@@ -92,7 +92,7 @@ int64_t dataloader_load_shard_(DataLoader *loader, int shard_index) {
         exit(EXIT_FAILURE);
     }
     // -1 uint16_t due to us taking B*T+1 tokens but moving by B*T tokens
-    loader->shard_num_samples = (loader->file_size_bytes - sizeof(uint16_t)) / loader->total_batch_size_bytes;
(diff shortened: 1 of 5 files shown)


gordicaleksa:
@karpathy we can close this one now

PR #575, opened 2024-06-10
rms norm forward kernel

AndreSlavescu:


dev/cuda/common.h
```diff
@@ -18,6 +18,25 @@ __device__ float warpReduceSum(float val) {
     return val;
 }
 
+// modified from: https://developer.nvidia.com/blog/faster-parallel-reductions-kepler/
+__inline__ __device__ float blockAllReduceSum(float val) {
```
karpathy, line 22: i think we already have this btw? see cuda_utils.cuh

karpathy:
happy to get RMSNorm in!

dev/cuda/common.h
AndreSlavescu: Yeah I found it there as well, just wasnt sure if it was a good idea to add it there too given that there is a warpReduceSum in both cuda_utils.h and common. I'll remove it.

AndreSlavescu:
> happy to get RMSNorm in!



> happy to get RMSNorm in!

Sounds good! I will also make a PR for RoPE as mentioned in the discord to get started on llama3 integration.

PR #578, opened 2024-06-12
Remove cublaslt from fp32cu versions

ngc92:
adds a reasonably efficient matrix  multiplication in plain cuda for use instead of the cublaslt implementation.
With tensor cores enabled, this is a _huge_ loss of compute performance, obviously.
Without tensorcores, we still are about 30% or so slower than cublaslt, so there is a pretty significant impact this has on speed. But for the "pure" fp32 version, maybe that is acceptable?

Interestingly, in terms of LOC, this almost breaks even; cublaslt has so much boilerplate that we basically can fit in the entire custom kernel in the same amount of lines!

### DIFF
--- a/dev/cuda/matmul_forward.cu
+++ b/dev/cuda/matmul_forward.cu
@@ -84,7 +84,8 @@ __global__ void add_bias(float* out, const float* bias, int B, int T, int OC) {
     }
 }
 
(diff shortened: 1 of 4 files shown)


karpathy:
Love it!

PR #581, opened 2024-06-12
Add link for SYCL runtime

abhilash1910:
From https://github.com/karpathy/llm.c/discussions/572
cc @karpathy thanks


abhilash1910:
The dev folder contains older kernels, which will be supported eventually. Currently it supports the prime kernels required for end to end runs and benchmarks - train gpt2 /32 which includes headers under llmc. This includes cudnn kernels as well. 
I guess it would take a few days to see the upgraded deb folder in the repository as currently focus is on public end to end benchmarks.

PR #583, opened 2024-06-12
Replaced hard-coded max float with FLT_MAX

vyom1611:
Just fixed some TODOs and replaced hard-coded values with FLT_MAX for floating point integers which comes with already included <float.h> headers in both files.


Munna5553 (commented):
👍

karpathy:
ty!

PR #585, opened 2024-06-13
Fix MFU printing

gordicaleksa:
We have a bug when the device we're running on is not supported: we print "-100%". Fixed it so we print "n/a" in that case.

Also added support for H100 PCIe which is the device I have.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -1672,7 +1672,7 @@ int main(int argc, char *argv[]) {
     }
 
     // train
-    char* mfu_str = (char*)malloc(16);
+    char mfu_str[16];
     cudaEvent_t start, end;
     cudaCheck(cudaEventCreate(&start));
     cudaCheck(cudaEventCreate(&end));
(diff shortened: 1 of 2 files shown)


train_gpt2.cu
```diff
@@ -1671,6 +1672,7 @@ int main(int argc, char *argv[]) {
     }
 
     // train
+    char* mfu_str = (char*)malloc(16);
```
karpathy, line 1675: I'm a bit nervous about introducing this, hardcoding 16 here, having to free it, etc.

gordicaleksa: I agree there is a slight overhead but this is done once and left. Plus even if we had a leak it's 16 bytes :D

gordicaleksa: Lmk if you have a better proposal! I still think in net it's better than having -100% for non-supported GPUs.

ngc92: can't we just put these 16 bytes on the stack?

gordicaleksa: definitely - that's cleaner!

karpathy:
char* makes me nervous. i'm not actually sure if the -100% is terrible... or we can make it 0% if it looks less scary.

rosslwheeler:
One method would be to define a constant MAX_STRING_SIZE - then you would set mfu_str = malloc (MAX_STRING_SIZE) and use it as the second param in the snprintf statement - snprintf(mfu_str, MAX_STRING_SIZE, "n/a");

You would need to free it too when done.

If you don't malloc it - you would still need the constant for the definition mfu_str[MAX_STRING_SIZE] in order to use it elsewhere in the file when using snprintf(mfu_str, MAX_STRING_SIZE, "n/a");. sizeof() only works correctly in the same function where the mfu_str char array is declared. I would use different names too. Using the same name makes it a bit confusing to me.

PR #586, opened 2024-06-13
Add a debugging tip to README

gordicaleksa:
For folks new to the C world this tip can be helpful.


ngc92:
there is also `-Og` that optimizes a little bit, but in a way that is supposed to work well with debugging (`-g` and `-Ox` are not mutually exclusive options)
if we are going to add such a section, `CUDA_LAUNCH_BLOCKING` might also be something to tell people about.

rosslwheeler:
Also, I've had to disable Cuda debugging at times to get things to work. You will get the "too many resources" error. Be a helpful tip to add.

karpathy:
Very happy to add these tips! Will merge this one for now but happy to add anything else related too.

PR #593, opened 2024-06-14
Zero 2 - WIP

ngc92:
Trying to get a first version working. Code isn't nice, we currently lose the asynchrony in the communication code because we need to reuse the buffer for the next layer, and it doesn't give correct results.

### DIFF
--- a/llmc/cuda_utils.cuh
+++ b/llmc/cuda_utils.cuh
@@ -106,7 +106,15 @@ __global__ void copy_and_cast_kernel(Td* dst, const Ts* src, size_t n, ptrdiff_t
     int idx = blockIdx.x * blockDim.x + threadIdx.x;
     // need to try grid stride looping for more perf later
     if (idx < n) {
-        dst[idx + stride_dst * blockIdx.y] = cast_value<Td, Ts>(src[idx + stride_src * blockIdx.y]);
+        dst[idx + stride_dst * blockIdx.y] = cast_value<Td>(src[idx + stride_src * blockIdx.y]);
+    }
+}
+
+template<class Td, class Ts>
+__global__ void vector_add(Td* dst, const Ts* src, size_t n) {
+    ptrdiff_t idx = (ptrdiff_t)blockIdx.x * blockDim.x + threadIdx.x;
+    if (idx < n) {
+        dst[idx] = (Td)((float)dst[idx] + (float)src[idx]);
     }
 }
 
--- a/llmc/zero.cuh
+++ b/llmc/zero.cuh
@@ -212,7 +212,7 @@ void multi_gpu_async_reduce_gradient(
                 ncclFloatX, ncclAvg,
                 multi_gpu_config->nccl_comm, multi_gpu_config->nccl_stream
             ));
-        } else if(multi_gpu_config->zero_stage == 1) {
+        } else if(multi_gpu_config->zero_stage == 1 || multi_gpu_config->zero_stage == 2) {
             assert(pointers_sizes[i] % multi_gpu_config->num_processes == 0);
             size_t shard_size = pointers_sizes[i] / multi_gpu_config->num_processes;
             ptrdiff_t shard_offset = (ptrdiff_t)shard_size * multi_gpu_config->process_rank;
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -89,14 +89,14 @@ void set_zero_configs(MultiGpuConfig* multi_gpu_config, int zero_stage, size_t t
     if (zero_stage == 0) {
         printf0("| Zero Optimization is disabled                                              |\n");
     }
-    else if (zero_stage == 1) {
+    else if (zero_stage == 1 || zero_stage == 2) {
         if (total_parameters % multi_gpu_config->num_processes != 0) {
             printf0("| Zero Optimization is disabled, Can't equally partition parameters          |\n");
             multi_gpu_config->zero_stage = 0;
         }
         else {
-            printf0("| Zero Stage1 is enabled                                                     |\n");
-            multi_gpu_config->zero_stage = 1;
+            printf0("| Zero Stage %d is enabled                                                    |\n", zero_stage);
+            multi_gpu_config->zero_stage = zero_stage;
             multi_gpu_config->shard_num_parameters = total_parameters / multi_gpu_config->num_processes;
         }
     }
@@ -308,7 +308,11 @@ typedef struct {
     size_t num_parameters_bytes;
     // gradients of the weights
     ParameterTensors grads;
+    size_t grads_bytes;
+    ParameterTensors grad_shards;   // ZeRO-2 gradient shards
+    size_t grad_shards_bytes;
     void* grads_memory;
+    void* grad_shards_memory;
     // buffers for the AdamW optimizer
     float* m_memory;
     float* v_memory;
@@ -352,6 +356,7 @@ void gpt2_init_common(GPT2 *model) {
     model->mean_loss = -1.0f; // -1.0f designates no loss, set at end of forward()
     // memory lazily initialized in backward()
     model->grads_memory = NULL;
+    model->grad_shards_memory = NULL;
     model->workload_indices = NULL; // on cpu, for encoder_backward
     model->bucket_info = NULL; // on cpu, for encoder_backward
     // memory lazily initialized in update()
@@ -716,7 +721,10 @@ void gpt2_forward(GPT2 *model, const int* inputs, const int* targets, size_t B, 
 void gpt2_zero_grad(GPT2 *model) {
     NVTX_RANGE_FN();
     if (model->grads_memory != NULL) {
-        cudaCheck(cudaMemset(model->grads_memory, 0, model->num_parameters * sizeof(floatX)));
+        cudaCheck(cudaMemset(model->grads_memory, 0, model->grads_bytes));
+    }
+    if(model->grad_shards_memory != NULL) {
+        cudaCheck(cudaMemset(model->grad_shards_memory, 0, model->grad_shards_bytes));
     }
     cudaCheck(cudaDeviceSynchronize());
 }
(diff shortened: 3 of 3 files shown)


llmc/cuda_utils.cuh
```diff
@@ -106,7 +106,15 @@ __global__ void copy_and_cast_kernel(Td* dst, const Ts* src, size_t n, ptrdiff_t
     int idx = blockIdx.x * blockDim.x + threadIdx.x;
     // need to try grid stride looping for more perf later
     if (idx < n) {
-        dst[idx + stride_dst * blockIdx.y] = cast_value<Td, Ts>(src[idx + stride_src * blockIdx.y]);
+        dst[idx + stride_dst * blockIdx.y] = cast_value<Td>(src[idx + stride_src * blockIdx.y]);
```
gordicaleksa, line 109: we don't need to be explicit with Ts here or why remove this, since I see that the cast_value is "templatized" with both Td and Ts

ngc92: yes, Ts is uniquely determined by the function arguments; whereas Td cannot be inferred, since it only appears in the return type

train_gpt2.cu
```diff
@@ -733,8 +741,40 @@ void gpt2_backward(GPT2 *model, int* inputs, bool last_step) {
...
-        printf0("allocating %d MiB for parameter gradients\n", (int)round(model->num_parameters * sizeof(floatX) / (1024 * 1024)));
-        model->grads_memory = malloc_and_point_parameters(&model->grads, model->param_elements, model->param_sizeof);
+        if(multi_gpu_config.zero_stage == 2) {
+            // Allocate parameter buffers for the current layers active "wave" of computation
+            size_t param_elements[NUM_PARAMETER_TENSORS];
+            size_t param_sizeof[NUM_PARAMETER_TENSORS];
+            GPT2Config wave_config = model->config;
+            // allocate space for two layers, so we can do double-buffering
+            wave_config.num_layers = 2;
+            fill_in_parameter_sizes(param_elements, param_sizeof, wave_config);
```
gordicaleksa, line 751: curious whether DeepSpeed shards also embedding layers?

train_gpt2.cu
```diff
@@ -733,8 +741,40 @@ void gpt2_backward(GPT2 *model, int* inputs, bool last_step) {
...
+            }
+            printf0("allocating %d MiB for ZeRO-2 active gradients\n",
+                    (int) round(alloc_bytes / (1024 * 1024)));
+            model->grads_memory = malloc_and_point_parameters(&model->grads, param_elements, param_sizeof);
+            model->grads_bytes = alloc_bytes;
+
+            alloc_bytes = 0;
+            fill_in_parameter_sizes(param_elements, param_sizeof, model->config);
+            for(int i = 0; i < NUM_PARAMETER_TENSORS; ++i) {
+                param_elements[i] /= multi_gpu_config.num_processes;
```
gordicaleksa, line 764: Will reread the paper so my knowledge is fresh but i thought we'd be dividing by the size of the node here and not by the total number of processes

train_gpt2.cu
```diff
@@ -919,14 +963,50 @@ void gpt2_backward_and_reduce(GPT2 *model, int* inputs, const int* targets, int
...
+                // this will result in a structure as follows:
+                // Main stream:     Calculate grads of layer 1 in buffer 1
+                // NCCL stream:     Wait for main stream
+                // Main stream:     Calculate grads of layer 2 in buffer 2
+                // NCCL stream:     Transmit data from buffer 1
+                // SYNC NNCL stream
+                // NCCL stream:     Wait for buffer 2 to be ready
+                // Main stream:     calculate grads of layer 3 in buffer 1
+                // ...
+                cudaCheck(cudaStreamSynchronize(multi_gpu_config.nccl_stream));
```
ademeure, line 979: cudaStreamSynchronize will wait on the host which isn't great for performance/overlap, would it be possible to synchronise with events between streams, similar to what we do else with:

    cudaCheck(cudaEventRecord(config->compute_nccl_sync, compute_stream));
    cudaCheck(cudaStreamWaitEvent(config->nccl_stream, config->compute_nccl_sync));

llmc/cuda_utils.cuh
```diff
@@ -260,4 +260,13 @@ __device__ __forceinline__ void stochastic_rounding(float in, float *out, unsign
     *out = in; // dummy function for when floatX is float (FP32 mode)
 }
 
+// Add two (potentially low-precision) vectors of size `n` together using stochastic rounding
+template<class Td, class Ts>
+__global__ void vector_add(Td* dst, const Ts* src, size_t n, unsigned seed) {
```
ademeure, line 265: Not sure if the time for this kernel is noticeable or not, but if it is, could make it much faster by using Packed128 and doing "src[idx] = 0" instead of a separate memset afterwards.

llmc/zero.cuh
```diff
@@ -593,5 +593,17 @@ float multi_gpu_cpu_float_sum(float value, MultiGpuConfig* config) {
...
 
+template<int N>
+void zero2_accumulate_grad(floatX* const (&dst)[N], floatX* const (&src)[N], const size_t (&nelem)[N], int layer, unsigned seed, cudaStream_t stream) {
+    for(int i = 0; i < N; ++i) {
+        size_t n = nelem[i] / multi_gpu_config.num_processes;
+        vector_add<<<CEIL_DIV(n, 512), 512, 0, stream>>>(dst[i] + layer * n,
+                                                              src[i] + multi_gpu_config.process_rank * n,
+                                                              n, seed + i);
+        cudaCheck(cudaGetLastError());
+        cudaCheck(cudaMemset(src[i], 0, nelem[i] * sizeof(floatX)));
```
ademeure, line 604: could be fused into the vector_add - these are tiny kernel calls in the same stream, so the idle time between might be significant (i.e. noticeable in Nsight Systems but not Nsight Compute)

ademeure (approved):
LGTM, looks like it could be made significantly faster/more efficient with relatively less extra complexity, but based on your performance data it already looks useful and possibly good enough for now as-is!

llmc/cuda_utils.cuh
ngc92: implemented the packed kernel; to keep things simple, that meant assuming that `dst` and `src` are of the same type, which is true for our current use-case.

llmc/zero.cuh
ngc92: If we move zeroing (pun intended) inside, I think we should then move the `vector_add` function from `cuda_utils.h`  to `zero.h`.

gordicaleksa (approved):
other than the comments i left which i think are all about refactoring - lgtm!

llmc/cuda_utils.cuh
```diff
@@ -260,4 +260,21 @@ __device__ __forceinline__ void stochastic_rounding(float in, float *out, unsign
...
+__global__ void vector_add(T* dst, const T* src, size_t n, unsigned seed) {
+    using t128 = Packed128<T>;
+    assert(n % t128::size == 0);
+    ptrdiff_t idx = ((ptrdiff_t)blockIdx.x * blockDim.x + threadIdx.x) * t128::size;
+    if (idx < n) {
+        t128 src_v = load128cs(src + idx);
+        t128 dst_v = load128cs(dst + idx);
+        for(int k = 0; k < t128::size; ++k) {
+            float sum = (float)dst_v[k] + (float)src_v[k];
+            stochastic_rounding(sum, &dst_v[k], seed + idx);
```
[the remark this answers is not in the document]
ngc92: if you do lots of gradient accumulation, you will incur more and more error because you end up adding small new gradients to the buffer of large accumulated gradients. With stochastic rounding, we at least stay correct in expectation, and will not systematically ignore small changes.

train_gpt2.cu
```diff
@@ -997,12 +1090,21 @@ float gpt2_calculate_grad_norm(GPT2 *model, MultiGpuConfig* multi_gpu_config) {
         // further sum the (partial) squared norm across all GPUs
         ncclCheck(ncclAllReduce(grad_norm_squared, grad_norm_squared, sizeof(float), ncclFloat, ncclSum, multi_gpu_config->nccl_comm, main_stream));
 #endif
-    } else {
```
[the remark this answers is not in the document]
ngc92: I generally would try to keep `#if` guards to a minimum, they just make it harder to reason about the code because you end up compiling different source code. So I'm trying to only hide code behind `#if` if it actually cannot compile in single-gpu mode, e.g., because we don't have the nccl_stream. Ideally, I think all that should be hidden inside `zero.h`.

train_gpt2.cu
```diff
@@ -364,9 +369,44 @@ void gpt2_allocate_weights(GPT2 *model) {
...
+    if(multi_gpu_config.zero_stage == 2) {
+        // Allocate parameter buffers for the current layers active "wave" of computation
+        size_t param_elements[NUM_PARAMETER_TENSORS];
+        size_t param_sizeof[NUM_PARAMETER_TENSORS];
+        GPT2Config wave_config = model->config;
+        // to prevent having to wait for comms to complete, we need to double-buffer gradients, so we need to
+        // allocate as if we had a two-layer network
+        wave_config.num_layers = 2;
+        fill_in_parameter_sizes(param_elements, param_sizeof, wave_config);
+        size_t alloc_bytes = 0;
```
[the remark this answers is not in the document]
ngc92: that seems like a good cleanup, but IMO should be a separate PR to keep the changes here to a minimum

train_gpt2.cu
```diff
@@ -919,14 +963,50 @@ void gpt2_backward_and_reduce(GPT2 *model, int* inputs, const int* targets, int
...
+                // Main stream:     Calculate grads of layer 2 in buffer 2
+                // NCCL stream:     Transmit data from buffer 1
+                // SYNC NNCL stream
+                // NCCL stream:     Wait for buffer 2 to be ready
+                // Main stream:     calculate grads of layer 3 in buffer 1
+                // ...
+                cudaCheck(cudaStreamSynchronize(multi_gpu_config.nccl_stream));
+#endif
+            }
             multi_gpu_async_reduce_gradient(pointers, nelem, &multi_gpu_config, main_stream);
```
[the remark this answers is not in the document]
ngc92: discussed on discord

train_gpt2.cu
```diff
@@ -976,7 +976,8 @@ void gpt2_backward_and_reduce(GPT2 *model, int* inputs, const int* targets, int
                 cudaCheck(cudaStreamSynchronize(multi_gpu_config.nccl_stream));
 #endif
             }
-            multi_gpu_async_reduce_gradient(pointers, nelem, &multi_gpu_config, main_stream);
+            nccl_wait_on_compute(&multi_gpu_config, main_stream);
```
gordicaleksa, line 977: cool! i like this solution

PR #596, opened 2024-06-15
Fix copy and casting from spawning excessive threads

gordicaleksa:
Due to Amdahl's law this optimization won't really matter (we only call this on the first call to `gpt2_update`), but for reference:

for a 124M model:
* `shard_num_parameters` (in zero stage 0, and single GPU setups) is 124M
* `shard.size` can be as small as 764 (layernorm shards)

So we're spawning 124M threads more than we need. :')


ngc92:
shouldn't shard size be 764/num_gpus, so even smaller?

gordicaleksa:
true! our code fails if we run on a 764+ GPU cluster? :P

PR #597, opened 2024-06-15
Fix stochastic rounding

gordicaleksa:
Previously our stochastic rounding logic didn't have a unique seed for each of the parameters we're rounding. This PR fixes that.

In more detail, previously:
* we were passing the same seed for all `NUM_PARAMETER_TENSORS` (16) sections of the param buffer.
* the above combined with this logic (extracted to demonstrate the gist of the logic):
```
unsigned int seed1 = Get2dNoiseUint(threadIdx.x, blockIdx.x + blockDim.y * gridDim.y, seed);
unsigned int seed2 = Get2dNoiseUint(threadIdx.x, blockIdx.x, seed1);
```
and given that `blockDim.y * gridDim.y` is constant and equal to either `1*1` or `1*12` (depending on whether we deal with one of the 4 layers like wte, wpe, lnfw, lnfb or with transformer layers respectively) - meant that many params were getting the same `seed2`!

In particular for the 4 layers (wte, wpe, lnfw, lnfb) we'd be doing:
```
unsigned int seed1 = Get2dNoiseUint(threadIdx.x, blockIdx.x + 1, seed);
unsigned int seed2 = Get2dNoiseUint(threadIdx.x, blockIdx.x, seed1);
```
which means that all params handled by a thread (blockIdx.x, threadIdx.x) are receiving the same seed.

Similarly for the 12 transformer layers - the offset would just be 12. But additionally all 12 shards would be reusing the same seed so the situation is even worse as `blockIdx.y` is not used in the generation of the seed.

The new logic makes sure that every single param has a unique seed.

### DIFF
--- a/llmc/adamw.cuh
+++ b/llmc/adamw.cuh
@@ -40,9 +40,7 @@ __device__ void adamw_update(Tp* params_memory, float* master_params_memory, Tg*
     float param = old_param - (learning_rate * (m / (sqrtf(v) + eps) + weight_decay * old_param));
     // update our low precision version of the parameters using stochastic rounding
     // this will be used in the next forward pass
-    // TODO: simply doing `params_memory[i] = (floatX)param;` breaks everything (why?)
-    unsigned int random = Get2dNoiseUint(threadIdx.x, blockIdx.x + blockDim.y * gridDim.y, seed);
-    stochastic_rounding(param, &params_memory[idx], random);
+    stochastic_rounding(param, &params_memory[idx], seed);
     // write the full, float version of the param into our master copy, if we maintain one
     // this will be used in the next update
     if (master_params_memory != NULL) { master_params_memory[idx] = param; }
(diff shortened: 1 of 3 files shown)


llmc/adamw.cuh
```diff
@@ -40,9 +40,7 @@ __device__ void adamw_update(Tp* params_memory, float* master_params_memory, Tg*
     float param = old_param - (learning_rate * (m / (sqrtf(v) + eps) + weight_decay * old_param));
     // update our low precision version of the parameters using stochastic rounding
     // this will be used in the next forward pass
-    // TODO: simply doing `params_memory[i] = (floatX)param;` breaks everything (why?)
-    unsigned int random = Get2dNoiseUint(threadIdx.x, blockIdx.x + blockDim.y * gridDim.y, seed);
-    stochastic_rounding(param, &params_memory[idx], random);
+    stochastic_rounding(param, &params_memory[idx], seed);
```
ngc92, line 43: isn't that destroying determinism, because now our  rounding depends on which address we get in memory?

gordicaleksa: I don't think it is? We save the seed inside our state?

As for indexing the same threads always take care of the same params?

ngc92: sorry, yes, I was thinking about  `Get2dNoiseUint` where the second argument is part of the seed. Here it is fine.

gordicaleksa: That's my second part of the comment - i don't think the PR as such breaks determinism? Because the indexing i pass into `Get2dNoiseUint` is deterministic.

PR #598, opened 2024-06-15
Add global norm prod kernels to dev

gordicaleksa:
Add the faster kernels we use in "prod" to our "dev" kernels.

### DIFF
--- a/dev/cuda/common.h
+++ b/dev/cuda/common.h
@@ -6,6 +6,7 @@
 #include <float.h>
 
 #define WARP_SIZE 32U
+extern cudaDeviceProp deviceProp;
 
 template<class T>
 __host__ __device__ T ceil_div(T dividend, T divisor) {
--- a/dev/cuda/common.h
+++ b/dev/cuda/common.h
@@ -25,8 +25,9 @@ __device__ float warpReduceSum(float val) {
 // the fact it's unique shared memory allows us to avoid an extra __syncthreads() call at the end
 // but if called inside a loop, the shared memory will be implicitly reused, so set final_sync to 1
 using reduction_func_t = float (*) (float);
+
 template<reduction_func_t warp_reduction>
-__device__ inline float blockReduce(float val, bool final_sync=false, float out_of_bounds=0.0f) {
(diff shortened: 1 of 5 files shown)


dev/cuda/common.h
```diff
@@ -5,6 +5,8 @@
 #include <cublasLt.h>
 #include <float.h>
 
+#define WARP_SIZE 32U
+extern cudaDeviceProp deviceProp;
```
ngc92, line 9: this seems like a dangerous change? You are promising the existence of a global variable deviceProp to all our dev kernels, but common.h  does not provide such a thing.
@gordicaleksa

gordicaleksa: We do the same thing in `llmc/cuda_common.h` afaiu - this just makes the variable globally accessible to all kernels in a specific file, but it's that file's responsibility to initiate it if it needs to use it.

I think that's how we do it elsewhere, e.g. in `train_gpt2.cu` we initialize it in the beginning of the main func (declaration is imported from `cuda_common.h`).

ngc92: there  are some important differences between the two situations:
in `llmc/cuda_common.h`, we expect that this header file will be included over multiple paths into the final executable, so we need to take care to ensure that we don't introduce duplicate definitions: hence, every function is marked inline, and variables are marked extern (in principle, we could mark them inline as well, C++17 makes inline available to variables). There is exactly one place where this extern variable gets defined, in `train_gpt2.cu` (that definition gets dragged into  profile and test due to the weird include thing we do). So for every header file in the llmc directory, you can use ´deviceProp` , and are guaranteed to get an actual object there.

In contrast, in dev, the header promises a  deviceProp, but that promise is broken in all but one file, leading to linker errors if you actually tried to use it.

Given that none of the functions are marked inline, I'm not sure if it actually makes sense to declare the device properties as extern. With the current code, maybe the best way would be to define them in that header, and populate them in `setup_main`

gordicaleksa: Ok, makes sense! Do you want to propose a patch? Happy to review & fix it

PR #600, opened 2024-06-16
Use faster kernel for LayerNorm forward

gordicaleksa:
I ran kernel 5 under `/dev/cuda/` (`./layernorm_forward 5`) on both RTX 3090 and H100 systems and it's faster on both of them.

Numbers:
kernel 3, optimal block size on:
- RTX 3090 → 32 (689.11 GB/s)
- H100 → 128-512 (1191.38 GB/s)

kernel 5, optimal block size on:
- RTX 3090 → 256 (773.13 GB/s)
- H100 → 256 (1244.5 GB/s)

It also seems to be more predictable in terms of what the optimal block size is, whereas kernel 3 perf varies a lot between these 2 systems.

Note: again, as per Ahmdal's law we're calling layernorm forward only once in our forward pass, the rest we are using the residual fused kernels. But still, it's a low hanging fruit unless I am missing out on some context as to why we didn't use kernel 5 here? Omission or by design? Is it because the way we're computing variance in kernel 5? (numerical stability?).

### DIFF
--- a/dev/cuda/fused_residual_forward.cu
+++ b/dev/cuda/fused_residual_forward.cu
@@ -133,7 +133,7 @@ __global__ void fused_residual_forward2(floatX* residual, floatX* normed, floatX
     for(int c = 0; c < C; ++c) {
         float out = (float)inp1[c] + (float)inp2[c];
         m += out;
-        residual[c] = out;
+        residual[c] = (floatX)out;
     }
 
     m = m / C;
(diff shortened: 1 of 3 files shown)


gordicaleksa:
Also, given that their caching profiles are the same (we use `cs` directive in both for `mean`, `rstd` and `out`), I don't expect e2e performance to be worse with the new kernel.

I did profile e2e run as well but the difference is too minor and the inter-run variations are large in comparison. Having said that It looks like it's slightly faster.

gordicaleksa:
Ran `profile_gpt2cu.py` on RTX 3090 and got the following:

kernel 3:
prof_1

kernel 6 (latest one):
prof_2

we went from 1.28 ms (0.29%) -> 1.11 ms (0.25%) per training step. It ain't much but it's honest work. :D

PR #602, opened 2024-06-16
Hotfix - proper handling of max num of block sums

gordicaleksa:
The previous assert logic was too restrictive as it depended on the number of layers of the model and the specification of the GPU (num of SMs & max number of threads per SM).

This PR fixes that. Tested, determinism works.

We can refactor this code afterwards, as it's ugly / I don't like the design:
`get_max_num_block_sums` is coupled to `global_norm_squared` if latter changes the former needs to follow, which is perfect for introducing bugs.


gordicaleksa:
Also fixes the numpy 2.0 CI failing error due to the fact that our CI is automatically pulling version 2.0 that's breaking every PR rn.

PR #603, opened 2024-06-17
Check determinism in CI

ngc92:
Extends the test script to validate determinism.
Some thoughts:
* With C memory management, it is quite easy to introduce memory leaks, e.g., loading from checkpoint twice without freeing in the middle causes a leak. To be somewhat defensive against this, I added a check that  validates that the pointer is null before allocating; which doesn't actually work, because the pointer value will be whatever random garbage was at that memory location before. In the end, I ended up requiring a separate gpt2_init call in front of either checkpoint loading or random generation.
* As the test script currently doesn't use the data loader, there are now some ugly ifdefs inside state saving/reading.

I did test cherry-picking this commit into an earlier version of the code, before the recent fixes, where it produces
>  Nondeterminism! Loss mismatch at step 11: 0.047421135008335 vs 0.049596484750509

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -350,6 +350,7 @@ void gpt2_init_common(GPT2 *model) {
     model->batch_size = 0;
     model->seq_len = 0;
     model->mean_loss = -1.0f; // -1.0f designates no loss, set at end of forward()
+    model->params_memory = NULL;
     // memory lazily initialized in backward()
     model->grads_memory = NULL;
     model->workload_indices = NULL; // on cpu, for encoder_backward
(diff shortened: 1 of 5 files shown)


train_gpt2.cu
```diff
@@ -1183,9 +1175,11 @@ void save_state(const char* filename, int step, GPT2* model, DataLoader* loader)
     state_header[10] = step; // step of the optimization
     // model state, state, start at 20 to leave some padding
     *((unsigned long long*)&state_header[20]) = model->rng_state; // random number generator state
-    // dataloader state, start at 30 to leave some padding
+    // dataloader state, start at 30 to leave some padding. Testing currently doesn't have a dataloader
+#ifndef TESTING
```
karpathy, line 1179: should we instead check if `loader` is `nullptr`?

ngc92: to me, that would imply that `loader` being `nullptr` is part of the normal operating conditions of this function, whereas the ifdef makes it look more like what it truly is, a hack to get some determinism checks working

PR #604, opened 2024-06-17
Fix guarding logic, refactor

gordicaleksa:
Minor refactor + made sure that the lack of guards in `residual_forward`, `gelu_forward`, and `gelu_backward_inplace` is justified.

### DIFF
--- a/dev/cuda/adamw.cu
+++ b/dev/cuda/adamw.cu
@@ -160,7 +160,7 @@ int main(int argc, char **argv) {
     float* params_memory = make_random_float(num_parameters);
     float* grads_memory = make_random_float(num_parameters);
     float* m_memory = make_random_float(num_parameters);
-    float* v_memory = make_random_float(num_parameters);
+    float* v_memory = make_random_float_01(num_parameters);
 
     // move to GPU
     float* d_params_memory;
--- a/dev/cuda/adamw.cu
+++ b/dev/cuda/adamw.cu
@@ -159,8 +159,8 @@ int main(int argc, char **argv) {
     // create random data on host (to be used for the CPU reference implementation)
     float* params_memory = make_random_float(num_parameters);
     float* grads_memory = make_random_float(num_parameters);
-    float* m_memory = make_random_float_01(num_parameters);
-    float* v_memory = make_random_float_01(num_parameters);
+    float* m_memory = make_random_float(num_parameters);
+    float* v_memory = make_random_float(num_parameters);
 
     // move to GPU
     float* d_params_memory;
--- a/llmc/layernorm.cuh
+++ b/llmc/layernorm.cuh
@@ -368,7 +368,7 @@ void layernorm_forward(floatX* out, floatX* mean, floatX* rstd,
 void residual_forward(floatX* out, const floatX* inp1, const floatX* inp2, int N, cudaStream_t stream) {
     NVTX_RANGE_FN();
     const int block_size = 256;
-    assert(N % block_size == 0);
+    assert(N % (block_size * x128::size) == 0);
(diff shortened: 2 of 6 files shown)


dev/cuda/adamw.cu
```diff
@@ -159,8 +159,8 @@ int main(int argc, char **argv) {
     // create random data on host (to be used for the CPU reference implementation)
     float* params_memory = make_random_float(num_parameters);
     float* grads_memory = make_random_float(num_parameters);
-    float* m_memory = make_random_float_01(num_parameters);
-    float* v_memory = make_random_float_01(num_parameters);
+    float* m_memory = make_random_float(num_parameters);
+    float* v_memory = make_random_float(num_parameters);
```
ngc92, line 163: I think v_memory should be >= 0, as its is mean squared gradient?

llmc/layernorm.cuh
```diff
@@ -368,7 +368,7 @@ void layernorm_forward(floatX* out, floatX* mean, floatX* rstd,
 void residual_forward(floatX* out, const floatX* inp1, const floatX* inp2, int N, cudaStream_t stream) {
     NVTX_RANGE_FN();
     const int block_size = 256;
-    assert(N % block_size == 0);
+    assert(N % (block_size * x128::size) == 0);
```
ngc92, line 371: can we  actually promise that?
x128::size is 8, with 256 block size, we get 2048, much larger than the 768 we have

gordicaleksa: `N` is not `C`, it's `B*T*C`?

The logic implicitly assumes that we don't need a guard so for that to be satisfied we need to make sure that not only is `N` divisible by the `block_size`, but it needs to be divisible by the `block_size * x128::size`.

That's the only way that the number of threads (`grid_size` * `block_size`) ends up being precisely N / x128::size and each handles x128::size floatX numbers.

Otherwise the ceiling operation will cause us to have more threads which would lead to an illegal memory access.

dev/cuda/adamw.cu
gordicaleksa: Correct!

On a side note: ideally we'd have Gaussian distribution supported and not only uniform in a certain range.

llmc/layernorm.cuh
ngc92: right, yes, this kernel ignores all shapes. Still not convinced that it actually is a good idea to not have guards inside the kernels.
Btw, I wouldn't count on illegal memory access errors to help us here. We allocate all activations as one big chunk, so we'd just end up messing up some data inside another tensor, but still inside the allocation.

gordicaleksa: even worse, true, that's what i meant by illegal (be it OS illegal or user illegal, maybe better call it "unintended" hah)

i can alternatively just add a guard? i didn't do it because i thought some of you profiled it and the guard somehow hurt the perf?

ngc92: I haven't profiled it, but I'd be extremely surprised if a single comparison operation in a memory-bound kernel had any noticeable effect on performance.
alternatively, maybe we should promote the asserts to if(...) exit(EXIT_FAILURE), so that compiling with -DNDEBUG (which is not an unreasonable choice if you want performance) doesn't break these super cheap checks

karpathy: @ngc92 I was thinking of creating our own `assertCheck` or things like that, which have whatever behavior we want. For example some checks we want to always remain in the code, even if compiling with DNDEBUG.

karpathy: (Also I agree that probably it's a good idea to have the guard checks inside kernels. It's more defensive to do a check as close to the place where it has to be true as possible. I merged a few PRs that removed them and then changed my mind about it later)

PR #605, opened 2024-06-17
Add learning rate schedulers

gordicaleksa:
Refactored the learning rate schedulers code - we'll keep all definitions inside "schedulers.h" as per our offline discussion.

Supported LR schedulers:
* Cosine with warmup
* Cyclic triangular (https://arxiv.org/abs/1506.01186)
* Constant
* Linear with warmup

### DIFF
--- a/llmc/schedulers.h
+++ b/llmc/schedulers.h
@@ -1,0 +1,43 @@
+
+
+// Cosine learning rate scheduler
+
+#ifndef SCHEDULERS_H
+
+#define SCHEDULERS_H
+
+#include <assert.h>
+#include <math.h>
+
+typedef struct {
+    float learning_rate;
+    int warmup_iterations;
+    int train_num_batches;
+    float final_learning_rate_frac;
+} CosineLearningRateScheduler;
+
+
+// learning rate schedule: warmup linearly to max LR, then cosine decay to LR * final_learning_rate_frac
+float get_learning_rate(CosineLearningRateScheduler *scheduler, int step) {
+    float step_learning_rate = scheduler->learning_rate;
+    if (step < scheduler->warmup_iterations) {
+        step_learning_rate = scheduler->learning_rate * ((float)(step + 1)) / scheduler->warmup_iterations;
+    } else {
+        float decay_ratio = ((float)(step - scheduler->warmup_iterations)) / (scheduler->train_num_batches - scheduler->warmup_iterations);
+        assert(0.0f <= decay_ratio && decay_ratio <= 1.0f);
+        float coeff = 0.5f * (1.0f + cosf(M_PI * decay_ratio)); // coeff starts at 1 and goes to 0
+        assert(0.0f <= coeff && coeff <= 1.0f);
+        float min_lr = scheduler->learning_rate * scheduler->final_learning_rate_frac;
+        step_learning_rate = min_lr + coeff * (scheduler->learning_rate - min_lr);
+    }
+    return step_learning_rate;
+}
+
+void lr_scheduler_init(CosineLearningRateScheduler *scheduler, float learning_rate, int warmup_iterations, int train_num_batches, float final_learning_rate_frac) {
+    scheduler->learning_rate = learning_rate;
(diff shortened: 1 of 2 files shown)


karpathy:
So I think we'll want to identify the scheduler using a string e.g. "cosine" from argparse, and inside the schedulers file should `strcmp` and build the right one.

karpathy:
Note there's also a previous PR on learning rate schedulers, including one that is trapezoidal.

gordicaleksa:
Argh apologies i saw your todo list and just picked it up assuming no one has done it, plus it's such an orthogonal contribution i thought it would have been merged by now - will check the PRs before submitting the next time :)

gordicaleksa:
https://github.com/karpathy/llm.c/pull/508 <- is this the one?

eliebak:
Yes i think it was mine, but your PR is much nicer nw aha. It can be nice to identify the schedule from argparse, and if we use trapezoidal, you also have to define the number of decay steps that you want. I can close mine and we can add this features to this one?

gordicaleksa:
Cool!

I can quickly implement the identification via a string. And happy to add trapezoidal as well. Let's see what Andrej thinks about the next steps.

gordicaleksa:
Pushed a very ugly switching between the schedulers based on the input argument.

In order to make this closer to what we're used to in Python/C++ we basically have to implement a mini virtual function pointer logic in order to allow for object oriented features.

Happy to do it, and won't be hard, just wanted to double check the decision makes sense because we'll be dealing with func pointers and i don't think we've done previously in the codebase.

cc: @karpathy

gordicaleksa:
It would look something along the lines of:
```
typedef struct LRScheduler {
    void (*initialize)(struct LRScheduler* self, double initial_rate);
    void (*step)(struct LRScheduler* self);
    double (*get_rate)(struct LRScheduler* self);
    double current_rate;
} LRScheduler;
```

train_gpt2.cu
```diff
@@ -1546,6 +1552,28 @@ int main(int argc, char *argv[]) {
     Tokenizer tokenizer;
     tokenizer_init(&tokenizer, "gpt2_tokenizer.bin");
 
+    // set up learning rate scheduler
```
karpathy, line 1555: we could move this new block inside the schedulers file. and the block below as well.

gordicaleksa: Done, simplified the code, it's ready.

llmc/schedulers.h
```diff
@@ -0,0 +1,137 @@
+/*
+Implements various learning rate schedulers.
+*/
+#ifndef SCHEDULERS_H
+
+#define SCHEDULERS_H
+
+#include <assert.h>
+#include <math.h>
+#include <cstring>
```
ngc92, line 10: string.h, if you stick with C headers

llmc/schedulers.h
```diff
@@ -0,0 +1,137 @@
...
+    // warmup_iterations <- not used.
+    // train_num_batches <- not used.
+    // final_learning_rate_frac <- not used.
+    return scheduler->learning_rate;
+}
+
+// switch to the appropriate learning rate scheduler
+float get_learning_rate(LRSchedulerType lr_scheduler_type, LearningRateScheduler *scheduler, int step) {
+    float step_learning_rate;
+    if (lr_scheduler_type == LR_SCHEDULER_COSINE) {
```
ngc92, line 121: switch/case?

gordicaleksa: i don't think there is any advantage using it?

llmc/schedulers.h
gordicaleksa: fixed

llmc/schedulers.h
ngc92: with switch, you usually get IDE warnings if you forget one enum value, not sure if you get the same with ifs. Otherwise, its pretty much equivalent.

karpathy:
forked this work into new PR
https://github.com/karpathy/llm.c/pull/627
made simplifications
added wsd
closing this one

PR #608, opened 2024-06-18
Clarified installing CUDNN front-end headers and arch linux install instructs

bgorlick:
Clarified instructions for installing CUDNN and front-end headers. Added cudnn installation steps for Arch Linux.


karpathy:
I don't want to bloat the main README file too much. Maybe we can have extra helpful instructions inside our `doc/` folder and link to it or something

PR #613, opened 2024-06-18
Fix dataloader & determinism testing

gordicaleksa:
Modified the code a bit in order to remove the hacky patch due to the original off by 1 error:
* Sample index should start from -1 (both after reset & shard advance)
* No need to do next batch call inside resume

Similarly for test:
* No need to call next batch before calling save state
* Inside the for loop we first load data and then do fwd/bwd/update as opposed to at the end

### DIFF
--- a/llmc/dataloader.h
+++ b/llmc/dataloader.h
@@ -109,9 +109,7 @@ void prepare_intra_shard_indices_(DataLoader *loader) {
 
 void dataloader_reset(DataLoader *loader) {
     loader->current_shard_idx = 0;
-    // we haven't drawn a current sample yet
-    // note that this line will underflow but that's ok, as +1 later will just overflow to 0
-    loader->current_sample_idx = (size_t) -1;
+    loader->current_sample_idx = 0;
 
     if (loader->should_shuffle) {  // shuffle the shards
         random_permutation(loader->shard_indices, loader->glob_result.gl_pathc, &loader->shuffle_rng);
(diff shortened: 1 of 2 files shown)


gordicaleksa:
will remove -1 since it's also hacky given that we have size_t, I'll just do ++ after the load so that index points to the current sample and not to the previous one.

ngc92:
have you check the recently merged fix. I think ++ after load doesn't work if you happen to save just at a shard boundary.

gordicaleksa:
Sorry might have missed it - which fix?

Fairly sure it will all work just fine.

Saving the state:
* We call `dataloader_next_batch`, `loader->current_sample_idx` is 1 smaller than `loader->shard_num_samples` so we don't kick off the `dataloader_advance_` function.
* We load the last batch and increment `loader->current_sample_idx` -> now we're right at the shard boundary
* We save the state

Loading the state:
* We call `dataloader_resume`
* `loader->current_shard_idx` is set to `loader->shard_num_samples`
* We load the old shard still (even though we're at the shard boundary)
* First call to `dataloader_next_batch` will realize we're at the boundary and just advance and reset the `loader->current_shard_idx` and `loader->current_sample_idx` to correct values

Notice this is the same thing that would have happened have we not called save/load state.

ngc92:
oh, right, in earlier versions of the code, the `advance` call was at the end of `next_batch`, now it is in the beginning. 
If we keep this semantics, we should call it `current_sample_idx`, that's just misleading

PR #614, opened 2024-06-18
Stricter FP32 tests

gordicaleksa:
* Stricter FP32 logit accuracy
* Much stricter FP32 loss accuracy
* Much stricter FP32 grad tensor accuracy (and somewhat stricter 16 bit accuracy)
* Copied over new expected loss values from PyTorch (they're a bit different and I took all 6 decimal points and not just 4)
* Also adapted our test logic to round loss to 6 decimal points

Regarding grad tensors: back when Andrej hardcoded the thresholds we had a bug in PyTorch that led to a bigger discrepancy between our PT vs C code - now that that's fixed we can be really strict and use `1e-6f` here.

### DIFF
--- a/test_gpt2.cu
+++ b/test_gpt2.cu
@@ -171,11 +171,12 @@ int main(int argc, char *argv[]) {
         logits_cpu[i] = (float)logits_cpu_raw[i];
     }
 
+    float logit_accuracy_threshold = 1e-3f;
+    float loss_diff_threshold = 0.00001f;
     // FP16 and lower require very high tolerances unfortunately. TODO look into more
-    float logit_accuracy_threshold = 1e-2f;
-    float loss_diff_threshold = 0.05f;
     #if defined(ENABLE_BF16) || defined(ENABLE_F16)
     logit_accuracy_threshold = 25.0f; // 15.0f was too low even without cuDNN?! :(
+    loss_diff_threshold = 0.05f;
     #endif
 
     // compare the output logits from the forward pass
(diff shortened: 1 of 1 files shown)


test_gpt2.cu
```diff
@@ -264,43 +257,54 @@ int main(int argc, char *argv[]) {
...
-            allok = allok & check_tensor(tensors1[7], tensors2[7], L * 4*C, "fcb", 5e-2f); // hmm a bit high
-            allok = allok & check_tensor(tensors1[8], tensors2[8], L * C * 4*C, "fcprojw", 5e-2f); // hmm a bit high
-            allok = allok & check_tensor(tensors1[9], tensors2[9], L * C, "fcprojb", 1.5e-2f);
-            allok = allok & check_tensor(tensors1[10], tensors2[10], L * C, "ln1w", 6e-4f);
-            allok = allok & check_tensor(tensors1[11], tensors2[11], L * C, "ln1b", 9e-3f);
-            allok = allok & check_tensor(tensors1[12], tensors2[12], L * C, "ln2w", 2e-3f);
-            allok = allok & check_tensor(tensors1[13], tensors2[13], L * C, "ln2b", 2.5e-3f);
-            allok = allok & check_tensor(tensors1[14], tensors2[14], C, "lnfw", 0.12f); // hmm bit higher
-            allok = allok & check_tensor(tensors1[15], tensors2[15], C, "lnfb", 2e-2f);
+
```
ngc92, line 260: can we turn this entire thing into a loop?
have  a small struct {size, name, threshold}, which also would make it easier to match thresholds and tensor names?

test_gpt2.cu
```diff
@@ -264,43 +257,54 @@ int main(int argc, char *argv[]) {
...
-        4.0607,
-        3.3202,
-        2.7176,
-        2.1811,
-        1.6538,
-        1.1680,
-        0.7367,
-        0.4008,
-        0.1874
+        5.270009,
```
ngc92, line 298: ideally, we'd export these from pytorch when we generate the reference file instead of having to copy them manually

gordicaleksa: hopefully it doesn't change that often? since we're not touching the reference imp that frequently

test_gpt2.cu
gordicaleksa: agree we could do that but i'd do it in a separate PR

rosslwheeler:
@gordicaleksa - this appears to be causing a failure when running:

`make testgpt2_cu USE_CUDNN=1 && ./testgpt2_cu`

It may or may not be seen on your environment but am able to see it here. 

@karpathy - I think this is what was causing the issue. I ran my tests one commit back from this in your repo and it passes consistently. If I checkout this commit, then the failures start. Not 100 percent sure but it does make some sense since this the test that's failing and there aren't many other changes to this file recently?

Can you confirm since you were seeing the failure consistently too? Thank you.

gordicaleksa:
It's certainly this PR - sad our CI didn't catch this! See https://github.com/karpathy/llm.c/pull/615 for a fix.

PR #615, opened 2024-06-19
Relax grad tensor thresholds in tests

gordicaleksa:
Our CI is using A4000 GPUs that didn't caught some of the tight grad tensor thresholds I've set in https://github.com/karpathy/llm.c/pull/614 

Relaxing a few of these reported by @rosslwheeler

On a side note: @karpathy would be cool to understand exactly how many GPUs we've got as CI runners, and potentially ask if we can trade a few of these for some different GPUs. :) (one can but hope)


rosslwheeler:
Passes my tests now.

PR #619, opened 2024-06-19
Cast Get2dNoiseUint computation to uint

gordicaleksa:
After the conversation with Squirrel (author of the noise rnd generator we're using) it might be a good idea to cast the intermediate computation to uint to avoid dealing with ints having UB (undefined behavior) overflow.

See [discord thread](https://discord.com/channels/1189498204333543425/1253010838596354189/1253010903176314970).


gordicaleksa:
cc: @ademeure

ademeure:
Looks good to me, although we could probably just change the parameters for indexX and indexY to unsigned int as well, rather than just casting them in the function, as I don't think we have any reason to ever pass negative values there. But either way solves the problem :)

gordicaleksa:
Agree, but left it there in case someone wants to experiment with negative values. Although it all boils down to unsigned.

PR #620, opened 2024-06-19
small fixes based on clang-tidy

ngc92:
Some of the changes based on running clang-tidy that make sense as an independent, much smaller PR.
This is
a) adding some missing checks
b) changing the layer index in the forward/backward loops to use 64 bit

The latter change means that `l *  ...` type products calculated as part of the offsets now do all their calculations in 64 bit, so they won't overflow when our models get large enough to have individual tensors with more than 2B entries.

I'm sure there are other places  where our code breaks with large models, so this is just a necessary, but not sufficient, step for further scaling up.


karpathy:
looks a bit weird that iterator over layers should be ptrdiff_t instead of just int. I guess this is a simple way to make a lot of the calculations happen in 64bit, it just looks weird... but doing it for all the calculations would be too much copy paste code adjustments and casts etc?

ngc92:
> looks a bit weird that iterator over layers should be ptrdiff_t instead of just int. I guess this is a simple way to make a lot of the calculations happen in 64bit, it just looks weird... but doing it for all the calculations would be too much copy paste code adjustments and casts etc?

yes; essentially, as soon as one factor is 64 bit, the entire product will be calculated in 64 bit, so instead of casting all over the place, this is just casting the factor that is common to all products.

ngc92:
oh, wait, that is actually the reason for the complaint in the first place... we _are_ calculating the large product in 64-bit, because  we use size_t for the other dimensions. Its just the `l * 3` part that clang-tidy finds offensive.

ngc92:
@karpathy removed the change of dtype for the loop iterator, because I'm not sure anymore if that is the right strategy. It might be, but in that case I think we should not use 64-bit numbers for B, T, etc., as this somehow ends up a redundant double-hack. So for the purpose of this minimal, hopefully uncontroversial, PR, lets set aside the integer multiplications.

invisiblepancake (commented):
just some small fixed for review. =)

PR #621, opened 2024-06-20
Added A10 to mfu.h

tiehexue:
values in mfu.h added for A10 are copied from https://www.nvidia.com/content/dam/en-zz/Solutions/Data-Center/a10/pdf/a10-datasheet.pdf

With two A10 or only one A10, it shows around "43.4% bf16 MFU".

BTW: it tooks 24 hours for two A10 to finish the training on fineweb10B.

### DIFF
--- a/llmc/mfu.h
+++ b/llmc/mfu.h
@@ -23,6 +23,7 @@ typedef struct {
 
 // basic default data from the nvidia whitepapers
 static const PerfData VOLTA = {125.0f, -1.f, 125.f, -1.f, -1.f, -1.f, 1530.f, 640.f};
+static const PerfData AMPERE = {62.5f, 125.f, 125.f, 125.f, 250.f, 500.f, 885.f, 72.f};
 static const PerfData AMPERE_DATACENTER = {156.f, 312.f, 312.f, 312.f, -1.f, -1.f, 1410.f, 432.f};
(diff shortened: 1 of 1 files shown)


karpathy:
Is Ampere a different class from Consumer and Datacenter? cc @ngc92  who originally wrote this code and did some of the researching.

llmc/mfu.h
```diff
@@ -39,6 +40,7 @@ typedef struct {
 static GPUEntry gpu_db[] = {
     {"Tesla V100-SXM2-16GB", &VOLTA, 640, 1530},
     {"Tesla V100-PCIE-32GB", &VOLTA, 640, 1530},
+    {"NVIDIA A10", &AMPERE, 72, 885},
```
ngc92, line 43: according  to https://www.techpowerup.com/gpu-specs/a10-pcie.c3793, the A10's boost clock is 1695 MHz
using the other ampere reference values, this would lead to ~62 bfloat TFlops, which would be consistent with other ampere cards, are the 125 with sparsity?

tiehexue: values were copied from https://www.nvidia.com/content/dam/en-zz/Solutions/Data-Center/a10/pdf/a10-datasheet.pdf . With sparsity, value are doubled to 250 or 500. 885MHz is base clock. I am not aware if it should be boost clock. If I change it to boost clock, it prints "22.5% bf16 MFU".

PR #623, opened 2024-06-20
feature/nccl only (delete MPI)

karpathy:


ngc92:
I don't think we can avoid the file system alltogether, as without MPI, this is the only "communication layer" that we actually have available before we initialize NCCL.
we should be much more careful about this file, though, e.g., if such a file remains left-over from a previous run, I think we just get race conditions in our file-system now.

README.md
```diff
@@ -134,19 +134,21 @@ sudo apt-get -y install libcudnn9-dev-cuda-12
 
 On top of this you need the [cuDNN frontend](https://github.com/NVIDIA/cudnn-frontend/tree/main), but this is just header files. Simply clone the repo to your disk. The Makefile currently looks for it in either your home directory or the current directory. If you have put it elsewhere, add `CUDNN_FRONTEND_PATH=/path/to/your/cudnn-frontend/include` to the `make` command-line.
 
-**multi-GPU training**. As of April 26, 2024 there is now also support for multi-GPU training using MPI and NCCL. Make sure you install MPI, e.g. on Linux:
+**multi-GPU training**. Support for multi-GPU training is availabel using NCCL. Make sure you download and install [NCCL](https://docs.nvidia.com/deeplearning/nccl/install-guide/index.html), e.g. on Linux:
```
chinthysl, line 137: typo in `availabel`

karpathy:
Copy pasting some Discord discussion. Adjustment to this PR

maybe one way we could do it is we
1) remove MPI for all actual compute
2) keep MPI as optional, use it for initializing NCCL
3) in case you hate MPI or can't get it to work, fall back on the janky file system idea
so in this world we keep the convenience of MPI and use it to init NCCL. and we don't delete that code, but we'd move it to its own mpi.h
if you don't have MPI we fall back to init NCCL through shared file system
certainly i think during the actual model run it feels very janky to use both MPI + NCCL for bits and pieces together.

PR #624, opened 2024-06-20
if available, use MPI env vars to initialize multi-gpu configs

ngc92:
let's see what windows thinks of this


train_gpt2.cu
```diff
@@ -1334,6 +1334,12 @@ void error_usage() {
     exit(EXIT_FAILURE);
 }
 
+int get_env_value(const char* env_name, int default_value) {
```
karpathy, line 1337: windows looks happy. maybe `get_env_int`?

PR #628, opened 2024-06-21
Check for the existence of CUDNN_FRONTEND_PATH and CUDNN_PATH variables.

koparasy:
Check for the existence of `CUDNN_FRONTEND_PATH` before looking in default directories. If it exists, skip additional checks. 

Further, add include and library paths based on `CUDNN_PATH` variable.


rosslwheeler:
IIRC - our algorithm explicitly wants to look in those directories first. Will review and comment back. Thanks.

koparasy:
> IIRC - our algorithm explicitly wants to look in those directories first. Will review and comment back. Thanks.

Unless I missed some part of the code (which is probable), the conditional statements ignore the value of the `CUDNN_FRONTEND_PATH` and thus result to only check under the current source directory and under the home directory.

The commit checks if the variable is set to bypass those checks. 

I hope this helps you out.

rosslwheeler:
Got it - will take a look. Thanks for clarifying.

Makefile
```diff
@@ -106,19 +106,26 @@ endif
...
-      $(error ✗ cuDNN not found. See the README for install instructions and the Makefile for hard-coded paths)
-    endif
+		ifndef CUDNN_FRONTEND_PATH
+			ifeq ($(shell [ -d $(HOME)/cudnn-frontend/include ] && echo "exists"), exists)
+				$(info ✓ cuDNN found, will run with flash-attention)
+				CUDNN_FRONTEND_PATH ?= $(HOME)/cudnn-frontend/include
+			else ifeq ($(shell [ -d cudnn-frontend/include ] && echo "exists"), exists)
+				$(info ✓ cuDNN found, will run with flash-attention)
+				CUDNN_FRONTEND_PATH ?= cudnn-frontend/include
+			else
```
rosslwheeler, line 116: Instead of the wrapping logic - try the below change and let me know if this works. Place above the "else" statement.

else ifneq ($(CUDNN_FRONTEND_PATH),)
  $(info ✓ cuDNN found, will run with flash-attention)

rosslwheeler:
Okay - so there's two changes in here. One a bug fix and the other is a feature. If you can split this into two PR's that would help. I would do the bug fix PR first. I'll comment directly on the bug fix change. Slightly different implementation.

koparasy:
> Okay - so there's two changes in here. One a bug fix and the other is a feature. If you can split this into two PR's that would help. I would do the bug fix PR first. I'll comment directly on the bug fix change. Slightly different implementation.

Alright. Here is the bugfix. I will create a new PR for the feature.

rosslwheeler:
Looks good. ty

PR #629, opened 2024-06-22
CI Dataloader test and ptx/sass file generator

rosslwheeler:
- New CI tests file - added dataloader test and ptx/sass file generator to it. 
- Cuda Makefile - added capability build from main Makefile. Added support for ptx and sass output files. 
- layernorm_forward.cu - remove #define (was already in header file)
- dev/test/Makefile - new file (based on main Makefile) for running tests
- dataloader.h - added casts to remove compiler warnings
- rand.h - removed unused parameter. Added casts to remove compiler warnings.
- train_gpt2.c - rename unsigned long long to uint64_t.

Pre-verified with CI.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -1,0 +1,60 @@
+name: Unit, Static and other Tests
+
+on:
+  create:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  dataloader_test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: test the dataloader without / with sanitize address
+        run: |
+          cd dev/test
+          make test_dataloader
+          ./test_dataloader   
+          make clean       
+          make test_dataloader TEST_CFLAGS="-fsanitize=address -fno-omit-frame-pointer"
+          ./test_dataloader          
+
+  ptx_and_sass_files:
+    runs-on: ubuntu-latest
+    container:
+      image: nvidia/cuda:12.4.1-devel-ubuntu22.04
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install OpenMP and OpenMPI
+        run: apt-get update && apt-get install -y libomp-dev libopenmpi-dev
+    
+      - name: Generate ptx/sass files and upload them to persistent storage
+        run: |
+          mkdir -p dev/cuda/ptx_sass_logs
+          make train_gpt2cu
+          cuobjdump --dump-ptx train_gpt2cu > dev/cuda/train_gpt2cu.ptx
+          cuobjdump --dump-sass train_gpt2cu > dev/cuda/train_gpt2cu.sass          
+          cd dev/cuda
+          make -j all_ptx
+          make -j all_sass
+          cp *.ptx ptx_sass_logs/
+          cp *.sass ptx_sass_logs/
+          ls ptx_sass_logs/
+
+      - name: Upload ptx/sass files
+        uses: actions/upload-artifact@v4
+        with:
+          name: ptx_sass_files
+          path: dev/cuda/ptx_sass_logs/
+          retention-days: 30 # days to retain
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -21,10 +21,10 @@ jobs:
       - name: test the dataloader without / with sanitize address
         run: |
           cd dev/test
-          make test_dataloader
+          make PRECISION=BF16 test_dataloader
           ./test_dataloader   
           make clean       
-          make test_dataloader TEST_CFLAGS="-fsanitize=address -fno-omit-frame-pointer"
+          make PRECISION=BF16 TEST_CFLAGS="-fsanitize=address -fno-omit-frame-pointer" test_dataloader 
           ./test_dataloader          
 
   ptx_and_sass_files:
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -52,9 +52,49 @@ jobs:
           cp *.sass ptx_sass_logs/
           ls ptx_sass_logs/
 
+      - name: Generate ptx/sass files for A100 and upload them to persistent storage
+        run: |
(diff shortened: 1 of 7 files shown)


.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install OpenMP and OpenMPI
+        run: apt-get update && apt-get install -y libomp-dev libopenmpi-dev
+    
+      - name: Generate ptx/sass files and upload them to persistent storage
+        run: |
+          mkdir -p dev/cuda/ptx_sass_logs
+          make train_gpt2cu
```
ngc92, line 45: should we specify which GPUs we want to generate SASS for here?

dev/cuda/Makefile
```diff
@@ -8,8 +8,20 @@ ifeq ($(NVCC),)
 		$(error nvcc not found.)
 endif
 
+ifneq ($(CI),true) # if not in CI, then use the GPU query
+  ifndef GPU_COMPUTE_CAPABILITY # set to defaults if: make GPU_COMPUTE_CAPABILITY= 
+    GPU_COMPUTE_CAPABILITY = $(shell __nvcc_device_query) # assume if NVCC is present, then this likely is too
```
ngc92, line 13: at least on my system, it actually is not present

dev/cuda/Makefile
```diff
@@ -8,8 +8,20 @@ ifeq ($(NVCC),)
...
+    GPU_COMPUTE_CAPABILITY := $(strip $(GPU_COMPUTE_CAPABILITY))
+  endif
+endif
+
 # Compiler flags
-CFLAGS = -O3 --use_fast_math
+ifeq ($(GPU_COMPUTE_CAPABILITY),) # set to defaults if: make GPU_COMPUTE_CAPABILITY= 
+  CFLAGS = -O3 --use_fast_math 
+else
+  CFLAGS = -O3 --use_fast_math --generate-code arch=compute_$(GPU_COMPUTE_CAPABILITY),code=[compute_$(GPU_COMPUTE_CAPABILITY),sm_$(GPU_COMPUTE_CAPABILITY)]
```
ngc92, line 22: this makes it impossible to generate code for more than one arch at a time, which might be useful, e.g., for sass export, or if you  want to run the same executable on  A100 and H100 nodes  without recompiling (and having to remember to recompile, if it's a shared cluster file system)

dev/test/Makefile
```diff
@@ -0,0 +1,164 @@
...
+OUTPUT_FILE = -o $@
+CUDA_OUTPUT_FILE = -o $@
+
+# NVCC flags
+# -t=0 is short for --threads, 0 = number of CPUs on the machine
+NVCC_FLAGS = -O3 -t=0 --use_fast_math -std=c++17
+NVCC_LDFLAGS = -lcublas -lcublasLt
+NVCC_INCLUDES =
+NVCC_LDLIBS =
+NCLL_INCUDES =
```
ngc92, line 22: nccl?

dev/test/Makefile
```diff
@@ -0,0 +1,164 @@
...
+    NVCC_LDLIBS += -lmpi -lnccl
+    NVCC_FLAGS += -DMULTI_GPU
+  else
+    $(info ✗ OpenMPI is not found, disabling multi-GPU support)
+    $(info ---> On Linux you can try install OpenMPI with `sudo apt install openmpi-bin openmpi-doc libopenmpi-dev`)
+  endif
+endif
+
+# Precision settings, default to bf16 but ability to override
+PRECISION ?= BF16
```
ngc92, line 117: should we actually define a default for CI?

dev/cuda/Makefile
rosslwheeler: Got it - at one point this was a request but it's easy to remove.

dev/test/Makefile
rosslwheeler: Funny - that's in the Main makefile too. We should fix it there as well. And it's not being used. Noticed this the other day.

dev/test/Makefile
rosslwheeler: Good point - yes may be best to force a PRECISION on the command line so we always know what we're building.

dev/cuda/Makefile
rosslwheeler: Can remove this - per the below comment too.

dev/cuda/Makefile
ngc92: it is a good convenience option, just want to make sure we're aware of its limitations; especially when we want  to build sass in CI for both  A100 and H100

.github/workflows/ci_tests.yml
rosslwheeler: This was @ademeure's request. Any comments?

dev/cuda/Makefile
ngc92: don't think we need to remove this; I think its the difference between the ubuntu cuda package (which is on PATH in my system) and those from the nvidia sources (which does have this utility in its bin folder; it's just not on PATH in my setup)
I think it would still break on my system because of the two different GPUs, i.e., its return value is  `86,89` 
Having two different GPUs might not be too uncommon in a workstation setup, where you have on cheap(ish) GPU handling the actual display output, and a second GPU for the heavy compute.

dev/cuda/Makefile
rosslwheeler: I see - should we add an explicit step to generate sass for the A100 and H100?

.github/workflows/ci_tests.yml
rosslwheeler: @ngc92 - so for your comment above - had the question below too - would be to add the A100 and H100's SASS in a step here?

rosslwheeler: Added H100 and A100 to ptx and sass file generation

dev/cuda/Makefile
rosslwheeler: Added a change for this.

dev/test/Makefile
rosslwheeler: Removed this.

dev/test/Makefile
rosslwheeler: Added this explicit requirement for PRECISION being set.

PR #426, opened 2024-05-17
NCCL only multi gpu training for slurm enabled cluster

chinthysl:
Scheduling jobs using Slurm seems much easier in a multi-node training setup compared to setting up MPI for the cluster.
This draft contains the changes to use mpirun for single-node training and Slurm for multi-node training.

PyTorch uses one of the backends from Gloo, MPI, and NCCL for DDP. Maybe we don't need to use both MPI and NCCL together. It should be either CUDA-aware MPI or NCCL. Something to discuss further.

I got some interesting performance numbers in a large-scale training setup using llm.c.
I used the Ahrefs DGX H100 (80GB) Superpod. The cluster uses NVLINK for intra-node communications and InfiniBand RDMA for inter-node communications.

Without CUDNN we can reach up to batch_size=12
```
NumNodes    NumProcess  Latency(ms) tokens/sec(k)
1           8           395         249
2           16          674         296
4           32          817         484
8           64          1056        748
16          128         1087        1455
24          192         1320        1794
48          384         2054        2303
```
oreo-multinode-llmc

Without CUDNN we can reach up to batch_size=24
```
NumNodes    NumProcess  Latency(ms) tokens/sec(k)
1           8           539         366
2           16          798         498
4           32          991         798
8           64          1158        1365
16          128         1247        2531
24          192         1478        3198
```
oreo-multinode-llmc-cudnn

PR #632, opened 2024-06-22
MPI/TCP/FS for NCCL-init

gordicaleksa:
Instead of mixing NCCL & Open MPI during training: let's transition to using only NCCL. To the best of my knowledge there are no downsides here, they're equivalent and speedwise i couldn't observe any difference.

By doing this we scope down our MPI dependence to `multi_gpu_config_init`.

Context: why are we trying to reduce dependence on Open MPI?
`slurm-wlm` package - which is the easiest and thus likely the most frequent way folks setup their slurm clusters - dropped PMIx support which means we can't use slurm in a multi node setup together with Open MPI.

Regarding `multi_gpu_config_init` we have a few options as far as i can tell:
* Keep using MPI (but as mentioned slurm users won't be able to run llm.c in multi-node setup)
* Transition to logic where we synchronize on file system (similar to https://github.com/karpathy/llm.c/pull/426)
* TCP sockets (nodes will have to contain IP address of each other to communicate eitherway so we can use that)

So:
* slurm users can toggle a switch, remove OpenMPI code dependence, and run in multi-node setup
* mpirun users can use llm.c as is and it will just work

### DIFF
--- a/llmc/zero.cuh
+++ b/llmc/zero.cuh
@@ -50,11 +50,9 @@ void mpi_check(int status, const char *file, int line) {
 #endif // MULTI_GPU
 
 // ----------------------------------------------------------------------------
-// MPI / multi-processing setup
-
 // Parameters specific to training on multiple GPUs.
 typedef struct {
-    int process_rank;      // Rank of this process among all MPI processes. 0 if no multi-GPU.
+    int process_rank;      // Rank of this process among all processes. 0 if no multi-GPU.
     int num_processes;     // Total number of processes. 1 if no multi-GPU.
     int local_device_idx;  // This process GPU index on current machine. 0 if no multi-GPU.
 
@@ -69,6 +67,7 @@ typedef struct {
     ncclComm_t nccl_comm;       // NCCL communication primitive, used for collective multi-GPU work.
     cudaStream_t nccl_stream;   // CUDA Stream to perform NCCL operations.
     cudaEvent_t compute_nccl_sync; // Event used to synchronize NCCL with the compute
+    float* unified_buffer;
 #endif
 } MultiGpuConfig;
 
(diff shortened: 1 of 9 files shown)


gordicaleksa:
I tried to use NCCL instead of MPI but realized there are some hard dependencies:
* `cudaSetDevice` has to be called before `ncclCommInitRank` thus we can't use NCCL to reduce values inside `multi_gpu_get_local_device_idx`.
* Similarly wanted to replace `MPI_Bcast` with NCCL, but since this needs to happen before `ncclCommInitRank` it can't be done.

So as mentioned above we're left with MPI, filesysystem sync, TCP. Will test the latter 2 tomorrow.

gordicaleksa:
The PR is tested and ready @karpathy - comments / feedback welcome!

gordicaleksa:
NOTE: Sockets are currently not cross-platform that's why Windows test is failing.

Linux & Mac are supported.

So we either have to tweak it a bit to make it x-platform OR we remove support on Windows OR we remove sockets altogether.

llmc/zero.cuh
```diff
@@ -69,70 +72,243 @@ typedef struct {
...
 } MultiGpuConfig;
 
 #ifdef MULTI_GPU
+void send_nccl_id_to_clients(ncclUniqueId *nccl_id, int client_sockets[], int num_clients) {
+    for (int i = 0; i < num_clients; ++i) {
+        if (send(client_sockets[i], nccl_id, sizeof(*nccl_id), 0) == -1) {
+            printf("Failed to send nccl_id");
+            exit(EXIT_FAILURE);
+        }
+        close(client_sockets[i]);
```
ngc92, line 86: closeCheck?

llmc/zero.cuh
```diff
@@ -69,70 +72,243 @@ typedef struct {
...
+        // Step 5) MAX_CLIENTS specifies the maximum number of clients that can be queued for this server
+        if (listen(server_socket, MAX_CLIENTS) < 0) {
+            printf("Listen failed");
+            exit(EXIT_FAILURE);
+        }
+
+        // Step 6) accept connections from clients
+        printf("Waiting for clients to connect...\n");
+        while (num_clients < MAX_CLIENTS) {
+            if ((new_socket = accept(server_socket, (struct sockaddr *)&address, (socklen_t*)&addrlen)) < 0) {
```
ngc92, line 140: why  is addrlen  of type `int` and not `socklen_t`?

train_gpt2.cu
```diff
@@ -1349,10 +1357,17 @@ int main(int argc, char *argv[]) {
     int recompute = 1; // recompute during backward setting, 0 = none, 1 = recompute gelu
     int zero_stage = 0; // Zero Optimization Stage for Multi-GPU training
     int hellaswag_eval = 0;
+    // multi-node settings
+    int num_processes = 1;  // this should be set by the slurm environment
+    int process_rank = 0;  // this should be set by the slurm environment
+    int gpus_per_node = 8;  // this should be set by the slurm environment
+    char nccl_init_method[256] = "mpi";  // "tcp" or "fs" or "mpi"
+    char server_ip[256] = "-1";  // used if init_method set to "tcp" -> set to your server ip address
+    char fs_path[256] = "/tmp";  // used if init_method set to "fs" -> set to a shared filesystem path
```
ngc92, line 1366: wouldn't  /tmp usually be a path local to the current node?

llmc/zero.cuh
```diff
@@ -193,6 +212,126 @@ ncclUniqueId get_nccl_id_via_tcp(MultiGpuConfig* result, const char* server_ip)
     return nccl_id;
 }
 
+#ifdef _WIN32
```
karpathy, line 215: any way to put this stuff into unistd.h, which is our windows-specific code? I think it would need a tiny refactor so that the needed imports are minimal, e.g. maybe not passing in a MultiGPUConfig but its members more directly.

train_gpt2.cu
karpathy: Kind of agree, in a previous PR it was `/dfs` or so, worried that a person might get a wrong impression of this flag if they see its default as `/tmp `.

llmc/zero.cuh
gordicaleksa: i think we can do that in the follow-up PR if you're ok with that? e.g. @rosslwheeler has a windows setup ready and he said he can pick it up

train_gpt2.cu
gordicaleksa: I'll change it to "" (empty) as it doesn't matter since the init method is set to "mpi" by default. This var is used only when init is set to "fs".

llmc/zero.cuh
gordicaleksa: Don't have a satisfying answer other than it's a pattern i've observe people use everywhere, see:
1) https://gist.github.com/SkrewEverything/2c535e83a3a7b8e5b7aa490009a87fbb
2) https://stackoverflow.com/questions/61106749/how-to-read-data-from-socket-correctly
3) your favorite LLM

it is casted to `socklen_t` before calling `accept`.

PR #635, opened 2024-06-24
On-device reductions

ngc92:
Moves loss calculation to backward, and ensures  we can do more on-device reductions and fewer host<->device transfers.
Also enables a micro-optimization, that validate does not calculate dlogits anymore.

### DIFF
--- a/llmc/cuda_common.h
+++ b/llmc/cuda_common.h
@@ -42,8 +42,8 @@ extern cudaDeviceProp deviceProp;
 #define CEIL_DIV(M, N) (((M) + (N)-1) / (N))
 
 // short-cuts for compile-time boolean values that can be used as function arguments
-constexpr std::bool_constant<true> TRUE;
-constexpr std::bool_constant<true> FALSE;
+constexpr std::bool_constant<true> True;
+constexpr std::bool_constant<true> False;
 
 // ----------------------------------------------------------------------------
 // Error checking
--- a/llmc/cuda_common.h
+++ b/llmc/cuda_common.h
@@ -8,6 +8,7 @@ Common utilities for CUDA code.
 #include <stdio.h>
 #include <math.h>
 #include <string>
+#include <type_traits>      // std::bool_constant
 #include <cuda_runtime.h>
 #include <nvtx3/nvToolsExt.h>
 #include <nvtx3/nvToolsExtCudaRt.h>
@@ -39,6 +40,10 @@ extern cudaDeviceProp deviceProp;
 
 // convenience macro for calculating grid/block dimensions for kernels
 #define CEIL_DIV(M, N) (((M) + (N)-1) / (N))
+
+// short-cuts for compile-time boolean values that can be used as function arguments
+constexpr std::bool_constant<true> TRUE;
+constexpr std::bool_constant<true> FALSE;
 
 // ----------------------------------------------------------------------------
 // Error checking
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -325,7 +325,6 @@ typedef struct {
     int seq_len; // the sequence length (T) of current forward pass
     int* inputs; // the input tokens for the current forward pass
     int* targets; // the target tokens for the current forward pass
-    bool has_targets; // has the targets buffer been populated during the forward pass
     float mean_loss; // after the last backward micro-batch, will be populated with mean loss across all GPUs and micro-steps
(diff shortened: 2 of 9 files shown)


llmc/cuda_common.h
```diff
@@ -40,6 +41,10 @@ extern cudaDeviceProp deviceProp;
 // convenience macro for calculating grid/block dimensions for kernels
 #define CEIL_DIV(M, N) (((M) + (N)-1) / (N))
 
+// short-cuts for compile-time boolean values that can be used as function arguments
+constexpr std::bool_constant<true> True;
```
karpathy, line 45: I don't love bool_constant, it looks scary and caused an issue in an earlier PR, are we not able to avoid its use? Maybe just 0/1?

train_gpt2.cu
```diff
@@ -686,45 +690,61 @@ void gpt2_forward(GPT2 *model, const int* inputs, const int* targets, size_t B,
     }
 
     matmul_forward_cublaslt(acts.output, acts.lnf, params.wte, NULL, B, T, C, Vp, main_stream);
+    cudaCheck(cudaDeviceSynchronize());
+}
 
-    // also forward the cross-entropy loss function if we have the targets
-    if (targets != NULL) {
+
+float gpt2_validate(GPT2 *model) {
```
karpathy, line 697: maybe a comment on what gpt2_validate does as a function? what is it validating exactly?

train_gpt2.cu
```diff
@@ -883,12 +911,24 @@ void gpt2_backward_and_reduce(GPT2 *model, int* inputs, bool last_step) {
...
+        cudaCheck(cudaMemcpyAsync(&model->mean_loss, model->accumulated_mean_loss, sizeof(float), cudaMemcpyDeviceToHost, main_stream));
+
         floatX* const pointers[] = {grads.wte, grads.wpe, grads.lnfw, grads.lnfb};
         const size_t nelem[] = {Vp * C, T * C, C, C};
         multi_gpu_async_reduce_gradient(pointers, nelem, &multi_gpu_config, main_stream);
     }
 
     cudaCheck(cudaDeviceSynchronize());
+    if(last_step) {
+        model->mean_loss /= B*T*grad_accum_steps;
```
karpathy, line 930: this part also means that model->mean_loss is invalid/wrong if we try to read it out when !last_step

llmc/cuda_common.h
ngc92: not really. The point of bool constant is that it encodes true/false in the _type_, so it can be used for function overloading/template argument deduction.
The alternative is to specify template arguments by hand, leading to
```cpp
kernel<<<block, grid>>><true>(args)
```
which is arguably much worse than bool_constant.
(or the other way round, not sure if <<<<>>>  or <> comes first).
The issue in the other PR was that we were not including the standard header where it is defined, so it wasn't really anything specific to bool_constant.

train_gpt2.cu
ngc92: yes, the name isn't great. Maybe we should actually move the gpt2_forward call for validation inside here, so that the API allows for a future optimization in which the forward pass doesn't store all activations, and thus can run much larger batch size in validation.

train_gpt2.cu
```diff
@@ -686,45 +690,61 @@ void gpt2_forward(GPT2 *model, const int* inputs, const int* targets, size_t B,
     }
 
     matmul_forward_cublaslt(acts.output, acts.lnf, params.wte, NULL, B, T, C, Vp, main_stream);
+    cudaCheck(cudaDeviceSynchronize());
```
karpathy, line 693: curious why

llmc/cuda_common.h
karpathy: I guess my question is why do we need template arguments for a switch

train_gpt2.cu
ngc92: it might not be necessary, but  I think for now, a programming model where every gpt2_... function guarantees that all its work is done by the time the function returns hopefully reduces the chances for errors.

llmc/cuda_common.h
ngc92: by having it as a template argument, we make the compiler generate two specialized versions of the function that can be optimized specifically for the task at hand. The performance gains are _probably_ really minor/below the measurement noise.

train_gpt2.cu
ngc92: yes, we can cheaply set it to -1 in the else branch to mark it as unavailable

PR #636, opened 2024-06-24
rolling checkpoints

karpathy:
checkpoints are either MINOR or MAJOR and minor checkpoints get deleted with a rolling window. This is an optimization that will allow us to save state more often, but preserve disk space overall. And if we ever see loss spikes and such, it is easier to reset the state to an earlier checkpoint that is more recent

Introduces two new flags:

```c
    fprintf(stderr, "  -nk <int>   max number of checkpoints to keep in the directory, removing old ones (0 = disable, default)\n");
    fprintf(stderr, "  -nm <int>   every how many step checkpoints are considered major? major checkpoints never get deleted.\n");
```

Example usage:

```
        -n 20 \
        -nk 3 \
        -nm 100 \
```

This will write checkpoint every 20 steps, only keeps up to 3 minor checkpoints in the directory at any time, except every 100th checkpoint is considered a MAJOR checkpoint, and will not be deleted.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -1284,7 +1284,7 @@ void load_state(int* step, GPT2* model, DataLoader* loader, const char* filename
 
 // ----------------------------------------------------------------------------
 // CLI, poor man's argparse
-// unclaimed flags lol: p
+// (all single letters have been claimed now)
 
 void error_usage() {
     fprintf(stderr, "Usage:   ./train_gpt2cu [options]\n");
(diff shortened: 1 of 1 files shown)


train_gpt2.cu
```diff
@@ -1284,7 +1284,7 @@ void load_state(int* step, GPT2* model, DataLoader* loader, const char* filename
 
 // ----------------------------------------------------------------------------
 // CLI, poor man's argparse
-// unclaimed flags lol: p
+// (all single letters have been claimed now)
```
ngc92, line 1287: :)

train_gpt2.cu
```diff
@@ -1697,6 +1703,26 @@ int main(int argc, char *argv[]) {
...
             }
             multi_gpu_barrier(&multi_gpu_config);
+            // we only keep checkpoints_keep checkpoints on disk to save space
+            // so now that we wrote a new checkpoint, delete one old one (unless it is a "major" checkpoint)
+            // we only do this is checkpoint keeping is turned on (checkpoints_keep > 0)
+            int step_delete = step - checkpoints_keep * checkpoint_every;
+            if (checkpoints_keep > 0 && step_delete > 0 &&
+               (major_checkpoint_every == 0 || step_delete % major_checkpoint_every != 0)
+                ) {
+                printf0("deleting minor checkpoint %d\n", step_delete);
```
ngc92, line 1713: make a dedicated `delete_checkpoint` function?

train_gpt2.cu
```diff
@@ -1697,6 +1703,26 @@ int main(int argc, char *argv[]) {
                 fclose(done_file);
             }
             multi_gpu_barrier(&multi_gpu_config);
+            // we only keep checkpoints_keep checkpoints on disk to save space
```
ngc92, line 1706: do we  need the barrier above? If so, maybe add a comment why

PR #637, opened 2024-06-25
add outlier detector, test for it, and start tracking z score of loss

karpathy:
still TODO:

- track grad norm in addition to loss (do we have to split the `gpt2_update` function)
- add argparse option to monitor loss and grad norm for outliers (e.g. z > 3)
- add handling of instabilities (e.g. if z > 3, skip update)
- possibly: instead of skipping update we could EXIT_FAILURE, and when -y 1 resumes the optimization from last checkpoint, we could try to somehow advance the DataLoader past the point of instability (this is what PaLM did)

### DIFF
--- a/dev/test/test_outlier_detector.c
+++ b/dev/test/test_outlier_detector.c
@@ -1,0 +1,52 @@
+/*
+Tests our OutlierDetector
+
+compile and run as (from dev/test directory)
+gcc -O3 -I../../llmc -o test_outlier_detector test_outlier_detector.c -lm && ./test_outlier_detector
+*/
+
+#include <stdlib.h>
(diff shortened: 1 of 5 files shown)


karpathy:
Introduced two new flags:

```c
    fprintf(stderr, "  -sl <float> outlier stability: skip update if loss goes above this in zscore (0.0f=off, default=3.0f)\n");
    fprintf(stderr, "  -sg <float> outlier stability: skip update if grad_norm goes above this in zscore (0.0f=off, default=3.0f)\n");
```

They default to 0.0 (old behavior) but if they are e.g. 2.5, then if the loss or grad z score are > 2.5, the update is skipped.

PR #638, opened 2024-06-25
add http status check for download_file

BilyZ98:
This pr add http status check for download_file and reports error if it fails to download file. 
I met a issue that curl does not report error when it failed to pull model from huggingface through proxy and get http error 503 .
But the output still is that 
`Downloaded gpt2_124M_bf16.bin to /home/bily/llm.c-fork/dev/../gpt2_124M_bf16.bin`
despite that it does not downloads the file successfully.

### DIFF
--- a/dev/download_starter_pack.sh
+++ b/dev/download_starter_pack.sh
@@ -37,8 +37,12 @@ download_file() {
     fi
 
     # Download the file
-    curl -s -L -o "$FILE_PATH" "$FILE_URL"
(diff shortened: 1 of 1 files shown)


miro5lav (approved):
Looks good.

BilyZ98:
Hi @miro5lav , thanks for review. Do you how this pr can be merged? It looks like it has to go through workflow that can only be triggered by maintainer.

PR #641, opened 2024-06-25
Add check versions of functions

gordicaleksa:
Add socket close check functions - consistent with the rest of the codebase.

### DIFF
--- a/llmc/utils.h
+++ b/llmc/utils.h
@@ -15,6 +15,8 @@
 // implementation of dirent for Windows is in dev/unistd.h
 #ifndef _WIN32
 #include <dirent.h>
+#else
+#include <winsock2.h>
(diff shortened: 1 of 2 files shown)


llmc/utils.h
```diff
@@ -15,6 +15,10 @@
 // implementation of dirent for Windows is in dev/unistd.h
 #ifndef _WIN32
 #include <dirent.h>
+#include <arpa/inet.h>
+#else
```
rosslwheeler, line 19: You may be able to move the winsock2.h file into dev/unistd.h. I would try to remove the pragma (if possible) and use an explicit /link Ws2_32.lib in LDLIBS in the Windows section of the Makefile (but have not tested this).

PR #643, opened 2024-06-25
dev/download_starter_pack.sh: adding SIGINT trap and current download…

Ricardicus:
The download_starter_pack.sh felt like it got stuck and I did not know why.
It only displays the files that got downloaded and not the ones in progress.
I also couldn't stop it with CTRL-C properly.

This adds:
- CTRL-C (SIGINT) stopping the script
- A display that shows what files are being downloaded/have been downloaded (see below)

It can look like this (when tree files have been downloaded, and there are three to go)

```bash
Downloading gpt2_124M.bin...
Downloading gpt2_124M_bf16.bin...
Downloading gpt2_124M_debug_state.bin...
Downloaded gpt2_tokenizer.bin to /Users/user/llm.c/dev/../gpt2_tokenizer.bin   
Downloaded tiny_shakespeare_train.bin to /Users/user/llm.c/dev/data/tinyshakespeare/tiny_shakespeare_train.bin   
Downloaded tiny_shakespeare_val.bin to /Users/user/llm.c/dev/data/tinyshakespeare/tiny_shakespeare_val.bin   
```

### DIFF
--- a/dev/download_starter_pack.sh
+++ b/dev/download_starter_pack.sh
@@ -23,9 +23,16 @@ FILES=(
     "tiny_shakespeare_val.bin"
 )
 
+move_cursor() {
+    local row=$1
+    local col=$2
+    echo -ne "\033[${row};${col}H"
+}
+
 # Function to download files to the appropriate directory
 download_file() {
     local FILE_NAME=$1
+    local ORDER=$2
     local FILE_URL="${BASE_URL}${FILE_NAME}?download=true"
     local FILE_PATH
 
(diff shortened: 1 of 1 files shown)


Ricardicus:
I think this looks pretty good.
I removed the clear screen at the beginning now.

Ricardicus:
I also added a little sanity check for the "curl" requirement. 
I ran this script in the Docker image "nvidia/cuda:12.4.1-devel-ubuntu22.04" from Dockerhub and
apparently it does not come shipped with curl and the scripts output looked a little messy.

PR #649, opened 2024-06-26
tune openmpi make a bit more convenient

karpathy:


Makefile
```diff
@@ -207,12 +207,16 @@ else
   endif
 endif
 
+# Attempt to find and include OpenMPI on the system
+OPENMPI_DIR = /usr/lib/x86_64-linux-gnu/openmpi
```
rosslwheeler, line 211: You might want to make this ?= that way you can override it at the command line?

karpathy: Good call

PR #650, opened 2024-06-26
muP (maximum update parametrization) [WIP]

gordicaleksa:
WIP.


gordicaleksa:
## Ablation study

Setting:
lr = `0.006`
width goes from `64` -> `4096` (geometric sequence, 2x coefficient)

Baseline muP:
image

At step 4 it seems we observe a bit more oscillations. These do dissapear with a bit lower learning rate so not sure what to think of it. Ultimately running a sweep of runs and confirming we have stable HPs will be the final test.

Here is what happens setting `attn_mult` to 1. It looks like it's a bit more stable? My implementation follows that of mutransformers but it's possible they had a bug when it comes to the `attn_mult` logic, will have to consult paper again:
image

SP (standard parametrization) baseline:
image
These clearly explode.

---

Comment out zeroing out of the embedding/readout layer & queries:
image
Observe that one of the layers is decreasing with width, which is undesirable.

Remove learning rate / weight decay modulating logic:
image

Remove `scale = (model->use_mup && i != 0 && i != 1) ? mup_scale_inv*scale : scale;`:
image

Remove logit scaling by `1/width_mult`:
image

Set `attn_mult` to 1 and replace `1/d` q*k scaling with the usual `1/sqrt(d)`:
image

alxndrTL:
Hello, I looked over your implementation of muP in both CUDA and Python, and found that in CUDA (`layernorm.cuh`) you scale the output of all the layernorms by `mup_scale` (which is defined as `sqrt(mup_width_multiplier)`) :

image

but I haven't seen the same scaling in your Pytorch version `(train_gpt2.py`) (which is very concise and clear btw!)
Also, this is not mentioned in the `doc/mup/mup.md` file so I was wondering about where did it come from.
Thank you

gordicaleksa:
it's mentioned in `mup.md` it is under 3.3 (`1. / model->mup_width_mult`), I don't use `sqrt` in the c/cuda code?

alxndrTL:
Ok I didn't realize that the layernorm code I showcased is only used pre-logits, as per 3.3 (I thought it was used for every layer norms).

gordicaleksa:
## Hyperparam sweeps

scheduler sweep:
image

Conclusion: `cosine` is a good choice.

`attn_mult` tunable param sweep:
image

Conclusion: Using `1` is a good default.

`lr` sweeps:
image

Conclusion: ~1/2^10 is a sweet spot for lr. The curves are stable as we increase the depth, i.e. the optimal `lr` is invariant to depth scaling.

`out_mult` sweep (`out_mult` is currently not supported in this PR but it's a minor tweak i've implemented locally):
image

Conclusion: `1` is a good default.

cc: @karpathy 

next steps:
* See whether `mup` authors reply to my issue here: https://github.com/microsoft/mup/issues/76
* Run `mup` at 7B scale

gordicaleksa:
@Yuchen Jin would be great to kick off a 7B mup run if you have some bandwidth! :)

gordicaleksa:
@YuchenJin would be great to kick off a 7B mup run if you have some bandwidth! :)

YuchenJin:
> @YuchenJin would be great to kick off a 7B mup run if you have some bandwidth! :)

Hey @gordicaleksa, happy to! Do you want me to just run the two scripts (`scripts/mup_coordinate_check.sh` and `dev/mup_coordinate_check_visualize.py`)? What LR should I use for the 7B model?

habanoz:
@gordicaleksa 

https://github.com/karpathy/llm.c/blob/b125cc6ae2b850cd3511720fb1eaf9db83130972/train_gpt2.py#L165

AFAIK this line zero initializes modules with 'LLMC_SKIP_INIT' flag if mup is enabled. There is only one module with 'LLMC_SKIP_INIT' flag, it is lm_head. lm_head weight is tied to wte.weight. 

Since embedding layers are initialized later in the code, what is the purpose of the zero initialization referenced above?

habanoz:
> @gordicaleksa
> 
> https://github.com/karpathy/llm.c/blob/b125cc6ae2b850cd3511720fb1eaf9db83130972/train_gpt2.py#L165
> 
> AFAIK this line zero initializes modules with 'LLMC_SKIP_INIT' flag if mup is enabled. There is only one module with 'LLMC_SKIP_INIT' flag, it is lm_head. lm_head weight is tied to wte.weight.
> 
> Since embedding layers are initialized later in the code, what is the purpose of the zero initialization referenced above?

After reading mup.md, I can see now that MUP requires output layers to be initialized to zero.

The code assumes that embeddings are initialized before linear layers, which is a correct but IMHO a weak assumption.

Thanks for the great work.

PR #651, opened 2024-06-27
Async optimizer state and model checkpointing

chinthysl:
Additional feature to checkpoint optimizer state and model parameters using a non blocking background thread.
Memcopy device buffers to host buffers in one shot and let the background thread do I/O operations.

In my 8xA100 setup rough latency improvement is 5.4 sec vs 2.3 sec ~ 2X improvement.
When it comes to the larger model sizes this feature will save a lot of time.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -9,6 +9,7 @@ GPT-2 Transformer Neural Net training loop. See README.md for usage.
 #include <string_view>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <pthread.h>
 // ----------- CPU utilities -----------
 // defines: fopenCheck, freadCheck, fcloseCheck, fseekCheck, mallocCheck
 // defines: create_dir_if_not_exists, find_max_step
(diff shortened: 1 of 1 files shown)


train_gpt2.cu
```diff
@@ -390,11 +408,29 @@ void gpt2_write_to_checkpoint(GPT2 *model, const char* checkpoint_path) {
...
+        // write the parameters
+        device_to_file(model_file, model->params_memory, model->num_parameters_bytes,
+                    IO_BUF_SIZE, main_stream);
+        // close file, we're done
+        fcloseCheck(model_file);
+    }
+    else {
+        // transfer device data to host memory
+        char* buffer_space;
+        cudaCheck(cudaMallocHost(&buffer_space, model->num_parameters_bytes));
```
ademeure, line 422: My experience with cudaMallocHost is it can be *much* slower than a regular malloc which eventually "pays for itself" because the data transfer is so much faster, but for a single transfer, I suspect it might be slower than just a regular malloc.

So if that holds true here, this might need to either keep that memory allocated permanently or just use a regular malloc.

PR #654, opened 2024-06-27
Set RNG seed manually with '-rg' parameter

ademeure:
This adds a '-rg' parameter to manually set the RNG seed. This is useful to see if a change is beneficial or not when the difference is potentially real but smaller than the noise threshold. As we are fully deterministic (for a given system), we can't just run multiple time with the same settings anymore.

The same seed will be used for both gpt2_init_common() and gpt2_build_from_random(), so this effectively changes the default seed from 42 to 13371337 for gpt2_build_from_random(). It does not change the default seed for the non-random initialisation path or test_gpt2cu in any way.


karpathy:
I tried to fix this PR properly but it gets really tricky with the seeds because there are multiple of them (for stochastic rounding vs. for init), of different dtypes  and it feels kind of delicate. Might revisit later.

PR #657, opened 2024-06-30
Remove per-layer attproj and fcproj activation tensors

ademeure:
I'm not sure how we missed this but we don't actually need these tensors at all for the backward pass at all! Possibly this wasn't the case when residual/layernorm/recompute were implemented differently, or I'm missing something? Either way, it's a significant GPU memory saving.

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -199,7 +199,7 @@ void* malloc_and_point_parameters(ParameterTensors* params, size_t* param_elemen
     return params_memory;
 }
 
-constexpr int NUM_ACTIVATION_TENSORS = 23;
+constexpr int NUM_ACTIVATION_TENSORS = 21;
 typedef struct {
     floatX* encoded; // (B, T, C)
     floatX* ln1; // (L, B, T, C)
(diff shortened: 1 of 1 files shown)


train_gpt2.cu
```diff
@@ -666,8 +661,8 @@ void gpt2_forward(GPT2 *model, const int* inputs, size_t B, size_t T) {
         // reuse the same activation buffer at each layer, as we'll re-compute the gelu during backward
         // very useful because we dramatically reduce VRAM usage, and may be able to fit larger batch size
         floatX* l_fch_gelu = (model->recompute < 1) ? acts.fch_gelu + l * B * T * 4*C : acts.fch_gelu;
-        floatX* l_fcproj = acts.fcproj + l * B * T * C;
         floatX* l_residual3 = acts.residual3 + l * B * T * C;
+        floatX* scratch = (floatX*)acts.output; // used for non-cudnn attention, fcproj, attproj, etc.
```
ngc92, line 665: could also use `acts.scratch_btc` directly, now that we have these available during the forward pass.

ngc92:
wow, this is great

train_gpt2.cu
ademeure: That's true, I think maybe it's possible to get rid of the "single layer" 'acts.lnf' and/or 'acts.fch_gelu' as well if we also have scratch_btc to play with as you said!

That wouldn't save a lot of memory  relative to my earlier change though ofc.

PR #658, opened 2024-07-01
Add optimized GPU kernels for encoder_backward using shared memory

vyom1611:
This commit introduces a new optimized kernel for the positional encoder backward pass:

Kernel Version 3: Uses shared memory to reduce global memory accesses and improve performance. Each thread block loads data into shared memory and performs atomic additions to dwte and dwpe.

Slightly improves performance as seen from this basic timing result:

Kernel 2 (pre-existing):
Screenshot 2024-07-01 at 1 18 13 PM

New Kernel 3: 

Screenshot 2024-07-01 at 1 18 45 PM

More detailed profiling (used Nsight systems on modal benchmark script: Nvidia A100 + 80GB MEM)  :

Kernel 2:
Screenshot 2024-07-01 at 1 59 23 PM

Kernel 3:
Screenshot 2024-07-01 at 2 00 06 PM

### DIFF
--- a/dev/cuda/encoder_backward.cu
+++ b/dev/cuda/encoder_backward.cu
@@ -11,6 +11,11 @@ parallelizes over B,T,C, uses atomics to add to dwte, dwpe
 version 2 is another naive port
 parallelizes over C, loops over B,T; much slower than version 1
 ./encoder_backward 2
+
+version 3 uses shared memory to reduce global memory accesses and improve performance
+parallelizes over B,T,C, uses atomics to add to dwte, dwpe, utilizes shared memory for intermediate results
+./encoder_backward 3
+
 */
 
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


karpathy:
I don't think we can merge this because we need determinism, this uses atomicAdd.
Summoning @ademeure  for comment too

ademeure:
It's fine for /dev/cuda/ which is the only thing this PR touches though - I don't mind including this new kernel as an extra comparison point which can be useful! It shows that using shared memory helps but only slightly. But I'm not sure whether it's worth it since we can't actually use it for the training code.

The bigger issue is /dev/cuda/encoder_backward.cu is out of date and doesn't include the kernels we actually use (which avoids the atomics). There's a few other /dev/cuda/ files which are out of date unfortunately.

dev/cuda/encoder_backward.cu
```diff
@@ -84,6 +89,29 @@ __global__ void encoder_backward_kernel2(float* dwte, float* dwpe,
     }
 }
 
+// Optimized implementation using shared memory
+__global__ void encoder_backward_kernel3(float* dwte, float* dwpe,
+                                         const float* dout, const int* inp,
+                                         int B, int T, int C) {
+    extern __shared__ float shared_mem[];
```
karpathy, line 96: wait this is not used right?

vyom1611: No, I added that on my own rather than from the actual GPT code.

PR #659, opened 2024-07-01
Add option to specify device when in a single GPU setup

gordicaleksa:
Useful for running multiple single GPU experiments in parallel (using this for mup experiments).

### DIFF
--- a/llmc/zero.cuh
+++ b/llmc/zero.cuh
@@ -407,7 +407,7 @@ int multi_gpu_get_local_device_idx(int process_rank, int num_processes) {
 
 #endif
 
-MultiGpuConfig multi_gpu_config_init(int num_processes, int process_rank, int gpus_per_node, char* server_ip, char* fs_path, char* init_method) {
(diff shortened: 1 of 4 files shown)


ngc92:
`export CUDA_VISIBLE_DEVICES=1`  :)

karpathy:
I think I prefer using `CUDA_VISIBLE_DEVICES`, less code footprint?

gordicaleksa:
I don't think that will work? we hardcode device 0 in our code? @karpathy @ngc92

ngc92:
AFAIK, the numbers correspond to available devices, but I'm not 100% sure

karpathy:
@gordicaleksa it will work, if it's anything like what i did before with PyTorch. I do this all the time on PyTorch when I do sweeps.

karpathy:
Will close for now and bring back in case the CUDA_VISIBLE_DEVICES doesn't work.

gordicaleksa:
@karpathy it doesn't work just tested, we should add this option for people who want to run multiple runs and schedule them across specified devices.

PR #662, opened 2024-07-01
fix the overfit single batch behavior to actually overfit batch, not …

karpathy:
…microbatch, we do this more cleanly by simply resetting the dataloader every step

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -1580,8 +1580,9 @@ int main(int argc, char *argv[]) {
     printf0("+-----------------------+----------------------------------------------------+\n");
 
     // build DataLoaders for both train and val
+    int permute_train_loader = (overfit_single_batch == 1) ? 0 : 1;
     DataLoader train_loader, val_loader;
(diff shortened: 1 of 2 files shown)


karpathy:
This recovers our ability to reproduce our test with our train. Example:

```
# reproduce our test with our train
make train_gpt2cu PRECISION=FP32
./train_gpt2cu -b 4 -t 64 -d 256 -l 0.0001 -v 200 -s 200 -a 1 -x 10 -r 0 -f 0 -e "gpt2_124M.bin"
```

We can also use grad accum 2:

```
./train_gpt2cu -b 2 -t 64 -d 256 -l 0.0001 -v 200 -s 200 -a 1 -x 10 -r 0 -f 0 -e "gpt2_124M.bin"
```

Same for PyTorch:

```
python train_gpt2.py --write_tensors 0 --batch_size 4
python train_gpt2.py --write_tensors 0 --batch_size 2
```

Now print the same things. Grad accum kicks in to recover the total desired batch size and gives the same results.

PR #666, opened 2024-07-02
Improved style & comments for gpt2_forward()

ademeure:
Obviously very subjective, but it feels like a big improvement to me!

My comments on Residual+LN fusion and how it corresponds to the relevant paper for Pre-LN is a good example of something that doesn't feel obvious to me and will hopefully be helpful to other humans/AIs.


karpathy:
I'm sorry @ademeure , I don't like it at all 😂 , it looks forced, with the whitespace aligning, renaming of main stream, added vertical space and indentation for function calls, and I think the paper ref comments are a bit on the excessive side. I'll close this sorry :p

ademeure:
No worries! I think the most useful part was the paper ref comments but I agree it's not quite the right format/place for them. If you ever end up making a video that mentions the residual/LN fusion then pointing to those diagrams and the subtle differences in terms of potential fusion for Pre-LN and Post-LN might be interesting :)

PR #667, opened 2024-07-03
Fix eval dataloader div by zero for < 4 batch size

gordicaleksa:
We need at least batch size of 4 to support the current eval logic.

Alternatively we can rewrite the eval a bit, but that's probably overengineering at this point?

Worst case that could happen is that due to this batch increase and the user missing the warning (classic :)) the training runs until we hit eval and then it crashes due to an OOM.


karpathy:
I added an assert for it instead right now in master, ty for finding it!

PR #674, opened 2024-07-09
move `set_zero_configs` into `zero.cuh`

ngc92:
So far, this function has remained in the main file, because it relied on `printf0`, which in turn required the global `multi_gpu_config` not available in the zero header.
This PR moves the global object to the header (as an inline variable to prevent ODR problems), defines printf0 as a macro (so we don't need C-style va-args, making format strings more easily checkable by static analysis tools),
and finally moves `set_zero_configs` to `zero.cuh`.


ngc92:
superseded by #689

PR #675, opened 2024-07-10
Add option to remove biases

gordicaleksa:
Add a cmdline option that allows us not to use biases for attn/fc layers.


test_gpt2.cu
```diff
@@ -304,16 +304,16 @@ int main(int argc, char *argv[]) {
...
-        4.060681f,
-        3.320085f,
-        2.717550f,
-        2.181066f,
-        1.653923f,
-        1.168050f,
-        0.736873f,
-        0.401021f,
-        0.187493f
+        6.319745f,
```
karpathy, line 307: ???

gordicaleksa: Biases are now set to 0 inside of PyTorch code -> this leads to different values.

gordicaleksa: This is obviously a destructive action on learned GPT-2 biases -> but we don't care about samples / we're just doing tests so should be fine.

Open to suggestions, such as setting `d12` as default? Instead of using a pretrained model?

PR #680, opened 2024-07-12
Add GPT3 model series

ngc92:
While GPT3 models are not uniquely identified by their _depth_, there is a one-to-one mapping of model parameter count to model _width_. This makes the specifications a bit inconsistent, as a 12-layer gpt2 model is `gpt2:12`, but for gpt3 it is `gpt3:768`, but we can get away with a single number.

Additionally, the changes enable interpolating of model sizes, so that one can adapt the number of parameters to the available GPU memory (this is not necessarily smooth, though, because interpolation adjusts just one parameter, whereas the others have discontinuous jumps based on the pre-defined models).


ngc92:
superseded by #688

PR #684, opened 2024-07-13
Adding CI check for exceeding loss tolerance

rosslwheeler:
- Modified the train_gpt32 (FP32) run with gpt2_124M.bin to add requested parameter changes in CI
- Added a check for the loss to see if it varies > 5 percent. This is configurable - so we can change this to a lower value if that's more appropriate.

Tested in CI


karpathy:
Sorry I don't understand the history/context for this change, is it following up on some conversation? Why are the args being changed around?

rosslwheeler:
Yes, this was a suggestion in our discord conversation. Just replied to it there for your reference. Let me know if this is still of interest.

PR #694, opened 2024-07-18
Model init cleanup

ngc92:
* consolidate model parameter allocation to a single source location
* made gradient buffer accumulation eager
* moved encoder determinism helper buffers so that they are eagerly allocated by forward -> backward is now free of mallocs
* moved dtype helpers into utils file

### DIFF
--- a/llmc/cuda_utils.cuh
+++ b/llmc/cuda_utils.cuh
@@ -80,6 +80,36 @@ typedef Packed128<float> f128;
 typedef Packed128<floatX> x128;
 
 // ----------------------------------------------------------------------------
+// DType support
+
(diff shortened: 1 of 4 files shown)


llmc/cuda_utils.cuh
```diff
@@ -79,6 +79,36 @@ __device__ void store128cg(ElementType* target, Packed128<ElementType> value) {
 typedef Packed128<float> f128;
 typedef Packed128<floatX> x128;
 
+// ----------------------------------------------------------------------------
```
karpathy, line 82: i think it makes a lot more sense that this be here, nice

PR #699, opened 2024-07-19
Simplified/faster "backward bias" kernel (column reduction)

ademeure:
Inspired by the simplified kernel I made yesterday for train_gpt2fp32.cu, this is a column reduction kernel (currently specialised to backward bias but really quite generic) that is much simpler and also faster than our current kernel.

It does not (need to) support the Auxiliary Buffer approach as even at OC=768, it gets ~40% better performance than the current kernel by handling as many B*T elements per threadblock as possible while maintaining 32B coalesced accesses (which is the L1 sector size; for the larger OC, it will increase the number of parallel OCs to do 64B or 128B coalesced reads).

Possibly even more performance available by managing to make the shared memory load/stores (and global memory stores) use 128-bit load/stores rather than "just" the global memory loads being x128, and maybe by manually unrolling some of the loops rather than using "#pragma unroll 4" (by assuming/asserting B*T is a multiple of a larger number) - but it already gets very good efficiency so possibly diminishing returns, at least on a H100 PCIe.

Lots(!) of comments and explanations in the /dev/cuda/ version of the kernel, possibly could remove even more of the comments in /llmc/matmul.cuh and just refer to the /dev/cuda/ comments instead...

### DIFF
--- a/llmc/matmul.cuh
+++ b/llmc/matmul.cuh
@@ -13,90 +13,51 @@ Matrix Multiplication, with help from cuBLASLt
 // ----------------------------------------------------------------------------
 // CUDA kernels
 
-template<typename OutFloat, bool UseAuxBuffer>
-__global__ void matmul_backward_bias_kernel9(OutFloat* dbias, const floatX* dout, int B, int T, int OC,
-                                             std::bool_constant<UseAuxBuffer>) {
-    constexpr const int bdx = 4;
-    constexpr const int bdy = WARP_SIZE / bdx;
-    assert(blockDim.x == bdx);
-    assert(blockDim.y == bdy);
+// this is really a general purpose column reduction kernel, could be generalised to other things!
+// OC columns and B*T rows ==> per-column sum reduction with OC outputs in dbias
+// each block handles (blockIdx.x * x128::size) columns and (blockIdx.y) rows
+// data is contiguous along rows, 128B coalesced loads with BF16 require blockIdx.x >= 8
+// for small OC, we want smaller blockIdx.x to get more blocks and better GPU utilisation
+// (see comments in /dev/cuda/matmul_backward_bias.cu for even more information)
+template <int block_dim_x=2, int block_dim_y=512, typename OutFloat=floatX>
+__global__ void matmul_backward_bias_kernel10(OutFloat* dbias, const floatX* dout, int B, int T, int OC) {
+    constexpr int block_size = block_dim_x * block_dim_y;
+    __shared__ float smem[block_size * x128::size];
 
-    int warp_d = (int)threadIdx.x;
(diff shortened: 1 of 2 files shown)


llmc/matmul.cuh
```diff
@@ -13,90 +13,51 @@ Matrix Multiplication, with help from cuBLASLt
...
-            }
+// this is really a general purpose column reduction kernel, could be generalised to other things!
+// OC columns and B*T rows ==> per-column sum reduction with OC outputs in dbias
+// each block handles (blockIdx.x * x128::size) columns and (blockIdx.y) rows
+// data is contiguous along rows, 128B coalesced loads with BF16 require blockIdx.x >= 8
+// for small OC, we want smaller blockIdx.x to get more blocks and better GPU utilisation
+// (see comments in /dev/cuda/matmul_backward_bias.cu for even more information)
+template <int block_dim_x=2, int block_dim_y=512, typename OutFloat=floatX>
+__global__ void matmul_backward_bias_kernel10(OutFloat* dbias, const floatX* dout, int B, int T, int OC) {
+    constexpr int block_size = block_dim_x * block_dim_y;
```
ngc92, line 24: put in asserts that the template args match the actual block dimensions?

llmc/matmul.cuh
```diff
@@ -13,90 +13,51 @@ Matrix Multiplication, with help from cuBLASLt
...
+        x128 packed_dout = load128(dout + column_idx + row * OC);
+        for (int k = 0; k < x128::size; k++) {
+            column_sum[k] += (float)packed_dout[k];
         }
     }
-
-    __shared__ float sub_results[x128::size][WARP_SIZE][bdy];
-
-    // reduce within-warp results
+    // todo - currently don't use x128 for smem, so we stride by block_size to avoid bank conflicts
```
ngc92, line 38: nit: since everything in smem is float, that comment should say f128 :)

invisiblepancake (approved):

llmc/matmul.cuh
ademeure: done, I am always a bit reluctant to do that inside the kernel rather than the launch code, because (at least in the way we compile llm.c) it does result in extra SASS instructions :( but I'd rather be safe than sorry as well, maybe at some point we can find a better way to have those checks but typically not pay the cost across the codebase.

ademeure:
I generalised the kernel as a generic column reduction :) (and clarified the terminology is assuming cuBLAS-like column major layout, rather than C++ array-like row major layout)

There shouldn't be any functional change except it now also uses x128 for the final output to global memory (and read from global memory when accumulating).

invisiblepancake (approved):

PR #702, opened 2024-07-20
Restore from master weights (& allow restoring from a checkpoint of different precision)

ademeure:
This is fully deterministic for new checkpoints where the new rng_state_last_update is saved, so that stochastic rounding from master weights is done with the exact same seeds (while restoring the actual final rng_state again afterwards, in case anything else changed it between that update and saving the checkpoint).

In the case where we are resuming from a checkpoint (not just a regular model file) and we have master weights, this simply skips loading the weights from the checkpoint completely, so it doesn't matter if they are not even the right number of bytes.

It should be useful to check if FP32 helps with runs exploding, and going forward it will allow FP8 runs to not care too much about what format the non-master weights are saved, so we don't need to worry about changes breaking compatibility etc...

### DIFF
--- a/llmc/adamw.cuh
+++ b/llmc/adamw.cuh
@@ -61,6 +61,26 @@ __global__ void adamw_kernel3(Tp* params_memory, float* master_params_memory, Tg
                  );
 }
 
+template <typename Tp>
+__global__ void params_from_master_kernel(Tp* params_memory, float* master_params_memory, size_t num_parameters, unsigned int seed, bool check_identical) {
+    size_t idx = blockIdx.x * blockDim.x + threadIdx.x;
+    if (idx >= num_parameters) { return; }  // guard
+
+    Tp rounded_param;
+    float param = master_params_memory[idx];
+    stochastic_rounding(param, &rounded_param, seed);
+
+    if (check_identical) {
+        // check if the rounded parameter is identical to the master parameter (for debugging only)
+        if (params_memory[idx] != rounded_param) {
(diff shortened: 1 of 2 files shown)


llmc/adamw.cuh
```diff
@@ -61,6 +61,16 @@ __global__ void adamw_kernel3(Tp* params_memory, float* master_params_memory, Tg
                  );
 }
 
+template <typename Tp>
+__global__ void init_from_master_kernel(Tp* params_memory, float* master_params_memory, size_t num_parameters,
+                                          ptrdiff_t w_stride, ptrdiff_t s_stride, unsigned int seed, bool check_identical) {
```
karpathy, line 66: unused `check_identical`?

ademeure: oops, that used to be for debug lots that I removed, thanks

train_gpt2.cu
```diff
@@ -1532,7 +1554,7 @@ int main(int argc, char *argv[]) {
     gpt2_init_common(&model);
     if (resuming == 1) {
         // if `-y 1` was set, then we are resuming from the latest checkpoint
-        gpt2_build_from_checkpoint(&model, filename_buffer);
+        gpt2_build_from_checkpoint(&model, filename_buffer, true);
```
karpathy, line 1557: at this point `model.use_master_weights` is not yet initialized with `use_master_weights`, that happens below, and is just the default (true). This variable is used inside `gpt2_build_from_checkpoint` and this is probably a bug?

PR #705, opened 2024-07-21
Refactor C code

gordicaleksa:
Realized we do activation size computation inline, so changed it to be consistent with how we compute parameter sizes. It's more readable than bloating the fwd pass.


invisiblepancake:
c code

PR #707, opened 2024-07-22
Add KV cache for inference

gordicaleksa:
WIP. Will update description after the draft has progressed. :)

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -1637,6 +1637,9 @@ int main(int argc, char *argv[]) {
     // set up the Tokenizer
     Tokenizer tokenizer;
     tokenizer_init(&tokenizer, "gpt2_tokenizer.bin");
+    if (tokenizer.init_ok == 0) {
+        exit(EXIT_FAILURE);  // error message already printed inside tokenizer_init
+    }
 
     // set up learning rate scheduler
     LearningRateScheduler lr_scheduler;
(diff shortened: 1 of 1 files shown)


train_gpt2.cu
```diff
@@ -647,11 +651,10 @@ void gpt2_forward(GPT2 *model, const int* inputs, size_t B, size_t T) {
     // forward pass
     ParameterTensors params = model->params; // for brevity
     ActivationTensors acts = model->acts;
-    encoder_forward(acts.encoded, model->inputs, params.wte, params.wpe, B, T, C, main_stream); // encoding goes into residual[0]
+    encoder_forward(acts.encoded, model->inputs, params.wte, params.wpe, model->use_kv, model->kv_offset, B, T, C, main_stream); // encoding goes into residual[0]
 
     // first layernorm isn't fused
-    layernorm_forward((model->recompute < 2) ? acts.ln1 : acts.lnf, acts.ln1_mean, acts.ln1_rstd, acts.encoded, params.ln1w, params.ln1b, B, T, C, main_stream);
-
+    layernorm_forward((model->recompute < 2) ? acts.ln1 : acts.lnf, acts.ln1_mean, acts.ln1_rstd, acts.encoded, params.ln1w, params.ln1b, model->use_kv, model->kv_offset, B, T, C, main_stream);
```
danielvegamyhre, line 657: If feasible, for readability's sake I would suggest not inlining a ternary expression for a function call parameter (`(model->recompute < 2) ? acts.ln1 : acts.lnf`), but rather storing the result of the ternary expression in a variable with a clear name and using that.

Something like:

```C
floatX* norm = (model->recompute < 2) ? acts.ln1 : acts.lnf;

layernorm_forward(norm, acts.ln1_mean, acts.ln1_rstd, acts.encoded, params.ln1w, params.ln1b, model->use_kv, model->kv_offset, B, T, C, main_stream);
```

Same suggestion goes for various other places in the file.

danielvegamyhre (commented):
This is a great contribution! I'll be curious to see if we can implement other inference optimizations and run some benchmarks to compare against well-known implementations.

I added a couple minor notes which may help improve readability in the main train_gp2.cu file where readability is a higher priority - feel free to accept or reject the suggestions at your own discretion :)

PR #708, opened 2024-07-23
Add high perf mode

gordicaleksa:
Add:
* Warnings when we take suboptimal branch
* High perf mode that will exit immediately if we're not running using all of the most optimal branches

Also added a fwd kernel config that will be useful for KV cache & mup PRs that are bloating up the argument list.

This way we get similar style like passing a config inside Python.


ngc92:
Currently, ZeRO also falls back to no sharding if the weight count is not divisible by the number of GPUs

PR #709, opened 2024-07-24
Allocate managed memory if device memory runs out

ngc92:
Use cudaMallocManaged to allocate optimizer states if we run out of device memory, so we can still train (slowly) even if we cannot fit the optimizer state
This is based on #694 , which should be merged first

### DIFF
--- a/train_gpt2.cu
+++ b/train_gpt2.cu
@@ -393,13 +393,13 @@ void gpt2_allocate_state(GPT2 *model, int B, int T) {
     printf0("allocating %zu MiB for AdamW optimizer state v\n", (shard_num_parameters * sizeof(float)) >> 20);
     assert(model->m_memory == nullptr);
     assert(model->v_memory == nullptr);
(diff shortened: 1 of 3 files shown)


train_gpt2.cu
```diff
@@ -393,13 +393,13 @@ void gpt2_allocate_state(GPT2 *model, int B, int T) {
     printf0("allocating %zu MiB for AdamW optimizer state v\n", (shard_num_parameters * sizeof(float)) >> 20);
     assert(model->m_memory == nullptr);
     assert(model->v_memory == nullptr);
-    cudaCheck(cudaMalloc((void**)&model->m_memory, shard_num_parameters * sizeof(float)));
-    cudaCheck(cudaMalloc((void**)&model->v_memory, shard_num_parameters * sizeof(float)));
+    cudaMallocConditionallyManaged((void**)&model->m_memory, shard_num_parameters * sizeof(float));
```
karpathy, line 396: imo we should try to update the import statements to show which file any function (e.g. the new Managed manaloc) comes from (here cuda_utils)

PR #712, opened 2024-07-25
Added permute kernel in dev/cuda

indianspeedster:
This PR primarily focuses on the educational aspects of llm.c. I observed that permute kernels are not present in dev/cuda directory, so I have added the implementation showing a small example of permuting a 4d tensor. Additionally, also have included explainations on how the permute operation works, along with guidance on developing a mental model for understanding permute operation while working with threads.


karpathy:
cool!

PR #714, opened 2024-07-28
Add RoPE positional encoding

gordicaleksa:
WIP. Will update description after i exit experimental stage.


llmc/encoder.cuh
```diff
@@ -17,7 +17,7 @@ In the backward pass, the gradients flow to both, handled by different kernels
 // CUDA kernels
 
 __global__ void encoder_forward_kernel3(floatX* out,
-                               const int* inp, const floatX* wte, const floatX* wpe,
+                               const int* inp, const floatX* wte, const floatX* wpe, int use_rope,
```
karpathy, line 20: I don't like this because `use_rope` has nothing to do with this encoding function, which we want to be nice and modular and self-contained. Maybe something like `use_positional` or something like that.

gordicaleksa: agree, can refactor a few of those bits quickly

PR #725, opened 2024-08-02
Add LLaMA 3 Python support

gordicaleksa:
Add LLaMA 3 support in our Python code acting as a reference.

The code supports only inference right now and is equivalent with [nano llama 3](https://github.com/karpathy/nano-llama31).


train_llama3.py
```diff
@@ -0,0 +1,1294 @@
...
+    if use_scaled:
+        freqs = apply_scaling(freqs)
+    freqs = torch.outer(t, freqs)
+    freqs_cis = torch.polar(torch.ones_like(freqs), freqs)  # complex64
+    return freqs_cis
+
+# -----------------------------------------------------------------------------
+# LLaMA building blocks
+
+class RMSNorm(torch.nn.Module):
```
karpathy, line 137: Add a comment about why we're not using nn.RMSNorm maybe?

train_llama3.py
```diff
@@ -0,0 +1,1294 @@
...
+        hidden_dim = 4 * config.n_embd
+        hidden_dim = int(2 * hidden_dim / 3)
+        # custom dim factor multiplier
+        if config.ffn_dim_multiplier is not None:
+            hidden_dim = int(config.ffn_dim_multiplier * hidden_dim)
+        hidden_dim = config.multiple_of * ((hidden_dim + config.multiple_of - 1) // config.multiple_of)
+        self.c_fc = nn.Linear(config.n_embd, hidden_dim, bias=False)
+        self.c_fc2 = nn.Linear(config.n_embd, hidden_dim, bias=False)
+        self.c_proj = nn.Linear(hidden_dim, config.n_embd, bias=False)
+        self.c_proj.LLMC_RESIDUAL_SCALE_FLAG = 1
```
karpathy, line 219: unused. two places

train_llama3.py
```diff
@@ -0,0 +1,1294 @@
...
+            loss = None
+
+        # there are performance reasons why not returning logits is prudent, if not needed
+        if not return_logits:
+            logits = None
+
+        return logits, loss
+
+    @staticmethod
+    def adapt_llama_state_dict_keys(checkpoint, config: LlamaConfig):
```
karpathy, line 326: Maybe a small comment/docs on these defs?

train_llama3.py
```diff
@@ -0,0 +1,1294 @@
...
+
+        self.n_words: int = num_base_tokens + len(special_tokens)
+        # BOS / EOS token IDs
+        self.bos_id: int = self.special_tokens["<|begin_of_text|>"]
+        self.eos_id: int = self.special_tokens["<|end_of_text|>"]
+        self.eot_id: int = self.special_tokens["<|eot_id|>"]
+        self.eom_id: int = self.special_tokens["<|eom_id|>"]
+        self.python_tag_id = self.special_tokens["<|python_tag|>"]
+        self.pad_id: int = self.special_tokens["<|finetune_right_pad_id|>"]
+        self.stop_tokens = [
```
karpathy, line 684: i fixed this in llama31. these stop tokens are incorrect for model.

train_llama3.py
```diff
@@ -0,0 +1,1294 @@
...
+                if current_slice_len > max_consecutive_slice_len:
+                    yield s[slice_start:i]
+                    slice_start = i
+                    current_slice_len = 1
+        yield s[slice_start:]
+
+# -----------------------------------------------------------------------------
+# Our own simple Distributed Data Loader
+
+def _peek_data_shard(filename):
```
karpathy, line 787: this code is all messed up and outdated atm here.
we need llama3 tokenizer encoded data. this is actively introducing bugs if someone tries to run it, reading GPT-2 tokenized data in uint16

train_llama3.py
gordicaleksa: just a tiny bit different numerics compared to llama's reference (tested), will leave a comment, we can later swap in nn.RMSNorm tbh

train_llama3.py
```diff
@@ -0,0 +1,1294 @@
...
+1) RoPE: LLaMA uses a different positional encoding scheme called Relative Positional Encoding (RoPE).
+2) GQA: Grouped Query Attention (GQA) is used to reduce the number of attention heads.
+3) SwiGLU: Swish-Gated Linear Unit (SwiGLU) is used as the activation function in the MLP.
+
+References:
+# 1) https://github.com/meta-llama/llama-models/blob/main/models/llama3_1/api/tokenizer.py
+# 2) https://github.com/meta-llama/llama-models/blob/main/models/llama3_1/api/model.py
+# 3) https://github.com/meta-llama/llama3/blob/11817d47e1ba7a4959b025eb1ca308572e0e3963/llama/generation.py
+
+Example launches to only benchmark the speed of bfloat16 compiled GPU training:
```
karpathy, line 15: delete launch commands that are incorrect atm

train_llama3.py
```diff
@@ -0,0 +1,1294 @@
...
+            )
+            new_freqs.append((1 - smooth) * freq / scale_factor + smooth * freq)
+    return torch.tensor(new_freqs, dtype=freqs.dtype, device=freqs.device)
+
+def apply_rotary_emb(
+    xq: torch.Tensor,
+    xk: torch.Tensor,
+    freqs_cis: torch.Tensor,
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    xq_ = torch.view_as_complex(xq.float().reshape(*xq.shape[:-1], -1, 2))
```
karpathy, line 116: can come later in a PR, but let's delete the use of complex. The use of complex here by Meta was a clear mistake, it created a lot of complexity for no good reason and iirc it broke torch.compile for me once earlier. In their latest code (for llamachat) Meta fixed this and they're now using a fully real-valued impl.

train_llama3.py
gordicaleksa: wasn't sure what to add and make it useful, should be fairly obvious (?)

train_llama3.py
```diff
@@ -0,0 +1,1294 @@
...
+        self.c_attn = nn.Linear(config.n_embd, (config.n_head + 2 * config.n_kv_head) * self.hd, bias=False)  # key, query, value projections
+        self.c_proj = nn.Linear(config.n_embd, config.n_embd, bias=False)  # output projection
+        self.c_proj.LLMC_RESIDUAL_SCALE_FLAG = 1
+
+        # static KV cache - we could alternatively allocate it outside of the model and just pass it in when needed
+        if self.use_kv:
+            self.cache_k = torch.zeros((config.max_gen_batch_size, config.block_size, config.n_kv_head, self.hd))
+            self.cache_v = torch.zeros((config.max_gen_batch_size, config.block_size, config.n_kv_head, self.hd))
+
+    def forward(self, x, freqs_cis=None, start_pos=None, mask: Optional[torch.Tensor] = None):
```
karpathy, line 171: Random thought I think type hints are dumb, if you delete them places I will basically always accept that change.
I think they are ok for cases where the type is not obvious.
And I prefer comments always strictly, because e.g. in Tensors the important thing is not that it's a tensor, but what its shape is, or what the dtype is, etc.

train_llama3.py
gordicaleksa: i fixed it just in a different place, by adding the right eos token; see `from_pretrained_llama3_hf` and `from_pretrained_llama3_meta`

we can refactor later once we support chat model?

train_llama3.py
```diff
@@ -0,0 +1,1294 @@
...
+            logits, _ = self.forward(tokens[:, prev_pos:cur_pos], start_pos=prev_pos)
+            if temperature > 0:
+                probs = torch.softmax(logits[:, -1] / temperature, dim=-1)
+                next_token = sample_top_p(probs, top_p)
+            else:
+                next_token = torch.argmax(logits[:, -1], dim=-1)
+
+            next_token = next_token.reshape(-1)
+            # only replace token if prompt has already been generated
+            next_token = torch.where(
```
karpathy, line 545: `black` pollution

train_llama3.py
gordicaleksa: kind of agree, esp. because they don't help catch an error, types are not enforced like in C

gordicaleksa: deleted the mask annotation everywhere

train_llama3.py
karpathy: I don't think so. Why adapt them?

train_llama3.py
karpathy: Ohh you override it there. Hmm I think leaving this here is a bit dangerous and possibly confusing, just as setting class attributes is. Maybe we take it as an arg in our code here?

train_llama3.py
gordicaleksa: Because their LLaMA class has different variable names compared to ours (we derive naming from GPT-2) (?)

kk will add it but tbh feels redundant as on a first skim people can see we're renaming keys in the checkpoint dict.

train_llama3.py
gordicaleksa: I fixed it using nano llama 3's solution

PR #733, opened 2024-08-10
Add llm.cpp(a port of this project using Eigen library, supporting CPU/CUDA), link to notable forks in readme

zhangpiu:
[llm.cpp](https://github.com/zhangpiu/llm.cpp/tree/master/llmcpp) is a fork of llm.c, which uses C++ (with Eigen) to reproduce GPT-2.

1. All calculations are done through the Eigen Tensor Module, so the same code can be used for CPU/CUDA calculations by simply switching the Device.
2. Currently, this repo has reproduced GPT-2 and the results are completely aligned with the PyTorch version.
3. It is worth noting that CPU calculations are about 20% faster than PyTorch, while GPU calculations are still far behind PyTorch's GPU due to the difficulty of Eigen Tensor Module to support BatchMatmul.

llm.c is an awesome project. I forked it and used C++ (with Eigen) to reproduce GPT-2, which taught me a lot. I am very grateful for this project.


invisiblepancake (commented):
Added a new section for C++/CUDA in README.md.
Introduced a new fork: [llm.cpp](https://github.com/zhangpiu/llm.cpp/tree/master/llmcpp) by [@zhangpiu](https://github.com/zhangpiu), which uses the [Eigen](https://gitlab.com/libeigen/eigen) library and supports both CPU and CUDA.

PR #736, opened 2024-08-10
Fix llama 3 data loader

gordicaleksa:
Add LLaMA 3 tokenization support for all our datasets:
1) tiny shakespeare
2) tiny stories
3) fineweb

### DIFF
--- a/dev/data/data_common.py
+++ b/dev/data/data_common.py
@@ -23,27 +23,30 @@ def download_file(url: str, fname: str, chunk_size=1024):
             bar.update(size)
 
 
-def write_datafile(filename, toks):
+HEADERS_INFO = {
+    "gpt-2": (20240520, 1),
+    "llama": (20240801, 7),
+}
+
+def write_datafile(filename, toks, model="gpt-2"):
     """
     Saves token data as a .bin file, for reading in C.
     - First comes a header with 256 int32s
-    - The tokens follow, each as a uint16
+    - The tokens follow, each as uint16 (gpt-2) or uint32 (llama)
     """
     assert len(toks) < 2**31, "token count too large" # ~2.1B tokens
+    assert model in ["gpt-2", "llama"], f"unknown model {model}"
     # construct the header
     header = np.zeros(256, dtype=np.int32)
-    header[0] = 20240520 # magic
-    header[1] = 1 # version
-    header[2] = len(toks) # number of tokens after the 256*4 bytes of header (each 2 bytes as uint16)
-    # construct the tokens numpy array, if not already
-    if not isinstance(toks, np.ndarray) or not toks.dtype == np.uint16:
-        # validate that no token exceeds a uint16
-        maxtok = 2**16
-        assert all(0 <= t < maxtok for t in toks), "token dictionary too large for uint16"
+    header[0] = HEADERS_INFO[model][0] # magic
+    header[1] = HEADERS_INFO[model][1] # version
+    header[2] = len(toks) # number of tokens after the 256*4 bytes of header
+    if model == "gpt-2":
         toks_np = np.array(toks, dtype=np.uint16)
+    elif model == "llama":
+        toks_np = np.array(toks, dtype=np.uint32)
     else:
-        toks_np = toks
-    # write to file
+        raise ValueError(f"unknown model {model}")
     print(f"writing {len(toks):,} tokens to {filename}")
     with open(filename, "wb") as f:
         f.write(header.tobytes())
--- a/dev/data/data_common.py
+++ b/dev/data/data_common.py
@@ -24,8 +24,14 @@ def download_file(url: str, fname: str, chunk_size=1024):
 
 
 HEADERS_INFO = {
-    "gpt-2": (20240520, 1),
-    "llama": (20240801, 7),
+    "gpt-2": {
+        "magic": 20240520,
+        "version": 1,
+    },
+    "llama": {
+        "magic": 20240801,
+        "version": 7,
+    },
 }
 
 def write_datafile(filename, toks, model="gpt-2"):
@@ -38,8 +44,8 @@ def write_datafile(filename, toks, model="gpt-2"):
     assert model in ["gpt-2", "llama"], f"unknown model {model}"
     # construct the header
     header = np.zeros(256, dtype=np.int32)
-    header[0] = HEADERS_INFO[model][0] # magic
-    header[1] = HEADERS_INFO[model][1] # version
+    header[0] = HEADERS_INFO[model]["magic"]
+    header[1] = HEADERS_INFO[model]["version"]
     header[2] = len(toks) # number of tokens after the 256*4 bytes of header
     if model == "gpt-2":
         toks_np = np.array(toks, dtype=np.uint16)
--- a/dev/data/fineweb.py
+++ b/dev/data/fineweb.py
@@ -64,33 +64,28 @@ elif args.type =="edu":
     fw = load_dataset("HuggingFaceFW/fineweb-edu", name=remote_name, split="train")
     name = "edu_fineweb"
 
-# init the tokenizer
-def tokenize(doc, model):
-    if model == "gpt-2":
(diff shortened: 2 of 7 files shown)


dev/data/fineweb.py
```diff
@@ -61,16 +65,33 @@
...
-eot = enc._special_tokens['<|endoftext|>'] # end of text token
-def tokenize(doc):
+def tokenize(doc, model):
+    if model == "gpt-2":
+        enc = tiktoken.get_encoding("gpt2")
+        encode = lambda s: enc.encode_ordinary(s)
+        eot = enc._special_tokens['<|endoftext|>'] # end of text token
+        tokens = [eot] # the special <|endoftext|> token delimits all documents
+    elif model == "llama":
+        tokenizer = AutoTokenizer.from_pretrained("meta-llama/Meta-Llama-3.1-8B")
```
karpathy, line 75: should we not use tiktoken in the exact same way as the official meta release shows? I like that it's more explicit. AutoTokenizer is a black box

dev/data/fineweb.py
```diff
@@ -61,16 +65,33 @@
...
+            return tokenizer(x).input_ids
+        eot = None
+        tokens = []
+    else:
+        raise ValueError(f"unknown model {model}")
+
     # tokenizes a single document and returns a numpy array of uint16 tokens
-    tokens = [eot] # the special <|endoftext|> token delimits all documents
-    tokens.extend(enc.encode_ordinary(doc["text"]))
+    text = doc["text"]
```
karpathy, line 84: this order was intentional, the delimiter should be prepended to docs, so you can inference just starting from that single token.

dev/data/fineweb.py
```diff
@@ -61,16 +65,33 @@
...
+    tokens.extend(encode(text))
     tokens_np = np.array(tokens)
-    assert (0 <= tokens_np).all() and (tokens_np < 2**16).all(), "token dictionary too large for uint16"
-    tokens_np_uint16 = tokens_np.astype(np.uint16)
-    return tokens_np_uint16
+
+    if model == "gpt-2":
```
karpathy, line 88: refactor to delete `if` and copy paste dcode, use ternary operator to set the `upper_bound`

dev/data/fineweb.py
```diff
@@ -99,7 +120,7 @@ def tokenize(doc):
             remainder = args.shard_size - token_count
             progress_bar.update(remainder)
             all_tokens_np[token_count:token_count+remainder] = tokens[:remainder]
-            write_datafile(filename, all_tokens_np)
+            write_datafile(filename, list(all_tokens_np), args.model)
```
karpathy, line 123: why convert to list

dev/data/fineweb.py
```diff
@@ -61,16 +65,33 @@
     name = "edu_fineweb"
 
 # init the tokenizer
-enc = tiktoken.get_encoding("gpt2")
-eot = enc._special_tokens['<|endoftext|>'] # end of text token
-def tokenize(doc):
+def tokenize(doc, model):
```
karpathy, line 68: instead of taking model and having a big `if` inside `def` we can have two defs for the two options, and dispatch accordingly

dev/data/fineweb.py
```diff
@@ -22,18 +22,22 @@
...
 import argparse
 import multiprocessing as mp
+
 import numpy as np
 import tiktoken
 from datasets import load_dataset
 from tqdm import tqdm
-import argparse
+
+from transformers import AutoTokenizer
```
karpathy, line 31: delete dep

dev/data/data_common.py
```diff
@@ -23,27 +23,30 @@ def download_file(url: str, fname: str, chunk_size=1024):
             bar.update(size)
 
 
-def write_datafile(filename, toks):
+HEADERS_INFO = {
```
karpathy, line 26: i prefer to use dict of dicts, is more backwards compatible, flexible, and self-documenting. instead of tuples.

dev/data/fineweb.py
gordicaleksa: autotokenizer is a blackbox, I agree, but i believe we can be confident that at least for an architecture as popular as LLaMA 3 HuggingFace is battle-tested!

I prefer that over downloading a tokenizer and passing in a path.

What do you think?

dev/data/fineweb.py
gordicaleksa: and a ternary in the assert statements as well?

dev/data/fineweb.py
gordicaleksa: i simplified `write_datafile` so that it doesn't have to handle both numpy & lists, it's cleaner i think

dev/data/fineweb.py
gordicaleksa: as in if we decide not to use AutoTokenizer (due to its black-box nature)?

dev/data/data_common.py
gordicaleksa: done!

dev/data/fineweb.py
gordicaleksa: Sure, at this point, I don't have a special preference as we only support 2 models.

dev/data/fineweb.py
karpathy: There could be many tokens, so creating a Python list could be very wasteful

dev/data/tinyshakespeare.py
```diff
@@ -37,7 +42,17 @@ def download():
...
 
-def tokenize():
+def tokenize(model):
+    if model == "gpt-2":
+        enc = tiktoken.get_encoding("gpt2")
+        encode = lambda s: enc.encode(s, allowed_special={'<|endoftext|>'})
+    elif model == "llama":
+        tokenizer = AutoTokenizer.from_pretrained("meta-llama/Meta-Llama-3.1-8B")
+        def encode(x):
+            return tokenizer(x).input_ids
```
karpathy, line 52: pretty sure this now creates bug for this code because <|endoftext|> (below) doesn't tokenize properly

dev/data/fineweb.py
karpathy:
```python
>>> a = np.random.randn(10)
>>> a
array([-1.39200423,  0.91909499,  0.49247546,  0.73578011, -0.46485352,
        0.06844696,  1.21521025,  0.18951044, -0.33376094,  1.03115886])
>>> list(a)
[-1.3920042324598616, 0.9190949922347375, 0.49247545796208686, 0.7357801064341112, -0.4648535191489631, 0.06844695804812885, 1.2152102515229188, 0.18951044050354424, -0.33376094056177236, 1.0311588596558752]
>>> z = list(a)
>>> z[0]
-1.3920042324598616
>>> type(z[0])
<class 'numpy.float64'>
```

karpathy: vs `tolist`

```python
>>> a = np.random.randn(10)
>>> a
array([-0.28416783,  3.61778557,  0.45557321,  0.6585392 , -0.54974637,
       -0.50662981,  0.36080734,  0.76378507, -1.60443242,  0.41719901])
>>> a.tolist()
[-0.2841678282966355, 3.6177855666263548, 0.45557321422210056, 0.6585391952854299, -0.5497463693208792, -0.5066298099246493, 0.36080734397795633, 0.7637850737170351, -1.6044324246329673, 0.4171990143035489]
>>> z = a.tolist()
>>> z[0]
-0.2841678282966355
>>> type(z[0])
<class 'float'>
```

dev/data/fineweb.py
```diff
@@ -80,6 +96,15 @@ def tokenize(doc):
     all_tokens_np = np.empty((args.shard_size,), dtype=np.uint16)
     token_count = 0
     progress_bar = None
+
```
karpathy, line 99: doesn't `def tokenize` break because:

```python
all_tokens_np = np.empty((args.shard_size,), dtype=np.uint16)
```

i.e. the init is using uint16

PR #737, opened 2024-08-12
multi-threaded model initialization

ngc92:
A simple demonstration of initializing from multiple threads.
Factored out the parallel part; by making it a separate function, it becomes less likely that we accidentally change shared state by using variables from the outer scope.


train_gpt2.cu
```diff
@@ -504,6 +504,47 @@ void gpt2_build_from_checkpoint(GPT2 *model, const char* checkpoint_path, bool w
     cudaCheck(cudaDeviceSynchronize());
 }
 
+void gpt2_init_layer(GPT2 *model, int l, mt19937_state* rng, floatX* params) {
+    int offset = 0;
```
ademeure, line 508: This should be size_t for larger model sizes.

ademeure:
Looks good to me, it improves startup time for -e "d72" from ~100s to ~15s on a 1xH100 node with 26 CPU cores! :)

PR #738, opened 2024-08-12
Improve compile time (simple makefile changes)

ademeure:
This improves compile times by:
1) Using nvidia-smi instead of __nvcc_device_query (the latter was wasting >1 second per build on a 1xH100 system!)
2) "--threads=0" allowss NVCC to use multithreading for some things (seems to improve compile times by ~5% locally.)
3) Making -O3 configurable with the FORCE_NVCC_O parameter (O0 improves compile times by >30% but significantly hurts run times; -O1 seems like a good compromise when iterating on some code.)

(1) and (2) reduce compile time with cuDNN from ~9.2s to ~8.1s.
(3) with O2/O1/O0 reduces it further to ~7.7s / ~6.5s / ~5.4s.


karpathy:
nice!

karpathy:
surprised about both 1 and 2...

PR #740, opened 2024-08-13
Gordicaleksa fix dataloader2

karpathy:
commit on top of @gordicaleksa PR that makes a bunch of bugfixes

- be more explicit with treatment of EOT token
- be careful with API for AutoTokenizer
- bugfix on dtype in fineweb.py
- use model_desc instead of model because the latter is usually a nn.Module
- few more

### DIFF
--- a/dev/data/tinyshakespeare.py
+++ b/dev/data/tinyshakespeare.py
@@ -7,12 +7,14 @@ The output is written to a newly created tinyshakespeare/ folder.
 The script prints:
 
 For GPT-2:
-writing 32,768 tokens to tinyshakespeare/tiny_shakespeare_val.bin
-writing 305,260 tokens to tinyshakespeare/tiny_shakespeare_train.bin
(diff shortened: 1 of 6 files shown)


dev/data/tinyshakespeare.py
```diff
@@ -42,36 +44,41 @@ def download():
...
-    text = text.replace('\n\n', '\n\n<|endoftext|>')
-    # encode the text
-    tokens = encode(text)
+    # let's treat every individual chunk of text as a separate "document"
+    sections = text.split("\n\n")
+    tokens = []
+    for i, s in enumerate(sections):
+        tokens.append(eot)
+        # there was a mild bug where I originally intended to remove \n\n, but instead just added
+        # the EOT right after each \n\n, so I'm keeping that behavior for backwards compatibility
```
gordicaleksa, line 66: backwards compatibility with what code?

karpathy: the previous version of this code

PR #741, opened 2024-08-13
[WIP] initial curand implementation for model init

ngc92:
as an alternative to the multi-threaded model init, this uses curand to generate initial weights directly on the GPU.
It is still work-in-progress, needs error-checking, and I dislike the cudamalloc inside the loop, but it should be enough to measure how much speed-up we would get.


writemorecode:
Perhaps a stupid question, but could the "fill_kernel" be replaced with a "cudaMemsetAsync"?

ngc92:
`cudaMemset` works on the level of individual bytes; we want to set floats or bf16s here; for zero, conveniently repeating the bit-pattern 0 also implies the number zero, so you can use it to initialize floats to 0.0; but unless your desired float has the same bits in every byte, it isn't going to work for other values.

PR #746, opened 2024-08-22
log with LINE and FILE for better addressing.

NEWPLAN:
log with line and file for better addressing.


NEWPLAN (commented):
OK, add more location info for log.

PR #748, opened 2024-08-25
Fix typo in `train_gpt2_fp32.cu`

gajanan-choudhary:
Fixes a simple (harmless) typo that uses `NUM_ACTIVATION_TENSORS` instead of `NUM_BACKWARD_TENSORS` in `train_gpt2_fp32.cu`.


karpathy:
Code works accidentally because 21 > 3 right?

gajanan-choudhary:
> Code works accidentally because 21 > 3 right?

Right. I tested the change on A100 locally CUDA 12.6 and it works fine.

PR #755, opened 2024-09-13
Add SwiGLU support - llama3 feature branch

gordicaleksa:
Implemented SwiGLU - swish GLU activation function from the "GLU Variants Improve Transformer" [paper](https://arxiv.org/pdf/2002.05202).

Note: there is an increase in memory footprint as a consequence of adding an additional FC layer (params/grads/optimizer states + 2 new activation buffers `(L+1)*B*T*4*C` in total for the activations). I'm sure we can optimize this, can be done in the subsequent PR as well.

Tests:
I ran an A/B experiment: trained a 124M GPT-2 on 10B tokens (FineWeb subset) with:
a) GELU
b) SwiGLU (note: currently SwiGLU model actually has 152M params)

Results:
image

Conclusion: SwiGLU converges to a lower loss, so I'm confident the implementation is correct, but unclear whether the perf comes from SwiGLU or more params (152M vs 124M).

After normalizing for the number of params (multiplying the inner FFN module dimension by 2/3) I get:
image

Conclusion: SwiGLU does start converging faster than GELU but they end up at a pretty much same loss. Actually the trend seems to go in favor of GELU looking at the gradient of the curves.

Given that SwiGLU complicates the code, it's unclear to me that it offers a clear advantage over GELU/ReLU/etc. In general i doubt any of these activation functions are more powerful than ReLU. You just want more params / compute. :)

Next steps (can be done in follow-up PRs):
* Reduce memory footprint
* Speed up kernels / logic

Appendix:
In case it helps here is a diagram I drew to make it easier for me to implement the backward pass:
swiglu_bwd

### DIFF
--- a/llmc/gelu.cuh
+++ b/llmc/gelu.cuh
@@ -48,17 +48,19 @@ __global__ void swiglu_forward_kernel(floatX* out, const floatX* inp1, const flo
     int idx = (blockIdx.x * blockDim.x + threadIdx.x) * x128::size;
 
     x128 packed_out;
-    x128 packed_inp1 = load128cs(inp1 + idx);
+    x128 packed_inp1 = load128cs(inp1 + idx); // load and do not keep in cache
     x128 packed_inp2 = load128cs(inp2 + idx);
     for(int k = 0; k < packed_inp1.size; ++k) {
         float x1 = (float)packed_inp1[k];
         float x2 = (float)packed_inp2[k];
+        // swish(x1) = x1 * sigmoid(x1) = x1 / (1.0 + exp(-x1))
+        // swiglu(x1, x2) = swish(x1) * x2
         packed_out[k] = (floatX)((x1 * x2) / (1.0f + expf(-x1)));
     }
     store128(out + idx, packed_out);
 }
 
-__global__ void swiglu_backward_kernel(floatX* dinp_out1, floatX* dinp2, const floatX* inp1, const floatX* inp2) {
+__global__ void swiglu_backward_inplace_kernel(floatX* dinp_out1, floatX* dinp2, const floatX* inp1, const floatX* inp2) {
     int idx = (blockIdx.x * blockDim.x + threadIdx.x) * x128::size;
 
     x128 packed_dinp1;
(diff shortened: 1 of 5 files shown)


train_llama3.cu
```diff
@@ -708,16 +743,17 @@ void gpt2_forward(GPT2 *model, const int* inputs, size_t B, size_t T) {
...
-        floatX* l_fch_gelu = (model->recompute < 1) ? acts.fch_gelu + l * B * T * 4*C : acts.fch_gelu;
+        floatX* l_fch_act = (model->recompute < 1) ? acts.fch_act + l * B * T * 4*C : acts.fch_act;
         floatX* l_residual3 = acts.residual3 + l * B * T * C;
         floatX* scratch = (floatX*)acts.output; // used for non-cudnn attention, fcproj, attproj, etc.
 
         // now do the forward pass
         #ifdef ENABLE_CUDNN
         float* l_att = (float*)acts.att + l * B * NH * T; // cuDNN needs a smaller FP32 tensor
-        matmul_forward_cublaslt(l_qkvr, l_ln1, l_qkvw, l_qkvb, B, T, C, 3*C, main_stream);
+        matmul_forward_cublaslt(l_qkvr, l_ln1, l_qkvw, l_qkvb, B, T, C, 3*C, main_stream, model->act_func);
```
ademeure, line 756: Would it make sense to make the act_func parameter of matmul_forward_cublaslt() default to NULL, and not pass it here for all the non-GELU/SwiGLU matmul calls, since it's not actually used?

train_llama3.cu
```diff
@@ -883,29 +925,39 @@ void gpt2_backward_and_reduce(GPT2 *model, int* inputs, const int* targets, int
...
 
         floatX* dl_bt4c = (floatX*)model->acts.scratch_bt4c;
+        floatX* dl_bt4c2 = (floatX*)model->acts.scratch_bt4c2;
 
         // start the backward pass for this layer
         if(model->recompute >= 1) {
-            // recompute >= 1 means we recompute gelu. in this case,
-            // l_fch_gelu is just a buffer, so re-compute the gelu from l_fch here
-            gelu_forward(l_fch_gelu, l_fch_pre_gelu, B*T*4*C, main_stream);
+            if (strcmp(model->act_func, "gelu") == 0) {
```
ademeure, line 940: would it make sense to do a call to a single activation_forward() call, and rename gelu.cuh to activation.cuh since it now also does SwiGLU?

Assuming we want to keep this flexible at all, but it feels like experimenting with different activation functions is an easy thing that'd be worth doing a little bit more in the future (e.g. loss impact & GPU throughput of RELU and RELU squared for example).

Right now we'd just pass all the arguments of swiglu_forward() since gelu_forward()'s arguments are a subset (that won't scale if adding more complicated activation functions but I think that's probably OK/still useful?)

train_llama3.cu
```diff
@@ -1004,11 +1059,12 @@ float gpt2_calculate_grad_norm(GPT2 *model, MultiGpuConfig* multi_gpu_config) {
         // grads_memory only contains the averaged gradients at the local shards,
         // so we only calculate the grad norm at the grads_memory belonging to the local shards
         for (int i = 0; i < NUM_PARAMETER_TENSORS; i++) {
+            if (!gated_ffn && (i == 12 || i == 13)) { continue; }  // skip the gated ffn weights
```
ademeure, line 1062: is this just for the case when the activation function is not SwiGLU, so there are no gated ffn weights? If so the comment is a bit misleading

PR #763, opened 2024-09-21
rmsnorm backward simple baseline kernel

ngc92:
baseline kernel and cpu versions

### DIFF
--- a/dev/cuda/rmsnorm_backward.cu
+++ b/dev/cuda/rmsnorm_backward.cu
@@ -1,0 +1,235 @@
+/*
+Kernels for layernorm backward pass.
+
+Compile example:
+nvcc -O3 --use_fast_math -lcublas -lcublasLt rmsnorm_backward.cu -o rmsnorm_backward
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <cuda_runtime.h>
+#include <assert.h>
+#include <cooperative_groups.h>
+#include <cooperative_groups/reduce.h>
+
+#include "common.h"
+
+// ----------------------------------------------------------------------------
+// CPU code reference
+
+void rmsnorm_forward_cpu(float* out, float* rstd,
+                         const float* inp, const float* weight,
+                         int B, int T, int C) {
+    float eps = 1e-5f;
+    for (int b = 0; b < B; b++) {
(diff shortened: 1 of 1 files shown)


dev/cuda/rmsnorm_backward.cu
```diff
@@ -0,0 +1,235 @@
...
+    for (int b = 0; b < B; b++) {
+        for (int t = 0; t < T; t++) {
+            const float* dout_bt = dout + b * T * C + t * C;
+            const float* inp_bt = inp + b * T * C + t * C;
+            float* dinp_bt = dinp + b * T * C + t * C;
+            const float rstd_bt = rstd[b * T + t];
+
+            float o = 0.0f;
+            for (int i = 0; i < C; i++) {
+                o += weight[i] * dout_bt[i] * inp_bt[i];
```
azret, line 60: Thank you for this! As a result of this PR I have found a bug in my own RMSNorm impl. where I forgot to multiply by the weight in the backward pass. Spent days chasing it.

PS: I've ran this rmsnorm_backward_cpu through my test cases that verify against torch and all is passing.

Thanks again!

azret: void rms_backward(
            float* _In, float* _δ_Out, float* _δ_In,
            float* _δ_Gamma, float* _Gamma,
            int N, float EPSILON) {
            float rms = 0;
            for (int i = 0; i < N; i++) {
                rms += _In[i] * _In[i];
            }
            rms = std.sqrtf(rms / N + EPSILON);
            float rrms = 1.0f / rms;
            float rnrms = 1.0f / (N * rms * rms);
            float δ = 0;
            for (int i = 0; i < N; i++) {
                δ += _δ_Out[i] * _In[i] * rrms * **_Gamma[i]**; // This is what I missed!!
            }
            for (int i = 0; i < N; i++) {
                _δ_In[i] += _δ_Out[i] * _Gamma[i] * rrms - δ * _In[i] * rnrms;
                _δ_Gamma[i] += _δ_Out[i] * _In[i] * rrms;
            }
        }

ngc92: great to hear that this passes an independent set of tests. I did some testing, but wasn't yet fully sure that my implementation was correct.

Issue #799, opened 2025-03-21
Loss computation is wrong


YAMWD:
The current loss computation is wrong. The correct loss is computed with pair (current logit, label of the next word) because the model is trained to predict the next token in a sequence. So the logits should be shifted right (last element ignored) and targets should be shifted left (first value ignored) to aligns the predictions (logits) with the correct target tokens.

`if targets is not None:`
`    # if we are given some desired targets also calculate the loss`
`    logits = self.lm_head(x)`
`    shift_logits = logits[..., :-1, :].contiguous()`
`    shift_labels = targets[..., 1:].contiguous()`
`    loss = F.cross_entropy(shift_logits.view(-1, shift_logits.size(-1)), shift_labels.view(-1), ignore_index=-1)`

It is also the implementation in huggingface transformers repo.

YAMWD:
In current test_gpt2.c, the expected_loss is 5.27, but the correct loss should be 9.579826 using the GPT2LMHead model from transformers with the same inputs.

YAMWD:
I noticed that such shifts is done in the dataloader so the original implementation should be fine.

PR #800, opened 2025-03-22
fix loss computation by using shifted logits and targets to create co…

YAMWD:
fix loss computation by using shifted logits and targets. 

issue link [https://github.com/karpathy/llm.c/issues/799](url)


YAMWD:
It also needs modification on softmax. Now the loss is correct.

image

YAMWD:
I noticed that such shifts are done in the dataloader so the original implementation should be fine.

PR #801, opened 2025-04-13
Fix gradient tests

ngc92:
The order of tensors inside our buffer differs from the order in which we run our gradient checks.
As the layout of the reference and test tensor are the same, we're still comparing corresponding elements, but the mapping of (named) tensors and tolerances into the buffer is wrong; potentially, we didn't compare some elements, and compared others twice.

This PR reorders the tensor names and thresholds to correspond to our actual model definition, and adapts the thresholds accordingly (this is much better visible by looking at the diff of just the second commit). For layernorms, the thresholds had to be increased quite a lot (maybe it should have been suspicious that the last LN needed so much larger tolerances than the others; now they're more equal), but for some other tensors we could actually tighten the error bounds.

The error thresholds have been tested on an A6000.

### DIFF
--- a/test_gpt2.cu
+++ b/test_gpt2.cu
@@ -267,29 +267,26 @@ int main(int argc, char *argv[]) {
             // Also, different GPUs may use different matrix multiplication algorithms, so the
             // actual errors can be hardware specific.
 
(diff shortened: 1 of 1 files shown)


karpathy:
good catch, agree on the order.

PR #819, opened 2025-06-21
Create PMLL.c

drQedwards:
And another one

MR. karpathy I’m ready to turn in my homework please also see
https://www.authorea.com/users/856117/articles/1297819-the-persistent-memory-logic-loop-from-proposal-design-to-formal-proof-of-p-np-using-the-pmll-algorithm

### DIFF
--- a/PMLL.c
+++ b/PMLL.c
@@ -1,0 +1,102 @@
+/********************************************************************
+  PMLL.c  —  Persistent-Memory Logic Loop (CPU reference)
+  • ring-buffer KV store per attention head
+  • zero external dependencies
+  • thread-safe if you call it from a single OpenMP thread per batch
+  • MIT licence (same as llm.c)
+*********************************************************************/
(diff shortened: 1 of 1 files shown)


PMLL.c
```diff
@@ -0,0 +1,102 @@
...
+    S->T = S->hs = 0;
+}
+
+/*---------------------------------------------------------------
+  Data flow
+----------------------------------------------------------------*/
+
+/* Read historic KV for head *h* to caller-provided scratch.
+   out_k/out_v length = (Tmem + Tctx) * hs floats. */
+void pmll_read(float *out_k, float *out_v,
```
PatrickvL, line 50: Unless there is a specific reason not to do so, `pmll_read()` should have the 
`const pmll_state *S` argument first for consistency.

drQedwards: Got it!

